PR #2, opened 2020-06-17
make MNO requests UI more like the design prototype

aldavidson:

styling, layout changes to replicate the design prototype as much as possible

### Guidance to review

Before:

Screen Shot 2020-06-17 at 16 57 01

After:

Screen Shot 2020-06-18 at 10 48 54


app/views/mno/recipients/index.html.erb
```diff
@@ -1,69 +1,81 @@
 <h1 class="govuk-heading-l">
```
fofr, line 1:
```suggestion
<h1 class="govuk-heading-xl">
```

app/views/mno/recipients/index.html.erb
```diff
@@ -1,69 +1,81 @@
...
-  Your requests
+  Requests for extra mobile data
+  <span class="govuk-caption-xl"><%= @mobile_network.brand %> customers</span>
```
fofr, line 3:
```suggestion
  <span class="govuk-caption-xl"><%= @mobile_network.brand %></span>
  Requests for extra mobile data
```

app/views/mno/recipients/index.html.erb
```diff
@@ -1,69 +1,81 @@
...
+                  |
+                  <%= link_to 'none', {select: 'none'} %>
+                </div>
+                <div id="all-none-checkbox" class="js-only govuk-checkboxes__item" data-controls="all-none-item">
+                  <input class="govuk-checkboxes__input" id="all-rows" name="all-rows" type="checkbox" value="all-rows">
+                  <label class="govuk-label govuk-label--s govuk-checkboxes__label" for="all-rows"></label>
+                </div>
+              </th>
+              <th class="govuk-table__header govuk-table__header--numeric">
+                <%= sortable_table_header( 'ID', :id ) %>
```
fofr, line 42: Can we drop this column?

app/views/mno/recipients/index.html.erb
```diff
@@ -1,69 +1,81 @@
...
+              </th>
+              <th class="govuk-table__header govuk-table__header--numeric">
+                <%= sortable_table_header( 'Mobile number', :mobile_number ) %>
+              </th>
+              <th class="govuk-table__header">
+                <%= sortable_table_header( 'Account holder', :account_holder_name ) %>
+              </th>
```
fofr, line 49: Can we swap these two columns around?

app/webpacker/packs/application.js
```diff
@@ -5,6 +5,10 @@ import Rails from 'rails-ujs';
 import Turbolinks from 'turbolinks';
 import { initAll } from 'govuk-frontend';
 
+
```
fofr, line 8: Rogue whitespace

app/webpacker/packs/mno/recipients.js
```diff
@@ -0,0 +1,19 @@
+hideOnLoad = function() {
+  $('.non-js-only').hide();
+  $('.js-only').show();
+}
+
+setupSelectAllNone = function() {
+  $('#all-rows').change(function() {
+    console.log('change!')
+    console.log($(this))
```
fofr, line 9: Delete?

app/webpacker/styles/application.scss
```diff
@@ -12,3 +12,4 @@ $govuk-image-url-function: frontend-image-url;
 @import "~govuk-frontend/govuk/all";
 
 @import "thisapp/pagination.scss";
+@import "thisapp/overrides.scss";
```
fofr, line 15: Why the `thisapp` namespace?

app/webpacker/styles/thisapp/overrides.scss
```diff
@@ -0,0 +1,16 @@
...
+    display: inline-block;
+  }
+}
+
+td {
+  .govuk-form-group {
+    margin-bottom: 0;
+  }
+}
+
```
fofr, line 12: These overrides are dangerous.

Better to use the helper classes in the HTML as they were intended:

`govuk-!-margin-bottom-0` and `govuk-!-display-inline-block`

app/webpacker/styles/thisapp/pagination.scss
```diff
@@ -1,4 +1,6 @@
+.count {
```
fofr, line 1: Suggest: `.pagination-count`

aldavidson:
> Looking good
> 
> Can we use the `govuk_link` helper to make links show in the correct colour?

done!

app/webpacker/styles/application.scss
aldavidson: No particular reason other than it makes it very explicit that these are from (and for) specifically this application, rather than anywhere else in the import path. Happy to change or remove if it's a problem?

app/webpacker/styles/thisapp/overrides.scss
aldavidson: Good call - I didn't know about those classes, now I do  :)

I've taken out the margin-bottom override, but I can't get a custom class name to pass through the `govuk_form_builder (...) govuk_collection_select` helper to actually appear in the HTML.  (`html_options: {class: '...'}` doesn't seem to work). 

OK to stick with the `govuk-form-group.inline` hack for now?

aldavidson: as discussed elsewhere, I've kept the CSS hack but with a more specific name (`.display-child-forms-inline`) and an explanatory comment

app/webpacker/styles/application.scss
fofr: I think that if the files are in the repo it's a given.

PR #9, opened 2020-06-20
port SHOW_DEBUG_INFO behaviour to use a FeatureFlag

aldavidson:
### Context

We've been using an env var SHOW_DEBUG_INFO as a temporary way to render the session & sign-in links into the footer during development. Now we have an actual FeatureFlag implementation, it should use that.

1. replace the SHOW_DEBUG_INFO with a show_debug_info FeatureFlag
2. add feature_flag spec for the behaviour
3. add a section to the README about FeatureFlags and how to set them locally and on PaaS

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,3 +59,31 @@ bundle exec brakeman
 3. Run `make (dev|staging|prod) build push deploy` to build, push and deploy the Docker image to Gov.uk PaaS
 
 The app should be available at https://get-help-with-tech-(dev|staging|prod).london.cloudapps.digital
+
+## Feature Flags
+
+Certain aspects of app behaviour are governed by a minimal implementation of Feature Flags.
+These are activated by having an environment variable FEATURES_(flag name) set to 'active', for example:
+
+```
+# start the rails server with debug info rendered into the footer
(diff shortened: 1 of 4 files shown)


aldavidson:
> Gonna approve to not block things but not convinced we'd ever want this turned on in production – I wonder whether we'd want to toggle the visibility based on which environment it's on, which would mean one less feature flag.

"not convinced we'd ever want this turned on in production " - I agree, but my experience with limiting things by environment is that :
- it doesn't work well with Docker (where everything tends to be built with RAILS_ENV=production so that you don't deploy dev/test gems to production, etc)
- it starts out clean and simple, but can very quickly end up with a growing list of environments that are either non-prod and should be excluded, or are prod-like (QA / staging / pre-live etc) and should be included
- Overall it's just much clearer to keep environment names out of your code
- From an ops point of view, it's better to be verbose and explicit about what flag does what. A set of vars like `FEATURES_show_debug_info=active` is much clearer than, say, IS_PROD_LIKE=true governing a whole set of behaviours that you have to dig through the whole codebase to figure out.

benilovj:
@aldavidson fair enough, 🚢

PR #10, opened 2020-06-20
Make sign_in_tokens expire after 30 mins by default

aldavidson:
### Context

See [Trello card](https://trello.com/c/3tRtyZrZ/121-make-signintokens-time-limited)

Make sign_in_tokens expire after 10 mins by default. 
Add option to set a different expiry period via env var
Add spec for expiry/non-expiry behaviour

### Guidance to review

1. Generate a sign-in token by clicking 'Sign in' and entering your email address. Copy the sign-in URL from the footer
2. From the console, check that `sign_in_token_expires_at` is set correctly (`User.where(email_address: '...').sign_in_token_expires_at`)
3. Wait until that expires (or update it from the console), and then try and use the sign-in token URL - it should not work

4.

### DIFF
--- a/app/concerns/sign_in_with_token.rb
+++ b/app/concerns/sign_in_with_token.rb
@@ -2,16 +2,22 @@ module SignInWithToken
   extend ActiveSupport::Concern
 
   included do
+    DEFAULT_SIGN_IN_TOKEN_TTL = 1800
+
     validates :sign_in_token, uniqueness: true, allow_nil: true
 
     def generate_token!(ttl: nil)
-      ttl ||= (ENV['SIGN_IN_TOKEN_TTL_SECONDS'] || 1800)
+      expiry_time = Time.now.utc + effective_ttl(ttl).seconds
 
       update!(
         sign_in_token: SecureRandom.uuid,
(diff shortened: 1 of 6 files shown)


fofr:
10 minutes feels a bit short to me.

aldavidson:
> 10 minutes feels a bit short to me.

It's just a default - the main point of this PR was to stop them having infinite lifetime. Once the code change is in, the TTL can be configured in seconds via the env var

app/concerns/sign_in_with_token.rb
```diff
@@ -4,13 +4,26 @@ module SignInWithToken
   included do
     validates :sign_in_token, uniqueness: true, allow_nil: true
 
-    def generate_token!
-      update!(sign_in_token: SecureRandom.uuid)
+    def generate_token!(ttl: nil)
+      ttl ||= (ENV['SIGN_IN_TOKEN_TTL_SECONDS'] || 1800)
```
benilovj, line 8: Can we please pull out 1800 as a constant at the top of the file? it's a bit buried here.

aldavidson: Yep, fair point - I've done that and also extracted the calculation of effective TTL out to a separate method

PR #18, opened 2020-06-24
BT wifi pilot docs

benilovj:
### Context

We need to host the privacy notice, user guide and provide a link to support

image

image


app/views/pages/bt_wifi_privacy_notice.html.erb
```diff
@@ -0,0 +1,81 @@
...
+
+<!--
+
+
+#
+
+
+For further information and how to request your data, please use the ‘contact form’ in the Personal Information Charter at
+
+https://www.gov.uk/government/organisations/department-for-education/about/personal-informa1tion-charter
```
aldavidson, line 78: is there a typo in `personal-informa1tion-charter` ?

aldavidson: I guess it's a comment, so not a blocker

benilovj: oops 😬  – that whole thing needs to go. fixed and force pushed

fofr:
This uses the old BT user guide.

PR #19, opened 2020-06-24
Send token mails via Notify

aldavidson:
### Context

The app so far has not been sending emails. We need to do this if we are to launch.

Add `sign_in_token_mailer.rb` to send mails
Add env vars for the Notify API key and template id, plus the hostname to use for generating emails
Add `log_email_delivery_observer.rb` to debug-log mail bodies out to the console (this won't affect production as the log level there is just set to INFO)

### Guidance to review

You'll need to set up the env vars:
`GOVUK_NOTIFY_API_KEY=(your test API key from notify)`
- you might want to put this into a .env file (they're ignored in .gitignore and .dockerignore)

Click 'sign in'
EITHER click 'yes' and enter a valid user email address, or click 'no' and sign-up 
When you get to the "Check your email" screen, in the console you should have a log message that outputs the body of a generated email, looking something like this:

Screen Shot 2020-06-24 at 15 33 27

You can also view it in the Notify UI:
Screen Shot 2020-06-24 at 16 37 11


app/helpers/view_helper.rb
```diff
@@ -8,9 +8,12 @@ def govuk_link_to(body, url, html_options = {}, &_block)
   end
 
   def ghwt_contact_mailto(subject: 'Increasing%20internet%20access')
-    govuk_link_to 'COVID.TECHNOLOGY@education.gov.uk',
-      'mailto:COVID.TECHNOLOGY@education.gov.uk' +
-      (subject.present? ? "?subject=#{subject}" : '')
+    mail_to_url = [
+      'mailto:COVID.TECHNOLOGY@education.gov.uk',
+      (subject.present? ? "subject=#{subject}" : nil),
+    ].compact.join('?')
```
benilovj, line 14: 👍

PR #24, opened 2020-06-25
Make new users pending

aldavidson:
### Context

Anyone can create an account and submit data, so long as they can validate their email with a token link.
We agreed in this morning's prioritisation session that to prevent abuse, we'll :
1. mark new users as pending
2. have a manual approval process for users
3. hide any requests submitted by unapproved users from the MNO interface
4. hide & disable form submission by users who have not signed in

All the above, plus removal of a lot of now-redundant cruft & complexity from the forms, views & controllers
Also removed the critical PII (name, address, postcode of young person) which we don't actually need

### Guidance to review

1. Visit a page without being signed in - you should only see the 'Guidance' & 'Sign in' links in the nav
2. Create an account - you should be able to see the 'Tell us...' links once you're signed in
3. Submit a 'eligible young person' form for a participating mobile network with which you have a user account (or create a user with `mobile_network_id` set from the console) - you should be able to submit data with no problems
4. Sign out, and sign in as the MNO user - you should not see the request you just submitted in 'Your requests'
5. From the console, set `approved_at` to a valid datetime on the user who submitted the data, and save that user
6. Reload the MNO 'Your requests' page - you should now be able to see the request

Simpler form UI, more inline with design prototype:

Screen Shot 2020-06-25 at 01 26 21

### DIFF
--- a/app/controllers/application_forms_controller.rb
+++ b/app/controllers/application_forms_controller.rb
@@ -1,17 +1,15 @@
 class ApplicationFormsController < ApplicationController
   before_action :require_sign_in!
-  
+
   def new
-    @application_form = ApplicationForm.new(user: @user)
+    @application_form = ApplicationForm.new
     @mobile_networks = MobileNetwork.order('LOWER(brand)')
   end
 
   def create
-    @application_form = ApplicationForm.new(user: @user, params: application_form_params)
+    @application_form = ApplicationForm.new(application_form_params.merge(created_by_user: @user))
     begin
       @application_form.save!
-      @user = @application_form.user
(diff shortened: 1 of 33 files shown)


app/controllers/application_forms_controller.rb
```diff
@@ -9,7 +9,7 @@ def create
     begin
       @application_form.save!
       @user = @application_form.user
-      save_user_to_session!
+      save_user_to_session! unless SessionService.is_signed_in?(session)
```
benilovj: would it be worth covering this case with a spec that makes multiple form submissions within one session (which is a quite realistic scenario)?

PR #27, opened 2020-06-25
Comms to schools

benilovj:
### Context

We want to provide trusts and local authorities with suggested comms to explain the offer to their schools.

This page will be deep-linked to from the Notify comms that we send out.

### Guidance to review

preview


fofr:
I think this page needs to set some context about what it is.

fofr:
Context added:

Screen Shot 2020-06-25 at 16 30 42

benilovj:
I've reviewed 8ee9efb and it LGTM 👍

PR #32, opened 2020-06-29
Fix ability to modify bulk_update params & update recipients from other MNOs

aldavidson:
### Context

The pentest identified that a malicious user could modify the `recipient_ids` parameter in a bulk_update and modify the status of recipients for other MNOs

Fix the loophole that allowed users to do that.
Add tests for this scenario

### Guidance to review

From the root directory in a shell, `rspec spec/controllers/mno/recipients_controller_spec.rb --format doc`
You should see output similar to this:

```
Mno::RecipientsController
  PUT /bulk_update
    with recipient_ids from the same MNO as the user
      updates all the recipients
      redirects_to recipients index
    with some recipient_ids from another MNO
      updates the recipients from the same MNO as the user
      does not update the recipient from the other MNO
      redirects_to recipients index
    when the given status is not valid
      does not update the statusses
      responds with :unprocessable_entity

Finished in 0.46959 seconds (files took 3.1 seconds to load)
7 examples, 0 failures
```

### DIFF
--- a/app/models/recipient.rb
+++ b/app/models/recipient.rb
@@ -27,4 +27,8 @@ class Recipient < ApplicationRecord
   def self.from_approved_users
     joins(:created_by_user).where.not(users: { approved_at: nil })
   end
+
+  def self.visible_to_user(user)
+    from_approved_users.where(mobile_network_id: user.mobile_network_id)
+  end
 end
--- a/app/controllers/mno/recipients_controller.rb
+++ b/app/controllers/mno/recipients_controller.rb
@@ -1,7 +1,6 @@
 class Mno::RecipientsController < Mno::BaseController
   def index
-    @recipients = Recipient.from_approved_users
-                           .where(mobile_network_id: @mobile_network.id)
+    @recipients = Recipient.visible_to_user(@user)
                            .order(safe_order)
 
     respond_to do |format|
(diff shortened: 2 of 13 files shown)


app/models/recipient.rb
```diff
@@ -27,4 +27,8 @@ def self.exportable_attributes
   def self.from_approved_users
     joins(:created_by_user).where.not(users: { approved_at: nil })
   end
+
+  def self.visible_to_user(user)
+    from_approved_users.where(mobile_network_id: user.mobile_network_id)
+  end
```
benilovj, line 33: I think it'd be clearer if this method was:
```suggestion
  def self.on_mobile_network(mobile_network:)
    from_approved_users.where(mobile_network_id: mobile_network.id)
  end
```
so on the caller side it'd be `Recipient.on_mobile_network(@user.mobile_network)`, which would make it easier to grok first time, IMHO.

aldavidson: I see your point, but there's also the limitiation to approved-only - how about just:
```
def self.on_mobile_network(mobile_network)
   where(mobile_network_id: mobile_network.id)
 end
```

so then the call becomes:

`Recipient.from_approved_users.on_mobile_network(@user.mobile_network)`

?

benilovj: sounds good, :shipit:

PR #36, opened 2020-06-30
185 add script to get env of rails process

aldavidson:
### Context

Getting access to the rails console in the running container on PaaS is not obvious - this adds a helpful subshell script and instructions

Add a script to launch a subshell with the correct environment variables and in the correct directory
Add instructions in the README

### Guidance to review

This has been deployed to dev and staging
Follow the instructions in the README at https://github.com/DFE-Digital/get-help-with-tech/pull/36/files#diff-04c6e90faac2675aa89e2176d2eec7d8R114


aldavidson:
> Works well 👍
> 
> A couple of usability suggestions:
> 
> * I think the script relies on your `cf` command being attached to the same space that you're sshing into. It failed for me when I was connected to staging but tried sshing into dev. May be good to automate the step of connecting to the right space
> * it would be helpful to echo the commands that are in the README in the terminal, to make it easy to copy-paste. That way we wouldn't need to refer to the README

Good suggestions - done! Thanks @benilovj

PR #37, opened 2020-06-30
fix for pentest issue - user can view success pages for submissions from other users

aldavidson:
### Context

The pentest found that if a user could guess the ID of another users' form submission, they could see the success page for it, which exposed the submitting users' email address (details in the [Trello card](https://trello.com/c/HROXHQ6u/205-username-email-enumeration-issue) )

Fix the issue by including created_by_user_id criteria in the success page query
Tidy up "TODO" issue with correct parameter names
Add specs

### Guidance to review

1. Sign-in as a regular user
2. Submit one of each kind of form. Copy the URL of the success page (e.g. `/allocation_request_forms/2/success`)
3. Sign-in as a different user
4. Enter the URL of the previous user's success page into your browser's address bar
5. Before, you would be able to see that users' email address. Now, you should only see a 404

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,10 +5,10 @@ Rails.application.routes.draw do
   get '/pages/guidance', to: redirect('/about-bt-wifi')
 
   resources :application_forms do
-    get 'success/:application_form_id', to: 'application_forms#success', as: :success
+    get 'success/:recipient_id', on: :collection, to: 'application_forms#success', as: :success
   end
   resources :allocation_request_forms do
-    get 'success', to: 'allocation_request_forms#success'
+    get 'success/:allocation_request_id', on: :collection, to: 'allocation_request_forms#success', as: :success
   end
 
   resources :sessions, only: %i[create destroy]
--- a/app/controllers/allocation_request_forms_controller.rb
+++ b/app/controllers/allocation_request_forms_controller.rb
@@ -16,12 +16,10 @@ class AllocationRequestFormsController < ApplicationController
   end
 
   def success
-    @allocation_request = AllocationRequest.where(id: params[:allocation_request_id], created_by_user_id: @user.id).first
-    if @allocation_request
-      @allocation_request_form = AllocationRequestForm.new(allocation_request: @allocation_request)
-    else
-      render template: 'errors/not_found', status: :not_found
-    end
+    @allocation_request = @user.allocation_requests.find(params[:allocation_request_id])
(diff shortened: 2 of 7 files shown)


config/routes.rb
```diff
@@ -5,10 +5,10 @@
   get '/pages/guidance', to: redirect('/about-bt-wifi')
 
   resources :application_forms do
-    get 'success/:application_form_id', to: 'application_forms#success', as: :success
+    get 'success/:recipient_id', on: :collection, to: 'application_forms#success', as: :success
   end
   resources :allocation_request_forms do
-    get 'success', to: 'allocation_request_forms#success'
+    get 'success/:allocation_request_id', on: :collection, to: 'allocation_request_forms#success', as: :success
```
misaka, line 11: This can be simplified as:

```suggestion
    get :success
```

Which would generate the route:

```
allocation_request_form_success GET    /allocation_request_forms/:allocation_request_form_id/success(.:format)      allocation_request_forms#success
```

config/routes.rb
```diff
@@ -5,10 +5,10 @@
   get '/pages/guidance', to: redirect('/about-bt-wifi')
 
   resources :application_forms do
-    get 'success/:application_form_id', to: 'application_forms#success', as: :success
+    get 'success/:recipient_id', on: :collection, to: 'application_forms#success', as: :success
```
misaka, line 8: Minor nit, but in terms of modelling the resources it looks like the routes could be done as so:

```ruby
  resources :recipients, only: [] do
     resource :application_form, only: [] do
       get :success
     end
  end
```

Giving the route:

```
success_recipient_application_form GET    /recipients/:recipient_id/application_form/success(.:format)             application_forms#success
```

That seems more accurate, but I don't know if the path chosen, `/applications_forms/:recipient_id/success`, is intentional ... but it does seem strange at first glance to key application forms by the recipient id.

benilovj:
> I would probably be worth at least think about changing the routes, since that could affect bookmarks.

It's fine to change the routes, since this functionality is all behind feature flags on production.

config/routes.rb
aldavidson: Yeah, it's a hangover artefact from the first draft, which mixed in user details with recipient details. Since we moved to a simpler model of requiring sign-in, we arguably don't even need the _form abstraction any more.... but I want to wait until I've had chance to think about implementing the latest designs on the design prototype first before refactoring it out

aldavidson:
I think I'm going to keep the routes as they are for this PR - we can (and probably should) refactor them in a separate standalone PR

PR #38, opened 2020-06-30
fix CSV code injection vulnerability

aldavidson:
### Context

The pentest found it was possible to inject content into an exported CSV file via starting a user-inputted field with a =,+,-, or @ - the [CSV Injection vulnerability](https://owasp.org/www-community/attacks/CSV_Injection)

Fix the issue by prepending a single '.' to any field in an exported CSV that starts with one of those characters

### Guidance to review

1. Sign in as an approved user
2. Submit a 'Tell us about an eligible person' form for a mobile network to which you have a login, and make the account holder name or device phone number something like `'=(1 + 2)'`
3. Sign in as the mobile network user, and download the CSV file
4. Open the CSV file in the spreadsheet program of your choice (Google spreadsheets, Open Office, Libre Office, Numbers, Excel)
5. Make sure that the field which had started with an equals sign ('=') :
  a. starts with a .
  b. is not interpreted as a formula (i.e. does not not contain '3')

### DIFF
--- a/app/concerns/exportable_as_csv.rb
+++ b/app/concerns/exportable_as_csv.rb
@@ -17,9 +17,16 @@ module ExportableAsCsv
         csv << exportable_attributes.values
 
         all.each do |item|
-          csv << exportable_attributes.keys.map { |attr| item.send(attr) }
+          csv << exportable_attributes.keys.map do |attr|
+            safe_value = prevent_csv_code_injection(item.send(attr))
+            safe_value
+          end
         end
       end
+    end
+
+    def self.prevent_csv_code_injection(value)
+      value.to_s.starts_with?(/[\-+=@]/) ? value.gsub(/^([\-+=@].*)/, '.\1') : value
     end
   end
 end
--- a/app/concerns/exportable_as_csv.rb
+++ b/app/concerns/exportable_as_csv.rb
@@ -18,8 +18,7 @@ module ExportableAsCsv
 
         all.each do |item|
           csv << exportable_attributes.keys.map do |attr|
-            safe_value = prevent_csv_code_injection(item.send(attr))
-            safe_value
+            prevent_csv_code_injection(item.send(attr))
           end
         end
       end
--- a/spec/models/recipient_spec.rb
+++ b/spec/models/recipient_spec.rb
@@ -1,5 +1,41 @@
 require 'rails_helper'
 
 RSpec.describe Recipient, type: :model do
-  pending "add some examples to (or delete) #{__FILE__}"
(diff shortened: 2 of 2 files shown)


app/concerns/exportable_as_csv.rb
```diff
@@ -17,9 +17,16 @@ def self.to_csv
...
+          csv << exportable_attributes.keys.map do |attr|
+            safe_value = prevent_csv_code_injection(item.send(attr))
+            safe_value
+          end
         end
       end
     end
+
+    def self.prevent_csv_code_injection(value)
+      value.to_s.starts_with?(/[\-+=@]/) ? value.gsub(/^([\-+=@].*)/, '.\1') : value
```
misaka, line 29:
```suggestion
      value.to_s.starts_with?(/[\-+=@]/) ? ".#{value}" : value
```

Aside from just being a little hard to read, either you don't need the `to_s` or you'll run into an issue with negative values:

```ruby
[2] pry(main)> value = -1
=> -1
[3] pry(main)> value.to_s.starts_with?(/[\-+=@]/) ? value.gsub(/^([\-+=@].*)/, '.\1') : value
NoMethodError: undefined method `gsub' for -1:Integer
from (pry):3:in `<main>'
```

And since `CSV.generate` doesn't appears to differentiate between numbers and strings, may as well just always return a string if you need to prepend the `.`.

app/concerns/exportable_as_csv.rb
```diff
@@ -17,9 +17,16 @@ def self.to_csv
         csv << exportable_attributes.values
 
         all.each do |item|
-          csv << exportable_attributes.keys.map { |attr| item.send(attr) }
+          csv << exportable_attributes.keys.map do |attr|
+            safe_value = prevent_csv_code_injection(item.send(attr))
+            safe_value
```
misaka, line 22: Doesn't look like you need the intermediary variable `safe_value` here.

app/concerns/exportable_as_csv.rb
aldavidson: Good point. We do sometimes get nil values, which was the reasoning behind `to_s` - but of course, the `gsub` is redundant if we've already detected that it starts with a formula character.

aldavidson: there _shouldn't_ be any negative numbers in this data, so I'm OK with them getting an extra '.' prefix if they do occur.

app/concerns/exportable_as_csv.rb
aldavidson: also true :)

PR #39, opened 2020-06-30
Add responsible bodies

benilovj:
### Context

Responsible bodies (trusts and local authorities) will use the service.

- create an STI `ResponsibleBody` model, with `LocalAuthority` and `Trust` subclasses
- add a Rake task to fetch local authority data from the [Local authorities in England register](https://www.registers.service.gov.uk/registers/local-authority-eng)
- add a Rake task to fetch trusts from the [Get information about schools service](https://get-information-schools.service.gov.uk/)

### DIFF
--- a/lib/tasks/import_responsible_bodies.rake
+++ b/lib/tasks/import_responsible_bodies.rake
@@ -11,4 +11,19 @@ namespace :import do
         )
     end
   end
+
+  desc 'Import single and multi-academy trusts'
+  task trusts: :environment do
+    GetInformationAboutSchools.trusts_entries.each do |entry|
+      Trust
+        .where(companies_house_number: entry['Companies House Number'])
+        .first_or_create!(
+          name: entry['Group Name'],
+          organisation_type: entry['Group Type'],
+        )
+    end
+  end
+
+  desc 'Populate the responsible body reference data table'
+  task responsible_bodies: %i[local_authorities_in_england trusts]
 end
--- a/lib/tasks/import_responsible_bodies.rake
+++ b/lib/tasks/import_responsible_bodies.rake
@@ -1,27 +1,12 @@
 namespace :import do
   desc 'Import local authorities in England'
   task local_authorities_in_england: :environment do
-    LocalAuthoritiesInEnglandRegister.entries.each do |entry|
-      LocalAuthority
-        .where(local_authority_eng: entry['local-authority-eng'])
-        .first_or_create!(
-          name: entry['name'],
-          organisation_type: entry['local-authority-type'],
-          local_authority_official_name: entry['official-name'],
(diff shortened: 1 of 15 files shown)


spec/models/local_authorities_in_england_register_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'rails_helper'
+
+RSpec.describe LocalAuthoritiesInEnglandRegister, type: :model do
+  it 'returns a list of entries' do
+    stub_request(:get, LocalAuthoritiesInEnglandRegister::URL)
+      .to_return(body: '{"BRD":{"index-entry-number":"11","entry-number":"11","entry-timestamp":"2016-10-21T16:11:20Z","key":"BRD","item":[{"local-authority-type":"MD","official-name":"City of Bradford Metropolitan District Council","local-authority-eng":"BRD","name":"Bradford","start-date":"1974-04-01"}]},"TEI":{"index-entry-number":"128","entry-number":"128","entry-timestamp":"2016-10-21T16:11:20Z","key":"TEI","item":[{"local-authority-type":"NMD","official-name":"Teignbridge District Council","local-authority-eng":"TEI","name":"Teignbridge"}]}}')
```
misaka, line 6: I dunno, you don't want to make this too easy to read do you? 😛

lib/tasks/import_responsible_bodies.rake
```diff
@@ -0,0 +1,29 @@
...
+  task trusts: :environment do
+    GetInformationAboutSchools.trusts_entries.each do |entry|
+      Trust
+        .where(companies_house_number: entry['Companies House Number'])
+        .first_or_create!(
+          name: entry['Group Name'],
+          organisation_type: entry['Group Type'],
+        )
+    end
+  end
```
misaka, line 25: These aren't tested because, well ... rake task. How about moving the code here into a service to make it testable?

misaka: Realised after I wrote that this is likely a one-off. Is that true or will this need to be re-run periodically?

aldavidson: this will be a one-off, I'm pretty sure 

_(famous last words)_

benilovj: Yes, we currently don't expect to run this more than once

benilovj: @misaka I've extracted the service as suggested in 253bb9d

spec/models/local_authorities_in_england_register_spec.rb
benilovj: Fair - fixed in 292dbaa

PR #40, opened 2020-06-30
fix mis-typed permisisions

misaka:
### Context

You guys think your so fancy with our pen-test and STI PRs ... well here's one for ya! ;)

Fix typo.


aldavidson:
Nice catch @misaka ! Adding value already :) 👍  💯

PR #135, opened 2020-07-20
Remove Rails UJS since we don't use it

benilovj:
### Context

We don't need/want to use Rails UJS, and its magic has bitten me once already during development.

- Remove Rails UJS.
- Change the sign out link to use GET instead of DELETE.

### DIFF
--- a/app/views/layouts/_nav.html.erb
+++ b/app/views/layouts/_nav.html.erb
@@ -13,9 +13,9 @@
   <%- end %>
 
   <%- if SessionService.is_signed_in?(session) %>
-    <%- sign_out_link = link_to("Sign out", session_path(id: session[:session_id]), method: :delete, class: 'govuk-header__link' ) %>
+    <%- sign_out_link = link_to("Sign out", session_sign_out_path(session_id: session[:session_id]), class: 'govuk-header__link' ) %>
     <%- content = [ @user.email_address, sign_out_link].join(' | ') %>
-    <%= nav_item(url: session_path(id: session.id), content: content.html_safe) %>
+    <%= nav_item(url: sessions_path(id: session.id), content: content.html_safe) %>
   <%- else %>
       <%= nav_item( title: "Sign in", url: sign_in_path )  %>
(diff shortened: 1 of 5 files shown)


tijmenb:
I suspect this might break the logout link, I think `method` only works with UJS: https://github.com/DFE-Digital/get-help-with-tech/blob/3ab357fdbac8be652f00751d5c9210c070cbaaae/app/views/layouts/_nav.html.erb#L16

benilovj:
> I suspect this might break the logout link, I think `method` only works with UJS:
> 
> https://github.com/DFE-Digital/get-help-with-tech/blob/3ab357fdbac8be652f00751d5c9210c070cbaaae/app/views/layouts/_nav.html.erb#L16

Ah good catch - thanks.

fofr:
Can confirm that with JS disabled our current "Sign out" link doesn't work.

benilovj:
> Can confirm that with JS disabled our current "Sign out" link doesn't work.

I've added a2e05ff which should fix the issue – @fofr you mind giving it another go using this branch?

fofr:
@benilovj Sign out works with JS disabled on this branch.

aldavidson:
I'm in favour of not relying on Javascript, but I'm not sure about the sign out link becoming a regular GET - it will work, but it breaks the RESTfulness model and introduces some potential issues (for instance, it's then possible for the users' browser to follow the link pre-emptively and sign the user out without them knowing).

I'm going to try an alternative, will add it as a suggested change if it works - please hold off merging this for the moment

benilovj:
> I'm going to try an alternative, will add it as a suggested change if it works - please hold off merging this for the moment

Sure. FWIW, I looked to Apply for Teacher Training for inspiration...

aldavidson:
I've put my alternative up as a separate PR - #137 . It solves the problems of signout-as-GET, at the expense of a few lines of CSS.

benilovj:
Closing since #137 has been merged instead.

PR #137, opened 2020-07-20
Remove rails ujs but keep signout link as delete

aldavidson:
### Context

Suggested alternative to #135 - please merge EITHER this OR #135.

* Change the "sign out" link to a form with `method="delete"`, but styled like the rest of the nav links.

That way we can:
* remove the last reliance on Rails UJS
* keep the routes RESTful
* avoid exposing an "unsafe" action as a GET
* still have a DELETE route that works without Javscript enabled.

For some more context, see [this discussion on Stack Overflow](https://stackoverflow.com/a/14587231) and the [HTTP 1.1 standard on "Safe Methods" (2nd para)](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.1.1)


benilovj:
I don't really have any opinions on this over #135 – @fofr is probably best-placed to weigh in on it, given his frontend background.

PR #138, opened 2020-07-20
Render markdown with ActionView::Template::Handler

benilovj:
### Context

Before this PR, the devices multi-page guidance is written in Markdown. The Markdown was being rendered in the `MarkdownPoweredPage` class using plain old Ruby, and was coupled with the handling of multi-part guides.

This meant that the generated HTML had to be passed through the controller into the views via view variables, and the views themselves are currently markdown-aware. This also meant it is harder to use Markdown in other parts of the app (e.g. for other guidance).

- Add a Rails view template handler that takes `.md` files and renders them using the Rails view mechanisms.
- Simplify the code that handles the multi-part devices guide. This should make it easier to reuse the code for the other multi-part guides in the app.


benilovj:
This PR triggered a Brakeman warning. I've ignored it (cd204d4) because I think it's a false-positive, but it's worth reviewing as part of the PR.

app/views/layouts/multipage_guide.html.erb
```diff
@@ -0,0 +1,39 @@
+<% content_for :devices_service, true %>
+<%= content_for :title, @page.title %>
```
fofr, line 2:
```suggestion
<% content_for :title, @page.title %>
```

app/views/layouts/multipage_guide.html.erb
```diff
@@ -0,0 +1,39 @@
...
+  <ol class="govuk-breadcrumbs__list">
+    <li class="govuk-breadcrumbs__list-item">
+      <%= link_to t('page_titles.devices_guidance_index'), devices_guidance_index_path, class: 'govuk-breadcrumbs__link' %>
+    </li>
+    <li class="govuk-breadcrumbs__list-item">
+      <%= @page.title %>
+    </li>
+  </ol>
+  </div>
+<% end %>
```
fofr, line 15:
```suggestion
<%- content_for :before_content do %>
  <div class="govuk-breadcrumbs">
    <ol class="govuk-breadcrumbs__list">
      <li class="govuk-breadcrumbs__list-item">
        <%= link_to t('page_titles.devices_guidance_index'), devices_guidance_index_path, class: 'govuk-breadcrumbs__link' %>
      </li>
      <li class="govuk-breadcrumbs__list-item">
        <%= @page.title %>
      </li>
    </ol>
  </div>
<% end %>
```

lib/action_view/template/handlers/markdown.rb
```diff
@@ -0,0 +1,29 @@
...
+      def rendered_table_of_contents_for(source)
+        @table_of_contents = Redcarpet::Markdown.new(Redcarpet::Render::HTML_TOC.new(nesting_level: 2))
+        @table_of_contents
+          .render(source)
+          .html_safe
+      end
+
+      # this makes it easier to target the generated code with CSS
+      def wrap_in_a_div(source)
+        "<div class=\"app-styled-content\">#{source}</div>"
```
fofr, line 25: One for later probably, but: You can add a custom RedCarpet renderer to add the required classes to generated markdown, meaning you wouldn't need the app-styled-content wrapper – that's how Find and Apply are doing it.

benilovj: Yep, I guess the trade-off here is whether we want to carry extra Ruby code to generate the right classes (which will add up to quite a few LOC to cover all HTML elements), or whether we want to carry extra styles to style the generated markdown appropriately. Not sure which is better...

lib/action_view/template/handlers/markdown.rb
```diff
@@ -0,0 +1,29 @@
+module ActionView
+  module Template::Handlers
+    class Markdown
+      def call(_template, source)
+        @markdown ||= Redcarpet::Markdown.new(Redcarpet::Render::HTML.new(with_toc_data: true))
+        "
+        content_for :html_list_of_headings_links do
+          #{wrap_in_a_div(rendered_table_of_contents_for(source)).inspect}.html_safe
```
aldavidson, line 8: This line looks really awkward and fragile..... but I admit I'm struggling to find a better alternative that works :/

benilovj: Yeah it's awkward because it's generating ruby code. I've tried to provide lots of spec coverage to ensure it keeps working.

lib/action_view/template/handlers/markdown.rb
aldavidson: I think this is probably the right tradeoff for now. We can refactor later if needed.

config/brakeman.ignore
```diff
@@ -0,0 +1,26 @@
+{
+  "ignored_warnings": [
+    {
+      "warning_type": "Dynamic Render Path",
+      "warning_code": 15,
+      "fingerprint": "aa162721dfe63daa1c04c888b19e85ba75855583622c20572118575e01c9426b",
+      "check_name": "Render",
+      "message": "Render path contains parameter value",
+      "file": "app/controllers/devices_guidance_controller.rb",
```
aldavidson, line 9: I'm a little uneasy about disabling brakeman warnings, but I think in this case, given that it only allows the subpage IDs that are present as keys in the particular section of the i18n file, there is sufficient protection against code injection.

app/controllers/devices_guidance_controller.rb
```diff
@@ -1,4 +1,6 @@
 class DevicesGuidanceController < ApplicationController
+  include Rails.application.routes.url_helpers
```
aldavidson, line 2: is this line redundant? I've run the specs  & clicked through with it commented-out, and it all seems to work

config/brakeman.ignore
aldavidson: Might be an idea to add a feature spec to `spec/features/devices_guidance_spec.rb` that tests that scenario (requesting a subpage that does not exist responds with a 404)?

app/controllers/devices_guidance_controller.rb
benilovj: Yes, you're right - it's not needed. I've removed it.

config/brakeman.ignore
benilovj: Agreed – added in ad4f9b9

PR #140, opened 2020-07-21
Make the Too Many Requests error be a statically-pre-rendered page

aldavidson:
### Context

As the rate-limiting is done in Rack middleware, it does not have access to the regular Rails rendering pipeline - so when the rate limit is hit the default behaviour is to just return the string 'Too many requests' as `text/plain` content. 

User testing showed that when this limit was hit, the raw nature of the error was jarring and confusing. The user did not know what to do and said they would contact support. A consistently-branded error message would help reassure the user that they just needed to wait a few seconds and try again.

* add a rake task to pre-render a govuk-styled 'Too many requests' error page into the `public/` folder
* call the rake task as part of the Docker build
* add `Rack::Throttle` config to read that file on startup, pass the contents to the middleware as `:message`, and serve it as `text/html`

### Guidance to review

Run the rake task with `rake release:render_429_to_file`
With the `rate_limiting` feature flag active (e.g. `FEATURES_rate_limiting=active bundle exec rails s`) , click `sign in`
Once you've entered your email address and got the 'Check your email' page, refresh it at least 4 times in quick succession.
You should see a page like this:
Screen Shot 2020-07-21 at 13 25 43

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,8 +26,10 @@ USER 1001
 
 # install all gems
 COPY --chown=deploy:deploy Gemfile Gemfile.lock .ruby-version ./
-ARG BUNDLE_FLAGS="--jobs 2 --no-cache --without development test"
-RUN bundle install ${BUNDLE_FLAGS}
+ARG BUNDLE_FLAGS="--jobs 2"
+RUN bundle config set no-cache 'true'
+RUN bundle config set without 'development test'
+RUN bundle install
 RUN bundle package --all
 
 COPY --chown=deploy:deploy . .
@@ -40,6 +42,9 @@ EXPOSE $APP_PORT
 RUN yarn
 ARG RAILS_ENV=production
 RUN RAILS_ENV=${RAILS_ENV} SECRET_KEY_BASE=$(bin/rake secret) bundle exec rails webpacker:compile
+
+# Render the 'too many requests' error page
+RUN RAILS_ENV=${RAILS_ENV} SECRET_KEY_BASE=$(bin/rake secret) bundle exec rake release:render_429_to_file
 
 # Cache the git commit sha & branch
 ARG GIT_COMMIT_SHA=""
--- a/lib/tasks/release.rake
+++ b/lib/tasks/release.rake
@@ -1,0 +1,15 @@
+namespace :release do
+  desc 'Render 429.html error page to the public folder'
+  task :render_429_to_file do
+    Rails.application.initialize!
+    renderer = ApplicationController.renderer.new(
+      http_host: Settings.hostname_for_urls,
+    )
+    content = renderer.render(template: 'errors/too_many_requests')
+    path = File.join(Rails.root, 'public', '429.html')
+    File.open(path, 'w+') do |f|
+      f << content
+    end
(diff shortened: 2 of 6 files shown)


benilovj:
Not sure about the "Too many requests" wording – should we look to use something from https://design-system.service.gov.uk/patterns/service-unavailable-pages/ ?

aldavidson:
> Not sure about the "Too many requests" wording – should we look to use something from https://design-system.service.gov.uk/patterns/service-unavailable-pages/ ?

The 'Too many requests' wording could totally be better, I agree.

I did look at the 'Service unavailable' guidance, but I don't think it quite fits this case.  Maybe an introductory line would help, something saying why we've limited the number of requests in the first place... one for the content designers to have a look at?

How about we put this in as-is, as it's certainly an improvement on the plain text error that we get at the moment, and add a card for one of the content designers to look at this wording?

lib/tasks/release.rake
```diff
@@ -0,0 +1,15 @@
...
+    Rails.application.initialize!
+    renderer = ApplicationController.renderer.new(
+      http_host: Settings.hostname_for_urls,
+    )
+    content = renderer.render(template: 'errors/too_many_requests')
+    path = File.join(Rails.root, 'public', '429.html')
+    File.open(path, 'w+') do |f|
+      f << content
+    end
+    puts ['Rendered ', content.length, 'bytes to ', path].join(' ')
```
benilovj, line 13:
```suggestion
    puts "Rendered #{content.length} bytes to #{path}"
```

Dockerfile
```diff
@@ -26,8 +26,10 @@ USER 1001
 
 # install all gems
 COPY --chown=deploy:deploy Gemfile Gemfile.lock .ruby-version ./
-ARG BUNDLE_FLAGS="--jobs 2 --no-cache --without development test"
-RUN bundle install ${BUNDLE_FLAGS}
+ARG BUNDLE_FLAGS="--jobs 2"
+RUN bundle config set no-cache 'true'
+RUN bundle config set without 'development test'
```
benilovj, line 31: are these related to this PR?

aldavidson:
> Just wondering why we're checking in `public/429.html` into the repo, if we're generating it. Is it so the middleware works in development?

Yes

Dockerfile
aldavidson: they just avoid a deprecation warning in the build - not strictly essential

aldavidson:
> I think this change in itself is an improvement. However
> 
> 1. I think the content design could use another pass (as mentioned already).
Agreed - I've put in a [Trello card](https://trello.com/c/BElpsjYS/323-improve-content-on-too-many-requests-error-page) for some content review on that.

> 2. I'm still slightly puzzled as to how the user was able to see this through legitimate usage – are our limits too low?
Quite possibly - we have a default rule of "4 GETs per IP per second" for most things, but that drops to 1 if it's the sign-in token controller to be paranoid about brute-forcing prevention. Happy to raise it (....suggestions about what's a better limit welcome! 4 across the board?)

benilovj:
> suggestions about what's a better limit welcome! 4 across the board?

I'd be ok with that limit

PR #144, opened 2020-07-22
Fix issues with magic-links and existing sessions

tonyheadford:
### Context
[Trello card](https://trello.com/c/m2ZggEBi/311-bug-fix-error-when-using-magic-link-if-already-signed-in)
If a user attempts to reuse their magic-link to access the application while there is still a current session an error is generated.

#### Current process causing failure
```
-> User signs-in
<- System generates a new magic link
-> User uses magic-link
-> User closes browser or opens new tab but does not click 'Sign out'
-> User reuses magic-link
<- System does not recognise the magic-link token and prompts user to sign-in again
-> User signs-in
<- System generates new magic link
-> User uses the magic-link
<- System generates an error
```

Only create a new session if the current user is not currently signed in.

### Guidance to review
Conscious I may not know the full expected lifecycle of sessions and simply not creating one if one already exists might not be ideal in all cases?


spec/features/session_behaviour_spec.rb
```diff
@@ -69,6 +69,26 @@
...
+      visit validate_token_url
+      expect(page).to have_current_path(download_responsible_body_bt_wifi_vouchers_path)
+    end
+
+    scenario 'using a new magic link redirects to the home page for user' do
+      sign_in_as(user)
+      expect(page).to have_current_path(download_responsible_body_bt_wifi_vouchers_path)
+    end
+  end
+
```
benilovj, line 91: I'd suggest we code the test against the `extra_mobile_data_offer` being enabled, since we're likely to remove it (and leave the functionality permanently on) within the week.

tonyheadford: OK thanks, that makes sense. I deactivated that feature as it was the first way I found that I could get a path I could test against.

aldavidson: ^ what @benilovj said, otherwise fine 👍

benilovj:
Looking good. Would be nice to rebase the merge commits out, so we have a clean commit history, but otherwise go ahead and merge.

PR #145, opened 2020-07-22
updated mobile phone number hint and error message

aldavidson:
### Context

As in the [Trello Card](https://trello.com/c/JISmia3v/248-review-mobile-phone-number-hint-error-message-wording), the wording around entering a mobile phone number could be better.

Apply updates from content designer 

Before:

Screen Shot 2020-07-22 at 14 23 06

After:

Screen Shot 2020-07-22 at 14 16 56


aldavidson:
> 👍 . Is this change an opportunity to move the 2 hint text lines to the `en.yml` file?

Yyyyyes, but I'm not going to take it - I'm going to add another trello card to move _all_ the hints and legends into `en.yml`

PR #149, opened 2020-07-23
Interstitial for magic links

benilovj:
### Context

Some responsible bodies appear to have link checkers that follow email links. This causes a problem for the single-use magic links that we use to authenticate the users of our service, because:

1. The link checker follows the link, thus invalidating the token
2. When the user tries to follow the magic link for real, they're told that their token is not recognised and they need a new magic link.
3. Rinse and repeat.

Part of the problem comes from having a destructive operation (token invalidation) happen on a GET.

Change the sign-in flow so that:

- The magic link takes users to an interstitial page:

screencapture-localhost-3000-token-validate-2020-07-23-16_26_41

- At this point, the user is signed-in but the token is still valid (as discussed on Slack)
- As soon as the user clicks 'Continue', their token is invalidated. We are assuming that link checker bots won't be hitting the button.

### Guidance to review

- I've done some clean-up of the `SignInTokensController`, to move more logic into `SessionService`.
- Added a controller spec to cover an edge case that is too granular for a feature spec

### DIFF
--- a/app/controllers/sign_in_tokens_controller.rb
+++ b/app/controllers/sign_in_tokens_controller.rb
@@ -1,4 +1,6 @@
 class SignInTokensController < ApplicationController
+  before_action :require_sign_in!, only: :destroy
+
   def new
     @sign_in_token_form ||= SignInTokenForm.new
     if FeatureFlag.active?(:public_account_creation)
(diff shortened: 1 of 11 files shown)


benilovj:
> Looks neat. Should we only sign-in the user in the `destroy` method? It looks like you wouldn't need to click the button as you'd already be signed in at that point?
> Actually I might not have followed the flow right from the diff

You've understood correctly – users wouldn't strictly need to click the Continue button but we'd expect almost everyone would...

benilovj:
I think this is ready for a proper review now.

PR #150, opened 2020-07-23
add Makefile task to setup the cdn-route to forward all headers

aldavidson:
### Context

See [Trello card 338](https://trello.com/c/q52ePQep/338-we-are-not-seeing-valid-user-agent-strings-in-logit) - while investigating issues with sign-in tokens, we found that all user-agent fields in Logit were empty. Turns out the default set up for the  cdn-route  service (which is how you get custom domains to work in Gov.UK PaaS) is to only forward the Host header to your app: https://docs.cloud.service.gov.uk/deploying_services/use_a_custom_domain/#forwarding-headers

Add a Makefile task to tell the cdn-route service to forward all headers.

### Guidance to review

The change is already made on prod, this is more just a record of what I did, so that it can be done again in future if needed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,10 @@ setup_paas_db: set_cf_target
 setup_paas_app: set_cf_target
 	cf scale $(APP_NAME)-$(env_stub) -k 2G
 
+setup_cdn_route: set_cf_target
+	# tell it to forward all headers from Cloudfront, otherwise we only get Host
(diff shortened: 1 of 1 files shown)


aldavidson:
> Looks good, as Jake said would be good to add this type of thing to the README

Sure, although TBH I should probably make that a separate task to document all the setup_ tasks in the Makefile (which were only ever run-one-time things like `setup_paas_db`) in the README. I'll add that to Trello

PR #154, opened 2020-07-24
Added notifying extra mobile data account holder by sms

tonyheadford:
### Context
[Trello card](https://trello.com/c/nNKhfK4c/305-text-message-for-all-mno-extra-mobile-data-requests)

Addition of sending an sms message to an account holder on creation of a extra mobile data request via GOV.UK Notify service.  One of two messages can be sent, depending on whether the chosen mno is participating in the data scheme or not.

### Guidance to review
Happy path only. Currently not persisting the response Id from Notify and no implementation of a background process to verify success or failure of delivery.

### DIFF
--- a/app/controllers/responsible_body/extra_mobile_data_requests_controller.rb
+++ b/app/controllers/responsible_body/extra_mobile_data_requests_controller.rb
@@ -25,7 +25,8 @@ class ResponsibleBody::ExtraMobileDataRequestsController < ResponsibleBody::Base
         session.delete(:extra_mobile_data_request_params)
         @extra_mobile_data_request.save!
 
-        NotifyExtraMobileDataAccountHolderService.call(@extra_mobile_data_request)
+        sms_service.deliver!(@extra_mobile_data_request)
+        # NotifyExtraMobileDataAccountHolderService.call(@extra_mobile_data_request)
 
         flash[:success] = I18n.t('responsible_body.extra_mobile_data_requests.create.success')
         redirect_to responsible_body_extra_mobile_data_requests_path
@@ -59,5 +60,9 @@ private
       mobile_network_id
       confirm
     ])
+  end
+
+  def sms_service
+    @sms_service ||= NotifyExtraMobileDataAccountHolderService.new
   end
 end
--- a/app/controllers/responsible_body/extra_mobile_data_requests_controller.rb
+++ b/app/controllers/responsible_body/extra_mobile_data_requests_controller.rb
@@ -24,6 +24,9 @@ class ResponsibleBody::ExtraMobileDataRequestsController < ResponsibleBody::Base
         # clear the stashed params once the user has confirmed them
         session.delete(:extra_mobile_data_request_params)
         @extra_mobile_data_request.save!
+
+        NotifyExtraMobileDataAccountHolderService.call(@extra_mobile_data_request)
+
         flash[:success] = I18n.t('responsible_body.extra_mobile_data_requests.create.success')
         redirect_to responsible_body_extra_mobile_data_requests_path
       else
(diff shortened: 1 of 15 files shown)


app/services/notify_extra_mobile_data_account_holder_service.rb
```diff
@@ -0,0 +1,56 @@
+class NotifyExtraMobileDataAccountHolderService
```
benilovj, line 1: (The following suggestion is just personal taste and comes from a general dislike of service objects, so take with a pinch of 🧂 )

This class could be made more object-oriented by renaming it to something like `ExtraMobileDataAccountHolderNotification`.  A `deliver!` method on that class would do the sending. You could construct it off the `ExtraMobileDataRequest` class:

```
extra_mobile_data_request.acount_holder_notification.deliver!
```

app/controllers/responsible_body/extra_mobile_data_requests_controller.rb
```diff
@@ -26,6 +26,9 @@ def create
         # clear the stashed params once the user has confirmed them
         session.delete(:extra_mobile_data_request_params)
         @extra_mobile_data_request.save!
+
+        NotifyExtraMobileDataAccountHolderService.call(@extra_mobile_data_request)
+
```
benilovj, line 31: Probably ok to keep these here for now, but I wonder whether the request saving and notification sending should live in a separate service, and not in the controller...

spec/services/notify_extra_mobile_data_account_holder_service_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+          personalisation: {
+            mno: request.mobile_network.brand,
+          },
+        },
+      )
+      subject.call
+    end
+  end
+
+  context 'an error with Notify' do
```
aldavidson, line 52: minor stylistic comment - can this read something more like `context 'when the Notify client raises an error' do` ?

That way when you run the specs with `--format=doc` you get a more human-readable flow, like:

```
NotifyExtraMobileDataAccountHolderService
  when the Notify client raises an error
    catches the error and raises a NotifySmsError
```

app/services/notify_extra_mobile_data_account_holder_service.rb
aldavidson: I second that 👍

tonyheadford: Not implemented it this as yet, although I do like it. Wanted to get it all working and inline with the rubocop rules as they are setup.

spec/services/notify_extra_mobile_data_account_holder_service_spec.rb
tonyheadford: Yes, agree. I've changed this to suit.

app/services/notify_extra_mobile_data_account_holder_service.rb
tonyheadford: Perhaps we should also consider backgrounding the call to Notify once @aldavidson has established the approach

aldavidson: yep, totally -  but don't let that hold up this PR. Once my PR has been merged, then we can do separate smaller PRs for porting individual tasks over to jobs

app/services/notify_extra_mobile_data_account_holder_service.rb
```diff
@@ -0,0 +1,58 @@
...
+  class NotifySmsError < StandardError; end
+
+  # def self.call(extra_mobile_data_request)
+  #   new(extra_mobile_data_request).call
+  # end
+
+  # def initialize(extra_mobile_data_request)
+  #   @extra_mobile_data_request = extra_mobile_data_request
+  # end
```
benilovj, line 10: I think it'd be better to get rid of the commented-out code...

tonyheadford: Good spot! I agree on the renaming, perhaps we could look at a bit of refactoring when this gets moved into background jobs along the lines you mention above so we can merge this in?

benilovj: 👍

PR #171, opened 2020-07-30
Markdown rendering improvements

benilovj:
### Context

This service contains a lot of guidance, and has a few content designers who are (or will be) working on it.

Some of the pages are already authored using Markdown, e.g. the devices guidance. However, the current approach is limited by inability to use Rails partials (e.g. for shared snippets) or do any kind of Ruby interpolation (e.g. to avoid hard-coding URLs all over the place).

1. Allow using ERB interpolation within Markdown documents. For example, this becomes possible now:

```markdown
### Eligibility

<%= render partial: 'shared/bt_wifi_eligibility' %>

### Who can help
...
```

2. Reduce the need for custom CSS by introducing a custom Markdown renderer that spits out GOV.UK-flavoured HTML. This means that the produced HTML is compatible with the GOV.UK Design System and uses its styles by default.

### Guidance to review

There are still a couple of bespoke styles relating to Markdown content:

* for images, which is a deliberate app-specific design choice
* for the table of contents – the entire `HTML_TOC` renderer would have needed to be re-implemented from scratch (in order to add the GOV.UK element classes in the right place), which felt like a lot of extra code to maintain for fairly minimal benefit.

Also worth noting: the custom GOV.UK renderer does not re-implement every possible element, but the most common ones that need extra classes that one comes across in Markdown. We should watch for missing styling when converting pages to Markdown, in case something has been missed.

The `ActionView::Template::Handlers::Markdown` class is still a bit horrible and brittle. Some of this is necessary complexity owing to how Rails template handlers work, but I'm open to suggestions on how to improve it. I tried to maintain good spec coverage, to prevent regressions for anyone brave enough to try to change that code at a later date 🤣 .

### DIFF
--- a/lib/action_view/template/handlers/markdown.rb
+++ b/lib/action_view/template/handlers/markdown.rb
@@ -1,3 +1,5 @@
+require 'govuk/markdown_renderer'
+
 module ActionView
   module Template::Handlers
     class Markdown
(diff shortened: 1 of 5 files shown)


lib/action_view/template/handlers/markdown.rb
```diff
@@ -6,17 +8,15 @@ def call(template, source)
 
         <<~CODE
           content_for :html_list_of_headings_links do
-            ('<div class="app-styled-content">' +
+            ('<div class="table-of-contents">' +
```
fofr, line 11:
```suggestion
            ('<div class="app-table-of-contents">' +
```

benilovj: Thanks – updated

PR #174, opened 2020-07-31
215 add docker image id to the healthcheck endpoint

aldavidson:
### Context

The healthcheck endpoint allows us to see what Git SHA was used to build the app, but not the actual image id of the Docker container. Adding this allows us to diagnose any issues more quickly by isolating exactly which version is where.

In the Makefile, extend the `deploy` task to add an extra post-push step for setting the DOCKER_IMAGE_ID as an env var. We must do this _after_ a successful push, not before.

In the github action, install `cf-cli` from a binary distribution rather than apt-get - cuts down execution time by about 45s

Add docker -> image_id to the info element in the `/healthcheck.json` endpoint

### Guidance to review

`make dev deploy` - then hit `/healthcheck.json`

### DIFF
--- a/.github/workflows/build-push-and-deploy-container.yml
+++ b/.github/workflows/build-push-and-deploy-container.yml
@@ -51,9 +51,8 @@ jobs:
         id: get_docker_image_id
         shell: bash
         run: |
-          # docker pull dfedigital/get-help-with-tech-dev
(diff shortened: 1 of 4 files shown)


aldavidson:
> Looks good. The commit history is very noisy – can I suggest you squash everything into one commit @aldavidson ?

Yep, absolutely - whenever we're changing Github actions, the only realistic way to test them is to set them to run on push to the branch you're on, and .... push. So the commit history gets v. noisy. Of course I will squash before merging.

PR #187, opened 2020-08-03
Update en.yml

gemmaemoore:
devices - phase 2

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -31,37 +31,22 @@ en:
   devices_guidance:
     about_the_offer:
       title: About the offer
-      description: See who is eligible for devices and why we’re providing them.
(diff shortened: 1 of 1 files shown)


benilovj:
@gemmaemoore I tweaked your changes because some of the indentation was mis-aligned and the file is format sensitive to the number of indented spaces.

Separate, I see that you've renamed the "Choosing devices" page to "Allocation and specification". As it currently stands, that won't automatically update the slug, which will still be `/devices/choosing-devices`. Is this desired, or would we want to replace the page title everywhere and put a redirect in place?

gemmaemoore:
@benilovj thanks Jake! We have redirects to /choosing-devices from Computacenter so I don't mind the slug staying the same but if we can have another redirect on top of that then it could be updated .

benilovj:
Closing in favour of #192.

PR #196, opened 2020-08-05
Responsible bodies in support

benilovj:
### Context

It's useful to see a list of responsible bodies in the support area of the service, so that:

- DfE ops users can see engagement stats around both pilots
- There is a jump-off points for further operations on the responsible bodies (e.g. managing users)

- Add a list of responsible bodies in the support area, only accessible to DfE staff
- Display some basic engagement stats per responsible body

### Guidance to review

Is the table layout appropriate?

* pros: can be copied directly into Excel for further manipulation
* cons: not very compact or scannable

image

Does the microcopy make sense?

(I intend to add the extra mobile data requests later when the data model has been improved for reporting.)

### DIFF
--- a/app/controllers/support/responsible_bodies_controller.rb
+++ b/app/controllers/support/responsible_bodies_controller.rb
@@ -1,0 +1,5 @@
+class Support::ResponsibleBodiesController < Support::BaseController
+  def index
+    @responsible_bodies = ResponsibleBody.joins(:users).distinct.order('name asc')
+  end
+end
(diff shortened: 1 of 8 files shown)


tonyheadford:
Looks like a great idea. One thought would be does the additional text in the table make it a pain for users wanting to use this in Excel? i.e `1 user` and `2 users signed in` instead of plain `1` and `2` ?

benilovj:
> Looks like a great idea. One thought would be does the additional text in the table make it a pain for users wanting to use this in Excel? i.e `1 user` and `2 users signed in` instead of plain `1` and `2` ?

Yep, you're right – I'll do a follow-up PR to properly separate things into a CSV and an HTML representation, and add a big green CSV button.

PR #202, opened 2020-08-06
330 add unavailable status

tonyheadford:
### Context
[Trello card](https://trello.com/c/Tx9RTGFF/330-add-unavailable-status)
This change adds `unavailable` to the status enum for the `ExtraMobileDataRequest` model. The status should be set to `unavailable` if the mobile network is not yet participating.  This status should be displayed in the list of requests for the user.
There is also altered messaging during submission to indicate that the network is not yet in the service and that the request has been saved rather than submitted.

### DIFF
--- a/app/controllers/responsible_body/mobile/manual_requests_controller.rb
+++ b/app/controllers/responsible_body/mobile/manual_requests_controller.rb
@@ -23,16 +23,16 @@ class ResponsibleBody::Mobile::ManualRequestsController < ResponsibleBody::BaseC
       if params[:confirm]
         # clear the stashed params once the user has confirmed them
         session.delete(:extra_mobile_data_request_params)
-        @extra_mobile_data_request.save!
 
-        @extra_mobile_data_request.notify_account_holder_later
+        @extra_mobile_data_request.save_and_notify_account_holder!
 
-        flash[:success] = I18n.t('responsible_body.extra_mobile_data_requests.create.success')
+        flash[:success] = build_success_message(@extra_mobile_data_request.mobile_network.participating?)
         redirect_to responsible_body_mobile_extra_data_requests_path
       else
         # store given params in session,so that we don't have to pass them back in the URL
         # if the user clicks 'Change' on the confirmation page
         session[:extra_mobile_data_request_params] = extra_mobile_data_request_params
+        @extra_mobile_data_request = presenter(@extra_mobile_data_request)
         render :confirm
       end
     else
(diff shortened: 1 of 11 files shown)


app/controllers/responsible_body/mobile/manual_requests_controller.rb
```diff
@@ -60,4 +60,16 @@ def extra_mobile_data_request_params
       confirm
     ])
   end
+
+  def build_success_message(mno_is_participating)
+    if mno_is_participating
+      I18n.t('responsible_body.extra_mobile_data_requests.create.success')
+    else
+      I18n.t('responsible_body.extra_mobile_data_requests.create.success_saved')
+    end
```
aldavidson, line 69: (non-blocking suggestion) do you think it might make more sense to rename these keys to something more aligned with the criteria under which they'll be used, rather than the content of the message, like `....success.participating_mno` and `...success.non_participating_mno` ?

spec/features/responsible_body/submitting_a_spreadsheet_request_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    before do
+      mobile_network
+      sign_in_as user
+      # prevent api call to Notify
+      ActiveJob::Base.queue_adapter = :test
+    end
+
+    after do
+      ActiveJob::Base.queue_adapter = :inline
+    end
```
aldavidson, line 30:
```suggestion
    before do
      mobile_network
      sign_in_as user
    end
  
    around do
      # prevent api call to Notify
      original_queue_adapter = ActiveJob::Base.queue_adapter
      ActiveJob::Base.queue_adapter = :test
      example.run
      ActiveJob::Base.queue_adapter = original_queue_adapter
    end
```

spec/models/extra_mobile_data_request_spec.rb
```diff
@@ -96,4 +96,57 @@
...
   end
+
+  describe '#save_and_notify_account_holder!' do
+    before do
+      ActiveJob::Base.queue_adapter = :test
+    end
+
+    after do
+      ActiveJob::Base.queue_adapter = :inline
+    end
```
aldavidson, line 107:
```suggestion
    around do
      # prevent api call to Notify
      original_queue_adapter = ActiveJob::Base.queue_adapter
      ActiveJob::Base.queue_adapter = :test
      example.run
      ActiveJob::Base.queue_adapter = original_queue_adapter
    end
```

spec/features/responsible_body/submitting_a_spreadsheet_request_spec.rb
benilovj: Some of this should be addressed by https://github.com/DFE-Digital/get-help-with-tech/pull/214

benilovj:
I've merged #214 so I'll rebase this PR...

spec/models/extra_mobile_data_request_spec.rb
benilovj: Updated in 9e9eea4

spec/features/responsible_body/submitting_a_spreadsheet_request_spec.rb
benilovj: Updated in 9e9eea4

app/controllers/responsible_body/mobile/manual_requests_controller.rb
tonyheadford: Nice idea

PR #216, opened 2020-08-10
374 add slack notifications

aldavidson:
### Context

See [Trello card 374](https://trello.com/c/x7D0iisc/374-add-slack-integration-to-service) - to get more visibility into usage of the service, we'll set up a Slack notification bot, that posts notices to a dedicated channel when various significant events happen

Add the framework for sending Slack messages

- a `SlackMessage` class with ActionMailer-style `send_now` / `send_later` semantics
- an ActiveJob class to enable the send_later path
- config settings to govern the above

Note that this PR _does not send any actual messages_ - just provides the framework for doing so. I'll hook in the actual sending of notifications as a separate PR, as that one may require several iterations before we're happy with it.

### Guidance to review

You will need the following environment variables setup:
```
GHWT__SLACK__EVENT_NOTIFICATIONS__WEBHOOK_URL
```
That is a secret value, so you'll need to either have access to the [Slack  app's settings](https://api.slack.com/apps/A018NEABVNY/), or ask @aldavidson for a usable value

From the console:

```
> msg = SlackMessage.new(text: 'Hello from my console :pikachu-wave:')
> msg.send_now!
or
> msg.send_later!
```

The time between sending messages and them appearing in the Slack channel is usually very fast (<1s) - but this will presumably vary with load.


aldavidson:
> LGTM, other than I don't see reference to the `SlackMessage.hyperlink` method so just wondered if it was needed?

I copied that from the Apply implementation - I figured it will come in very handy when we actually start sending messages

tonyheadford:
> > LGTM, other than I don't see reference to the `SlackMessage.hyperlink` method so just wondered if it was needed?
> 
> I copied that from the Apply implementation - I figured it will come in very handy when we actually start sending messages

Just wondering what it's for - is it for formatting links to embed in Slack messages?

PR #222, opened 2020-08-11
Hide nav menu on you're signed in and check your email

tonyheadford:
### Context
[Trello card](https://trello.com/c/ylCz6nEW/360-dont-show-nav-links-on-the-interstitial-youre-signed-in-and-check-your-email-pages) - hide nav menus from users on the 'You're signed in' and 'Check your email' pages during sign-in process

Make rendering of nav (and mobile menu link) conditional based on `controller_name` and `action_name` to prevent rendering where not required.
 
### Guidance to review
Nav menu should not be visible on the 'You're signed in' and 'Check your email' pages but visible as expected elsewhere.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -16,6 +16,10 @@ module ApplicationHelper
     ]
   end
 
+  def hide_nav_menu?
+    controller_name == 'sign_in_tokens' && (%w[validate sent].include? action_name)
+  end
+
   def nav_item(title: nil, url: nil, content: nil, html_options: {})
     class_name = 'govuk-header__navigation-item'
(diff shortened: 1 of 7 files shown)


app/helpers/application_helper.rb
```diff
@@ -16,6 +16,10 @@ def yes_no_options
     ]
   end
 
+  def hide_nav_menu?
+    controller_name == 'sign_in_tokens' && (%w[validate sent].include? action_name)
+  end
+
```
benilovj, line 22: It feels icky for the superclass knowing about the subclass. An alternative could be defining

```
def hide_nav_menu?
  false
end
```

here

and overriding it withing the `SignInTokensController`?

tonyheadford: OK, moved the helper into the controller

PR #224, opened 2020-08-12
add school device allocations model, migration and factory

aldavidson:
### Context

We need to add a model for allocations of devices to schools, so that we can build a provisional API endpoint for computacenter to hit when an order is placed

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_08_12_140013) do
+ActiveRecord::Schema.define(version: 2020_08_12_155338) do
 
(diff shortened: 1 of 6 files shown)


db/schema.rb
```diff
@@ -70,6 +70,18 @@
     t.datetime "updated_at", precision: 6, null: false
   end
 
+  create_table "school_device_allocations", force: :cascade do |t|
+    t.bigint "schools_id"
```
benilovj, line 74: am wondering should this be `school_id`?

aldavidson: ooh, well-spotted, fixed

aldavidson:
> LGTM although I think we should get some kind of auditing gem in the service sooner rather than later, because the `created_by_*` and `last_updated_by_*` columns might start getting out of hand soon (or we may forget to update them in places).

Yep, agreed

PR #226, opened 2020-08-12
Add responsible body users through the support console

benilovj:
### Context

The responsible body contacts we have on the system aren't always correct, and we have had a number of support requests asking for new users to be added. Up to now, only developers have been able to do this by ssh-ing into production, which led to delays in resolving the support requests.

- display the list of existing responsible body users in the support console
- allowing DfE staff to add new users to on-boarded responsible bodies

### Guidance to review

image

image

image

image


app/views/support/users/new.html.erb
```diff
@@ -0,0 +1,35 @@
...
+          <%= f.govuk_text_field :full_name, label: {size: 'm', text: 'Full name'} %>
+
+          <%= f.govuk_text_field :email_address, label: {size: 'm', text: 'Email address'} %>
+
+          <%= f.govuk_submit 'Submit' %>
+      <%- end %>
+    </div>
+
+  </div>
+</div>
```
fofr, line 35:
```suggestion
<div class="govuk-grid-row">
  <div class="govuk-grid-column-two-thirds">
      <%= form_for @user, url: support_responsible_body_users_path, method: :post do |f| %>
          <%= f.govuk_error_summary %>
          
          <h1 class="govuk-heading-xl">Add a user</h1>

          <%= f.govuk_text_field :full_name, label: {size: 'm', text: 'Full name'} %>

          <%= f.govuk_text_field :email_address, label: {size: 'm', text: 'Email address'} %>

          <%= f.govuk_submit 'Submit' %>
      <%- end %>
    </div>
</div>
```

spec/features/support/adding_new_responsible_body_users_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+
+    when_i_sign_in_as_a_dfe_user
+    and_i_visit_a_support_responsible_body_page
+    then_i_can_see_the_responsible_body_with_users
+
+    when_try_adding_a_new_user_with_invalid_details
+    then_i_can_see_error_messages
+
+    and_when_i_retry_adding_a_new_user_with_valid_details
+    then_i_can_see_the_responsible_body_with_that_new_users_as_well
```
aldavidson, line 18: typo in `_that_new_users_` ?

benilovj: fixed, thanks

PR #225, opened 2020-08-12
408 add computacenter namespace

aldavidson:
### Context

First part of [Trello card 408](https://trello.com/c/fsNGK22g/408-spin-up-first-api-end-point-to-share) - spin up a first draft API endpoint for Computacenter to notify us of orders placed

### Changes proposed in this pull request

* Add `/computacenter` namespaced `base_` and `home_` controllers
* Add distinguishing of computacenter users based on their email address
* Add basic placeholder computacenter homepage
* Restrict computacenter-namespaced controllers to computacenter users

PR #227, opened 2020-08-13
408 add API tokens

aldavidson:
### Context

Second part of [Trello card 408](https://trello.com/c/fsNGK22g/408-spin-up-first-api-end-point-to-share)

This PR builds on #225 and adds a basic UI for computacenter users to manage their API tokens

Screen Shot 2020-08-13 at 11 28 05

### Guidance to review

You'll need to create a user with a computacenter.com email address, and log in as them
Then click 'API tokens' in the nav


app/models/api_token.rb
```diff
@@ -0,0 +1,32 @@
+class APIToken < ApplicationRecord
+  belongs_to :user
+
+  validates :user, presence: true
+  validates :token, presence: { message: I18n.t('activerecord.errors.models.api_token.token.blank') },
+                    uniqueness: { scope: :user_id },
+                    length: { minimum: 36, maximum: 36 }
+  validates :name,  presence: { message: I18n.t('activerecord.errors.models.api_token.name.blank') },
```
benilovj, line 8:
```suggestion
  validates :name,  presence: true,
```
You shouldn't need to hard-code the path to the message – if the message is in the right place, it will be picked up automatically.

app/models/api_token.rb
```diff
@@ -0,0 +1,32 @@
...
+                    length: { minimum: 2, maximum: 64 },
+                    uniqueness: { scope: :user_id, message: I18n.t('activerecord.errors.models.api_token.name.uniqueness') }
```
benilovj, line 10: As above, message shouldn't be needed if the message is in the right place in `en.yml`

app/models/api_token.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def fill_in_defaults!
+    generate_token! if token.blank?
+    self.status ||= APIToken.statuses[:active]
+  end
+
+  def self.generate!(user)
+    token = create(user: user)
+    token
```
benilovj, line 30:
```suggestion
    create(user: user)
```

app/models/api_token.rb
```diff
@@ -0,0 +1,32 @@
+class APIToken < ApplicationRecord
+  belongs_to :user
+
+  validates :user, presence: true
+  validates :token, presence: { message: I18n.t('activerecord.errors.models.api_token.token.blank') },
```
benilovj, line 5:
```suggestion
  validates :token, presence: true},
```
Message shouldn't be needed if the message is in the right place in `en.yml`

app/views/computacenter/api_tokens/index.html.erb
```diff
@@ -0,0 +1,52 @@
...
+    (You don't have any API tokens at the moment)
+  </p>
+<%- else %>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+      <table class="govuk-table">
+        <thead class="govuk-table__head">
+          <tr class="govuk-table__row">
+            <th class="govuk-table__header">
+              <%= sortable_table_header( 'Name', :name ) %>
```
benilovj, line 23: I'd take the sortable headers out – we won't ever have that many tokens to need that, and the controller doesn't support it?

aldavidson: ah, yes, copy-and-paste UI artefact - done

PR #230, opened 2020-08-13
408 allow access via API token

aldavidson:
### Context

Third PR for [Trello Card 408 - standup a first draft API endpoint for Computacenter](https://trello.com/c/fsNGK22g/408-spin-up-first-api-end-point-to-share)

* Add an APIAuthenticationService to identify users by given API token
* Add a `computacenter/api` namespace, to keep the not-truly-RESTful endpoint separate from the rest (heh) of the application
* Add an empty placeholder `computacenter/api/usage-caps/bulk-update` method

### Guidance to review

You will need an active API token for a Computacenter user 

_to create one, first create a user with a computacenter.com email address, then log in as them. Click on 'API tokens' in the nav, generate a new token. Copy the token value_

From a shell:

```bash
# Replace YOUR_TOKEN_VALUE_HERE with your token value
$ curl -v -H "Authorization: Bearer YOUR_TOKEN_VALUE_HERE" -H "Content-Type: application/xml" -X POST  http://localhost:3000/computacenter/api/cap-usage/bulk-update
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 3000 (#0)
> POST /computacenter/api/cap-usage/bulk-update HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/7.54.0
> Accept: */*
> Authorization: Bearer YOUR_TOKEN_VALUE_HERE
> Content-Type: application/xml
> 
< HTTP/1.1 204 No Content
< X-Frame-Options: SAMEORIGIN
< X-XSS-Protection: 1; mode=block
< X-Content-Type-Options: nosniff
< X-Download-Options: noopen
< X-Permitted-Cross-Domain-Policies: none
< Referrer-Policy: strict-origin-when-cross-origin
< Cache-Control: no-cache
< X-Request-Id: 3c8149f5-f368-4ce8-85a6-ac53a03829a4
< X-Runtime: 0.179905
< 
* Connection #0 to host localhost left intact

```

Trying it with an incorrect token:

```bash
$ curl -v -H "Authorization: Bearer __INVALID_TOKEN__" -H "Content-Type: application/xml" -X POST  http://localhost:3000/computacenter/api/cap-usage/bulk-update
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 3000 (#0)
> POST /computacenter/api/cap-usage/bulk-update HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/7.54.0
> Accept: */*
> Authorization: Bearer __INVALID_TOKEN__
> Content-Type: application/xml
> 
< HTTP/1.1 403 Forbidden
< X-Frame-Options: SAMEORIGIN
< X-XSS-Protection: 1; mode=block
< X-Content-Type-Options: nosniff
< X-Download-Options: noopen
< X-Permitted-Cross-Domain-Policies: none
< Referrer-Policy: strict-origin-when-cross-origin
< Content-Type: application/xml; charset=utf-8
< Cache-Control: no-cache
< X-Request-Id: 04f2c2b6-eb65-4611-8a29-3571740359bd
< X-Runtime: 0.020717
< Transfer-Encoding: chunked
< 
<?xml version="1.0" encoding="UTF-8"?>
<error>
  <status>forbidden</status>
  <message>You are not authorized to perform this action</message>
</error>
* Connection #0 to host localhost left intact
```

Trying it without providing an Authorization header:

```bash
$ curl -v  -H "Content-Type: application/xml" -X POST  http://localhost:3000/computacenter/api/cap-usage/bulk-update
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 3000 (#0)
> POST /computacenter/api/cap-usage/bulk-update HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/7.54.0
> Accept: */*
> Content-Type: application/xml
> 
< HTTP/1.1 401 Unauthorized
< X-Frame-Options: SAMEORIGIN
< X-XSS-Protection: 1; mode=block
< X-Content-Type-Options: nosniff
< X-Download-Options: noopen
< X-Permitted-Cross-Domain-Policies: none
< Referrer-Policy: strict-origin-when-cross-origin
< Content-Type: application/xml; charset=utf-8
< Cache-Control: no-cache
< X-Request-Id: f6b83578-8c43-4a2b-ade0-a81daba9d490
< X-Runtime: 0.010958
< Transfer-Encoding: chunked
< 
<?xml version="1.0" encoding="UTF-8"?>
<error>
  <status>unauthorized</status>
  <message>You must provide an Authorization header with a valid Bearer token</message>
</error>

```

### DIFF
--- a/app/controllers/computacenter/api/base_controller.rb
+++ b/app/controllers/computacenter/api/base_controller.rb
@@ -1,5 +1,27 @@
-class Computacenter::API::BaseController < Computacenter::BaseController
+# We're designing our API endpoint to fit the client's specification.
+# We'd normally do it the other way round, but due to the urgency of the
+# need and the fact that it's probably faster for to iterate our system
+# than theirs, this is the most pragmatic solution.
+# The client's specification here is XML-based, and not strictly RESTful -
+# all XML packets will be sent as a POST, regardless of the action.
+# So we're pulling this out to a dedicated controller in order to keep the
+# rest of the application 'clean' and RESTful.
+class Computacenter::API::BaseController < ApplicationController
+  before_action :require_cc_user!
+  rescue_from APIError, with: :api_error!
 
 private
 
+  # overriden method tailored for XML responses
+  def require_cc_user!
+    if @user
+      raise APIError.new(status: :forbidden, message: 'You are not authorized to perform this action') unless @user.is_computacenter?
+    else
+      raise APIError.new(status: :unauthorized, message: 'You must provide an Authorization header with a valid Bearer token')
+    end
+  end
+
+  def api_error!(error)
+    render xml: error, status: error.status and return
+  end
 end
--- a/app/controllers/computacenter/api/base_controller.rb
+++ b/app/controllers/computacenter/api/base_controller.rb
@@ -1,6 +1,6 @@
 # We're designing our API endpoint to fit the client's specification.
 # We'd normally do it the other way round, but due to the urgency of the
-# need and the fact that it's probably faster for to iterate our system
+# need and the fact that it's probably faster for to iterate our system
 # than theirs, this is the most pragmatic solution.
 # The client's specification here is XML-based, and not strictly RESTful -
 # all XML packets will be sent as a POST, regardless of the action.
--- a/app/controllers/computacenter/api/base_controller.rb
+++ b/app/controllers/computacenter/api/base_controller.rb
@@ -14,8 +14,8 @@ private
 
   # overriden method tailored for XML responses
   def require_cc_user!
-    if @user
-      raise APIError.new(status: :forbidden, message: 'You are not authorized to perform this action') unless @user.is_computacenter?
+    if APIAuthenticationService.authorization_given?(request)
+      raise APIError.new(status: :forbidden, message: 'You are not authorized to perform this action') unless @user&.is_computacenter?
     else
       raise APIError.new(status: :unauthorized, message: 'You must provide an Authorization header with a valid Bearer token')
     end
--- a/app/controllers/computacenter/api/base_controller.rb
+++ b/app/controllers/computacenter/api/base_controller.rb
@@ -12,9 +12,17 @@ class Computacenter::API::BaseController < ApplicationController
 
 private
 
+  def identify_user!
+    @user = APIAuthenticationService.identify_user(bearer_token)
+  end
+
+  def bearer_token
+    request.headers['Authorization'].to_s.gsub(/Bearer\s+([^\s]*)$/, '\1')
+  end
+
   # overriden method tailored for XML responses
   def require_cc_user!
-    if APIAuthenticationService.authorization_given?(request)
+    if bearer_token.present?
       raise APIError.new(status: :forbidden, message: 'You are not authorized to perform this action') unless @user&.is_computacenter?
     else
       raise APIError.new(status: :unauthorized, message: 'You must provide an Authorization header with a valid Bearer token')
--- a/app/controllers/computacenter/api/base_controller.rb
+++ b/app/controllers/computacenter/api/base_controller.rb
@@ -13,7 +13,7 @@ class Computacenter::API::BaseController < ApplicationController
 private
 
   def identify_user!
-    @user = APIAuthenticationService.identify_user(bearer_token)
+    @user = APIToken.active.find_by(token: bearer_token)&.user
   end
 
   def bearer_token
--- a/app/controllers/computacenter/api/base_controller.rb
+++ b/app/controllers/computacenter/api/base_controller.rb
@@ -1,0 +1,5 @@
+class Computacenter::API::BaseController < Computacenter::BaseController
+
+private
+
+end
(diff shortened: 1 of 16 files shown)


app/services/api_authentication_service.rb
```diff
@@ -0,0 +1,16 @@
...
+    end
+  end
+
+  def self.authorization_given?(request)
+    request.headers['Authorization'].present?
+  end
+
+  def self.given_token(request)
+    request.headers['Authorization'].to_s.gsub(/Bearer\s+([^\s]*)$/, '\1')
+  end
```
benilovj: It feels like this class is mixing two concerns:

1. Extracting the bearer token from the request.
2. Identifying the user whose token it is

IMHO, the extraction of the bearer token should be a method on `Computacenter::API::BaseController` because it's related to translating from HTTP requests to higher-level data, which is a controller responsibility:

```
def bearer_token
  request.headers['Authorization']&.to_s&.gsub(/Bearer\s+([^\s]*)$/, '\1')
end
```

Then this service simply becomes:

```
class APIAuthenticationService
  def self.identify_user(token)
    if matched_token = APIToken.active.where(token: token).first
      matched_token&.user
    end
  end
end
```
and you call `APIAuthenticationService.identify_user(bearer_token)` in the controller.

WDYT?

benilovj: Actually it could even be

```
class APIAuthenticationService
  def self.identify_user(token)
    APIToken.active.find_by(token: token)&.user
  end
end
```

aldavidson: It's a fair point, and a good idea... except that we need to call this from `ApplicationController`. So we'd then need to override the `populate_user_from_session_or_api_token!` method in the Computacenter::BaseController, which also feels a bit messy in a different way.

I'll try a few more things

benilovj: > It's a fair point, and a good idea... except that we need to call this from ApplicationController

`bearer_token` could sit in `ApplicationController` so that problem goes away...

aldavidson: Yes! If I also -
* rename the `populate_user_from_session_or_api_token!` method to `identify_user!`
* put it back to just considering the user session
* override it in the computacenter/api/base_controller
- then that works, and looks a lot cleaner too. Thanks!

_(Naming Things Right Is Hard But Often Crucial part 174825)_

app/controllers/computacenter/api/base_controller.rb
```diff
@@ -12,9 +12,17 @@ class Computacenter::API::BaseController < ApplicationController
 
 private
 
+  def identify_user!
+    @user = APIAuthenticationService.identify_user(bearer_token)
```
benilovj, line 16: you can safely inline this now and get rid of the service

```suggestion
    @user = APIToken.active.find_by(token: bearer_token)&.user
```

app/controllers/computacenter/api/base_controller.rb
```diff
@@ -12,9 +12,17 @@ class Computacenter::API::BaseController < ApplicationController
...
+  end
+
+  def bearer_token
+    request.headers['Authorization'].to_s.gsub(/Bearer\s+([^\s]*)$/, '\1')
```
benilovj, line 20: As is, this returns `""` rather than nil when the `Authorization` header isn't set, which doesn't feel completely right. How about:

```suggestion
    request.headers['Authorization']&.gsub(/Bearer\s+([^\s]*)$/, '\1')
```

app/controllers/computacenter/api/base_controller.rb
aldavidson: Yep, agreed. Related, I've put a Trello card in to review the service objects we have - I feel like the concept has got muddled and they're becoming a dumping ground for stuff that in a previous life I might have put in the dreaded 'utils' library....

aldavidson:
> Don't want to stop this going in but just to clarify in my mind - we're not planning to authenticate a user by a token - only a system - so no front end for the token user? Would it make sense to separate the Computacenter concept from the API i.e. API accessed by a token (nothing to do with a computacenter naming), Computacenter users have a namespace and access to the UI and can generate tokens? Seems a better separation to me?

I've thought about this quite a lot over the last day or so - I would ordinarily agree, but in this case there's a few factors that lean me towards keeping this explicitly computacenter-specific:

* this API is not RESTful, nor general-purpose 
* the methods, content-type and XML packets supported are not how we would have designed them - it's entirely to Computacenter's requirements 
* we may well expose other methods as a more general RESTful API later, but we'd still need to support this method for the Computacenter integration

I think there is something in the idea of making this API independent of the concept of a user, but the per-user concept is so fundamentally interwoven into the application that the right way to do that would be to have a completely separate application, probably done via `rails-api` or something similar. That would mean non-trivial changes in the build / test / deploy pipeline and the PaaS setup, and given that we have 10 days to launch I think at this point we need to be really pragmatic about the tradeoffs.

PR #235, opened 2020-08-16
408 add cc api docs

aldavidson:
### Context

See [Trello card 408](https://trello.com/c/fsNGK22g/408-spin-up-first-api-end-point-to-share). We need to provide some documentation for the CC API, that we can point the Computacenter developers at. 
 
Add some API docs. Rather than send all this information in an email, it seemed sensible to write it up as a markdown file on our repo.

### Guidance to review

Read [the docs](https://github.com/DFE-Digital/get-help-with-tech/blob/6fa0898aff84c0688a941a7e8eb6f1ed4312b684/docs/computacenter_api.md)!

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,10 @@ All the above are run automatically on GitHub Actions when pushing a PR.
 [GOV.UK Notify](https://www.notifications.service.gov.uk/) for sending emails
 [Slack](https://api.slack.com/apps/A018NEABVNY/) for notifications of various significant events
 
+### APIs
+
+[Computacenter API](/docs/computacenter_api.md)
+
 ## Deploying on GOV.UK PaaS
 
 ### Prerequisites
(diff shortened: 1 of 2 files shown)


aldavidson:
> Looks good for MVP.
> 
> One thing that the docs don't make clear is what could cause a `CapUsage` bulk update to fail. I can think of at least one scenario, which is the school isn't found...

I've added a paragraph explaining that - thanks! 👍

benilovj:
> > Looks good for MVP.
> 
> > 
> 
> > One thing that the docs don't make clear is what could cause a `CapUsage` bulk update to fail. I can think of at least one scenario, which is the school isn't found...
> 
> 
> 
> I've added a paragraph explaining that - thanks! 👍 

Did you definitely push it up? I can't see it...

PR #237, opened 2020-08-17
380 import address data for schools into the service

tonyheadford:
### Context
[Trello card](https://trello.com/c/xl8hAFXt/380-import-address-data-for-schools-into-the-service) We need the school address so that devices can be delivered to the school. This change adds the address information to the `School` model  and populates it during the import from the GIAS data.

Migration to add address fields and change to import routines to include the address information in the GIAS data file.

### Guidance to review
Schools records should now have an address


db/migrate/20200814143849_add_address_to_schools.rb
```diff
@@ -0,0 +1,10 @@
+class AddAddressToSchools < ActiveRecord::Migration[6.0]
+  def change
+    add_column :schools, :address_1, :string
+    add_column :schools, :address_2, :string
+    add_column :schools, :address_3, :string
```
aldavidson, line 5: should any of these have `null: true` ?

tonyheadford: They get that by default? I would only normally add `null: false` if I wanted the 'NOT NULL' clause?

aldavidson: no, I guess the default is fine 👍

PR #240, opened 2020-08-18
fix "No such file or directory" error on first setup

aldavidson:
### Context

There was no `log/` directory in the repo, which meant running `rake db:setup` on a fresh install failed with an error `Errno::ENOENT: No such file or directory @ rb_sysopen - log/mail.log`.

Make the .gitignore directive more specific and add an empty `log/placeholder.txt` file, to make the repo contain a `log/` dir

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@
 /.bundle
 
 # Ignore all logfiles and tempfiles.
-/log/*
+/log/*.log
 /tmp/*
 /coverage
 
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@
 /.bundle
 
 # Ignore all logfiles and tempfiles.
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -8,7 +8,7 @@
 /.bundle
 
 # Ignore all logfiles and tempfiles.
-/log/*
+/log/*.log
```
aliuk2012, line 11: Is this needed because this rule already exists https://github.com/DFE-Digital/get-help-with-tech/blob/main/.gitignore#L38

aliuk2012:
Could we rename the `log/placeholder.txt` to the more standard `.keep` or `.gitkeep`?

.gitignore
aldavidson: good spot 👍

PR #250, opened 2020-08-19
Fix table of content link styles

aliuk2012:
### Context
We are using RedCarpet to render our markdown files to html. RedCarpet is not adding the relevant govuk classes to links.

Setting `$govuk-global-styles` to true, means that paragraphs (`<p>`) and links (`<a>`) can be styled to look like govuk without explicitly having to apply the govuk-body and govuk-link classes.

#### Before
image

#### After 
image


benilovj:
> We are using RedCarpet to render our markdown files to html. RedCarpet is not adding the relevant govuk classes to links.

Is the issue with the table of contents links or with other links? They are generated in different ways.

benilovj:
For context, the Markdown body is rendered using https://github.com/DFE-Digital/get-help-with-tech/blob/main/lib/govuk/markdown_renderer.rb which sets the right classes.

The table of contents could potentially be done using a custom renderer as well but that would basically involve re-writing the Redcarpet out-of-the-box functionality which felt like overkill. The TOC is wrapped in a div, `.app-table-of-contents`, which is used in other places to fix styling.

fofr:
I think we should apply styles only to the TOC container. Not sure what effect enabling global styles will have now.

PR #253, opened 2020-08-19
Fix issue with generating Computacenter cap types in outgoing API

aldavidson:
### Context

`Computacenter::CapTypeConverter` was only returning first character of matched device types. This was causing `500 internal server error` responses on their API

Fix it to return the full match
Add a spec to explicitly test this behaviour


benilovj (approved):

tonyheadford (approved):
LGTM

PR #255, opened 2020-08-19
add verbose explicit logging of computacenter api requests

aldavidson:
### Context

Computacenter are experiencing an issue on staging that we can't replicate based on the information available to us. 
When they POST a cap usage update to the computacenter API, they get a `500 Internal Server Error` - based on the [info in Sentry](https://sentry.io/organizations/dfe-get-help-with-tech/issues/1846682979/?project=5320558&query=is%3Aunresolved#exception) it looks like the server is trying to interpret it as a HTML request and failing when looking for a HTML template to render. This happens even though Sentry says they're supplying a correct `"Content-Type: application/xml"` header.

Add some explicit verbose logging so that we can see in the logs exactly what they're supplying in terms of headers & body.


tonyheadford (approved):
LGTM!

PR #256, opened 2020-08-19
Remove local authorities that don't maintain schools

benilovj:
### Context

We have a large number of local authorities that don't actually maintain schools, as evidenced by no schools being associated with them within Get Information About Schools. These LAs are also not involved in the connectivity pilots or have any users:

```ruby
[1] pry(main)> LocalAuthority.where(organisation_type: %w[combined_authority non_metropolitan_district strategic_regional_authority], in_connectivity_pilot: true).count
=> 0
[2] pry(main)> LocalAuthority.where(organisation_type: %w[combined_authority non_metropolitan_district strategic_regional_authority]).select {|la| la.users.count > 0 }
=> []
```

- remove those LAs from the database
- don't import them from the LAs in England register in the first place


tonyheadford (approved):
LGTM

PR #259, opened 2020-08-20
Fix cap update endpoint when only one record present

benilovj:
### Context

The Cap update is failing when there's only one cap update being sent. The investigation showed that this stems from differences in how the XML to Ruby conversion works, depending on whether it's one or multiple records in the payload.

- don't use stubs in the CapUsageController specs because they were obscuring the failure
- fix the bug and and add spec coverage for that case


tonyheadford (approved):
Very nice!

PR #272, opened 2020-08-20
New home page for "Get help with Tech"

aliuk2012:
### Context
First iteration of the Home page. 

The home page will be used to link up all the guidance from the various parts of the programme.

#### Out of scope
**Header** - We need to clearly define from "Guidance" link is for and be consistent with the links in the header across the programme.

**Footer** - In this PR we updated the support content in the footer. This will need to be updated across the service. Two policies have been removed from the home page footer as the home page does not request personal information. There is another piece of work going on to try bring the difference privacy policies together so that there is a single privacy policy across the programme. The same goes for accessibility link but this will be done at a later stage.
 
image

### Guidance to review

Visit https://ghwt-review-pr-272.herokuapp.com/

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,7 +1,10 @@
 require Rails.root.join('lib/constraints/require_dfe_user_constraint')
 
 Rails.application.routes.draw do
+  root 'pages#home_page', as: :guidance_page
+
   get '/start', to: 'pages#start'
+
   get '/about-bt-wifi', to: 'pages#about_bt_wifi'
   get '/about-increasing-mobile-data', to: 'pages#about_increasing_mobile_data'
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -1,7 +1,11 @@
 require Rails.root.join('lib/constraints/require_dfe_user_constraint')
 
 Rails.application.routes.draw do
+  root 'pages#home_page', as: :guidance_page
+
   get '/start', to: 'pages#start'
+  # TODO: Temporary redirect can be removed once new page is deployed
+  get '/new-home-page', to: redirect('/')
```
aliuk2012, line 8: Will be removed before merging this PR

aldavidson (approved):
LGTM

PR #291, opened 2020-08-22
Filter on schools table

benilovj:
### Context

For responsible bodies with lots of schools, they may want to filter down the list to help them process it.

- add a form that allows filtering by any text in the school table row
- bugfix: make the content in the 'missing schools' callout conditional on the type of responsible body
- split the view template into partials for readability

### Guidance to review

Screenshot:

image

Code-wise, I'm not sure I've brought the JS across correctly from the prototype.


fofr:
> These need some JS unit tests.
This JS is ported from a GOV.UK repo, we can port the tests too.

PR #293, opened 2020-08-23
Stop the UI crashing when a school does not have a PreorderInformation record

aldavidson:
### Context

Until we have done [Trello card 476-ensure-data-consistency-when-a-new-school-is-added](https://trello.com/c/31RXvNJF/476-ensure-data-consistency-when-a-new-school-is-added) it's possible (likely, in fact) that a school could have no associated `PreorderInformation` record. The UI will currently raise an error in the schools list and the school details page if this is the case.

Default the displayed `who_will_order_devices` value to the value from the school's RepsonsibleBody, if it has no `PreorderInformation` record

Make the `:la_maintained` and `:academy` traits on the `school` factory set the `responsible_body` association to a LocalAuthority or a Trust respectively

### DIFF
--- a/app/components/school_details_summary_list_component.rb
+++ b/app/components/school_details_summary_list_component.rb
@@ -21,7 +21,7 @@ class SchoolDetailsSummaryListComponent < ViewComponent::Base
       },
       {
         key: 'Who will order?',
-        value: "The #{@school.preorder_information.who_will_order_devices_label.downcase} orders devices",
+        value: "The #{(@school.preorder_information || @school.responsible_body).who_will_order_devices_label.downcase} orders devices",
       },
     ]
   end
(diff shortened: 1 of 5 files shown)


tonyheadford (approved):
Not wanting to block this, but it feels like this would be better as a method on a School rather than the logic being exposed in the view and component?

benilovj (changes requested):
+1 to what @tonyheadford said...

Al and I already discussed offline that we should try to prevent inconsistent data, or rectify it when it occurs, rather than trying to pave over it in the view layer...

aldavidson:
> +1 to what @tonyheadford said...
> 
> Al and I already discussed offline that we should try to prevent inconsistent data, or rectify it when it occurs, rather than trying to pave over it in the view layer...

I will close this PR then

PR #297, opened 2020-08-23
Set initial status on PreorderInformation and disallow it to be null

benilovj:
### Context

There's no good reason for `PreorderInformation#status` to ever be null. It would introduce lots of conditional complexity if we allowed this to creep into the data.

- infer the status when `PreorderInformation` is initialised
- validate `status` in the AR model
- disallow `status` to be null in the DB

### Guidance to review

We don't need a data migration because these records haven't been created anywhere yet.


aldavidson (approved):
LGTM

PR #302, opened 2020-08-24
Update deprecated option in webpacker config

benilovj:
Without this change, running specs printed out loads of warnings:

The resolved_paths option has been deprecated. Use additional_paths instead.


aldavidson (approved):
👍  yeah, that's been irritating me too, thanks :)

PR #311, opened 2020-08-24
Loosen cap usage xsd restrictions

benilovj:
### Context

When DfE pushes a zero cap to Computacenter, it's valid for a zero cap and a zero cap usage to be pushed back. This XML capUsage payload was being rejected because the XSD was too restrictive.

- loosen the XSD restrictions
- add a spec to cover this case

### Guidance to review

I've converted the linefeeds on CapUsage.XSD from Windows to Unix, so it's better to review this while ignoring whitespace.


aldavidson (approved):
This looks reasonable to me :+1

PR #312, opened 2020-08-24
Fix display of school contacts

benilovj:
### Context

The `<br>` was being eaten and so the contact details ran into each other as one long string.

Escape the user-generated content but mark the `<br>` tags as safe.

### Guidance to review

Before:

image

After:

image


benilovj:
@tonyheadford I think I couldn't reproduce this before due to running stale configuration on the local dev server.

tonyheadford (approved):
LGTM

PR #314, opened 2020-08-25
replace is_dfe?/is_computacenter? methods with explicit user flags

aldavidson:
### Context

We've been inferring whether users could access the support & computacenter interfaces based on their email domain.
This has worked OK, but it makes testing difficult - and especially for the external pentesting company, who haven't been able to test the support or computacenter interfaces yet as their email domain is their company.

Replace the `is_dfe?` and `is_computacenter?` methods on `User` with explicit boolean flags
*NOTE* I also changed `is_dfe?` to `is_support?` - as it's more reflective of reality, and better matches the application file structure.


tonyheadford (approved):
Look good!

PR #317, opened 2020-08-25
[527, 528] fix bugs with preorder information around schools needing chromebooks

aldavidson:
### Context

[Trello card 528](https://trello.com/c/UwDJhUpJ/528-bug-school-preorder-info-status-should-update-to-ready-when-they-dont-need-chromebooks) In the RB interface, clicking 'No, they will not need chromebooks' + 'save' on a school should update its status to READY.
It was not doing that, because of Rails' auto-created methods & type inference meaning that `will_need_chromebooks?` was returning `true` when the field had a value of 'no' (because it's not `nil`, therefore '(attr_name)?' is true)

[Trello card 527](https://trello.com/c/IuWBUbSX/527-dont-show-chromebook-details-if-chromebooks-wont-be-ordered) - it was still showing the domain & recovery email fields when `will_need_chromebooks` was `no` 

Fix both bugs, and add explicit tests for this behaviour

### DIFF
--- a/app/controllers/responsible_body/devices/chromebook_information_controller.rb
+++ b/app/controllers/responsible_body/devices/chromebook_information_controller.rb
@@ -16,7 +16,8 @@ class ResponsibleBody::Devices::ChromebookInformationController < ResponsibleBod
       { school: @school }.merge(chromebook_params),
     )
     if @chromebook_information_form.valid?
-      @preorder_info.update!(chromebook_params.merge(status: @preorder_info.infer_status))
+      @preorder_info.update!(chromebook_params)
+      @preorder_info.update!(status: @preorder_info.infer_status)
       redirect_to responsible_body_devices_school_path(urn: @school.urn)
     else
       render :edit, status: :unprocessable_entity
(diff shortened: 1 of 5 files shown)


app/controllers/responsible_body/devices/chromebook_information_controller.rb
```diff
@@ -16,7 +16,8 @@ def update
       { school: @school }.merge(chromebook_params),
     )
     if @chromebook_information_form.valid?
-      @preorder_info.update!(chromebook_params.merge(status: @preorder_info.infer_status))
+      @preorder_info.update!(chromebook_params)
```
tonyheadford, line 19: Could this be pushed into a method on the PreorderInformation class e.g. `@preorder_info.update_chromebook_information!(chromebook_params)` which then handles the changes and updates the status?

tonyheadford (approved):
Looks good other than the comment about handling updates to the preorder information

app/controllers/responsible_body/devices/chromebook_information_controller.rb
aldavidson: Good call - done

PR #321, opened 2020-08-25
452 manage rb users invite an rb user

tonyheadford:
### Context
[Trello card](https://trello.com/c/vZGLZ3do/452-manage-rb-users-invite-an-rb-user) Take adding RB users out from behind the FeatureFlag curtain and add the invite email to Notify.

Enabling the invite responsible body user functionality. Sending an invite email to the new user.

### Guidance to review
Invite button enabled and Invite a new user form/functionality accessible without enabling the `:rbs_can_manage_users` feature flag.
Screenshot 2020-08-25 at 16 03 19
Screenshot 2020-08-25 at 16 04 07

### DIFF
--- a/app/mailers/invite_responsible_body_user_mailer.rb
+++ b/app/mailers/invite_responsible_body_user_mailer.rb
@@ -1,0 +1,22 @@
+class InviteResponsibleBodyUserMailer < ApplicationMailer
+  def invite_user_email
+    @user = params[:user]
+    rb_name = @user.responsible_body.local_authority_official_name || @user.responsible_body.name
+    personalisation = {
+      email_address: @user.email_address,
+      responsible_body_name: rb_name,
+    }
+
+    template_mail(
+      notify_template_id,
+      to: @user.email_address,
+      personalisation: personalisation,
+    )
+  end
+
+private
+
+  def notify_template_id
+    Settings.govuk_notify.templates.devices.invite_responsible_body_user
+  end
+end
--- a/app/views/responsible_body/users/index.html.erb
+++ b/app/views/responsible_body/users/index.html.erb
@@ -34,7 +34,7 @@
       <%- end %>
     </ul>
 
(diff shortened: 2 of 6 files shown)


app/views/responsible_body/users/index.html.erb
```diff
@@ -24,18 +24,17 @@
...
+      <li>view the list of existing users</li>
       <li>order devices for centrally managed schools</li>
       <%- if @responsible_body.in_connectivity_pilot? %>
         <li>download BT vouchers</li>
         <li>make requests for mobile data</li>
       <%- end %>
     </ul>
 
-    <%- if FeatureFlag.active?(:rbs_can_manage_users) %>
       <%= govuk_button_link_to 'Invite a new user', new_responsible_body_user_path %>
```
benilovj, line 37:
```suggestion
    <%= govuk_button_link_to 'Invite a new user', new_responsible_body_user_path %>
```

app/mailers/invite_responsible_body_user_mailer.rb
```diff
@@ -0,0 +1,22 @@
+class InviteResponsibleBodyUserMailer < ApplicationMailer
+  def invite_user_email
+    @user = params[:user]
+    rb_name = @user.responsible_body.local_authority_official_name || @user.responsible_body.name
```
benilovj, line 4: How come we're not just using `responsible_body.name`?

any thoughts @fofr? The `local_authority_official name` corresponds to the official name in https://www.registers.service.gov.uk/registers/local-authority-eng

benilovj (approved):

app/mailers/invite_responsible_body_user_mailer.rb
tonyheadford: When I was testing emails, the rb name was Stoke-on-Trent which sounded a bit odd in the context of the email e.g. invite additional users from Stoke-on-Trent instead of Stoke-on-Trent City Council so I went with the long name if we had one.

PR #326, opened 2020-08-26
Refactor the complexity from _nav.html.erb into a view component

aldavidson:
### Context

The nav code was getting increasingly complex as time went on, with increasing amounts of logic in the view template & helpers. Seemed like a good opportunity for a refactor into a view component

Refactor the complexity out of the view template and into a new view component, where the logic can be broken up into comprehensible chunks, and properly tested

This is pretty much a like-for-like refactor - the user should not notice any visible change. But it will make it a lot easier to do additional improvements. For instance, determining the active nav section by checking the current URL for an exact match is not enough anymore, we should probably set the active nav section in controllers and pass that in, etc etc. Those can come in future PRs


tonyheadford (approved):
Looks really good!

PR #328, opened 2020-08-26
[490] Import wave 1 RB contacts from Computacenter CSV file

aldavidson:
### Context

We have some spreadsheets from Computacenter with the contacts they hold for each of their accounts from wave 1.
We need to import those contacts into the DB as users, but also make sure they're linked to the right Responsible Bodies.
The task is made more complex by the fact that the source data we have, uses Computacenter references & GIAS UIDs to link things together.

* Add db fields to store GIAS Id, GIAS Group UID, and computacenter_reference
* Add a new `ImportComputacenterReferencesService` to import Computacenter's 'SoldTo' (RB account ID) and `ShipTo` (school as shipping location) references to our schools & responsible_bodies
* Add a new `ImportLocalAuthorityGiasIdsService` to import GIAS Ids for Local Authorities from @benilovj 's manually-constructed CSV file
* Add  a new `ImportResponsibleBodyUsersFromComputacenterCsvService` to import the contacts
* Add rake task wrappers for all the above importers

### Guidance to review

You'll need to run the importers in this order:

```bash
# re-run existing importer to pick up the GIAS Group UIDs for trusts:
rake import:trusts
rake import:local_authority_gias_ids 
rake import:school_and_rb_computacenter_references CC_REFERENCES_FILE_URI=(file path or URI)
rake import:rb_users_from_cc_csv RB_USERS_CSV_URL=(file path or URI)
```

Ask @aldavidson or @benilovj for the data files if you need to run those last 2 cmds

### DIFF
--- a/app/models/responsible_body.rb
+++ b/app/models/responsible_body.rb
@@ -1,9 +1,13 @@
+require 'computacenter/responsible_body_urns'
+
 class ResponsibleBody < ApplicationRecord
   has_one :bt_wifi_voucher_allocation
   has_many :bt_wifi_vouchers
   has_many :users
   has_many :extra_mobile_data_requests
   has_many :schools
+
+  extend Computacenter::ResponsibleBodyUrns::ClassMethods
 
   def humanized_type
     type.demodulize.underscore.humanize.downcase
@@ -30,35 +34,6 @@ class ResponsibleBody < ApplicationRecord
       'School'
     when 'responsible_body'
       humanized_type.capitalize
-    end
-  end
-
-  def self.find_by_computacenter_urn!(cc_urn)
-    our_identifier = convert_computacenter_urn(cc_urn)
-    # given URNs starting with 't' are for Trusts
-    if cc_urn.starts_with?('t')
-      find_by_companies_house_number!(our_identifier)
-    else
-      # If it's not a Trust, assume it's a school
-      find_by_gias_id!(our_identifier)
-    end
-  end
-
-  # Computacenter's list of RBs has type-dependent prefixes for URNs
-  def self.convert_computacenter_urn(cc_urn)
-    # Trusts start with 't'
-    if cc_urn.starts_with?('t')
-      # take off the leading 't' and pad it to 8 chars with leading zeroes
-      cc_urn[1..-1].rjust(8, '0')
-    elsif cc_urn.starts_with?('LEA')
-      # just remove the leading 'LEA'
-      cc_urn[3..-1]
-    else
-      # if we can't convert it, just return the given string
-      # - that way we can use it as-is in queries and it will raise
-      # an ActiveRecord::RecordNotFound error. If we'd returned nil, we'd
-      # get spurious matches
-      cc_urn
     end
   end
 end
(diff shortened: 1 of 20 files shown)


app/models/responsible_body.rb
```diff
@@ -32,4 +32,27 @@ def who_will_order_devices_label
       humanized_type.capitalize
     end
   end
+
+  def self.find_by_computacenter_urn!(cc_urn)
```
benilovj, line 36: I think we should pull this and the subsequent methods into a module and include it into this class, to group these and make it explicit that these are standalone and are related to CC.

app/services/import_responsible_body_users_from_computacenter_csv_service.rb
```diff
@@ -0,0 +1,65 @@
+require 'open-uri'
+require 'csv'
+
+class ImportResponsibleBodyUsersFromComputacenterCsvService
```
benilovj, line 4: Not a fan of this class name or that it's a service. Could this be a `ComputacenterCsvFile` model with a `import_responsible_bodies` method?

db/schema.rb
```diff
@@ -77,10 +77,10 @@
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
     t.string "status", null: false
-    t.bigint "school_contact_id"
-    t.string "will_need_chromebooks"
     t.string "school_or_rb_domain"
     t.string "recovery_email_address"
+    t.bigint "school_contact_id"
```
benilovj, line 82: @aldavidson you mind re-creating your DB so this noise goes away?

benilovj (commented):
Looking good so far, but it'd be really good to change the services to models that are based around files – this would make it clearer to understand at a glance IMHO.

app/models/responsible_body.rb
aldavidson: Yeah, I originally tried having these in a standalone class in the `Computacenter::` namespace, but it didn't quite work - a module may work better, will give it a try

app/services/import_responsible_body_users_from_computacenter_csv_service.rb
aldavidson: It was for consistency with the rest of the `ImportXYZService` classes - there's a Tech Debt ticket in the backlog to review our usage of the `Service` concept, it's got very muddled.

db/schema.rb
aldavidson: i'll fix it

tonyheadford (approved):
LGTM!

app/models/responsible_body.rb
aldavidson: Done, that makes it cleaner 👍

db/schema.rb
aldavidson: fixed

aldavidson:
> Looking good so far, but it'd be really good to change the services to models that are based around files – this would make it clearer to understand at a glance IMHO.

I agree, just have not had chance to do that - I suggest we do that as a separate PR given the timescales we're working to

benilovj:
> > Looking good so far, but it'd be really good to change the services to models that are based around files – this would make it clearer to understand at a glance IMHO.
> 
> I agree, just have not had chance to do that - I suggest we do that as a separate PR given the timescales we're working to

Yep, agreed it should be a different PR. @tonyheadford has introduced some really nice abstractions that will help.

PR #331, opened 2020-08-27
Add device allocation importer

tonyheadford:
### Context
[Trello card](https://trello.com/c/geNW8AEG/392-import-devices-allocations-into-the-service-from-an-online-csv)
Import school device allocations into the system from a private CSV file.  These are the `std_device` type allocations for years 3-10.  CSV data file is private so will be accessed via a URL held in `ALLOCATIONS_FILE_URL` environment variable.

Add importer for allocations data, refactor downloading of data files into separate `RemoteFile` module.
Importer will add or update existing device allocations on School records.
The importer creates a `WARN` level entry in the Rails logs for any schools specified in the allocations file that we do not currently have in the DB.

### Guidance to review
Assumes that Schools have already been imported and that you have the allocations csv data file.
Run `ALLOCATIONS_FILE_URL="<file location url>" bundle exec rake import:allocations` to import the allocation data
From the rails console you can check allocations with `School.find_by(urn: <urn>).std_device_allocation` and see that the `allocation` field represents the value for the school in the csv data.

### DIFF
--- a/app/models/allocation_data_file.rb
+++ b/app/models/allocation_data_file.rb
@@ -1,0 +1,16 @@
+class AllocationDataFile < CsvDataFile
+  def allocations(&block)
+    records(&block)
+  end
+
+protected
+
+  def extract_record(row)
+    {
+      urn: row['URN'],
+      name: row['Name'],
+      y3_10: row['Y3-Y10'].to_i,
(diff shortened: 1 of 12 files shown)


benilovj (approved):
Nice!

PR #333, opened 2020-08-27
Show JS warning if users try to leave a page with unsaved changes

fofr:
### Context

- Don't let users partially fill in a form then click away (we've seen this happen in research)
- Run this on any form with a textarea or radio button
- Adds jest for JS unit tests

Functionality largely ported from Apply:
https://github.com/DFE-Digital/apply-for-teacher-training/pull/1824

### DIFF
--- a/app/webpacker/packs/warn-on-unsaved-changes.js
+++ b/app/webpacker/packs/warn-on-unsaved-changes.js
@@ -1,0 +1,42 @@
+const getTextArea = () => {
+  return document.querySelector(".govuk-textarea");
+};
+
+const getRadio = () => {
+  return document.querySelector(".govuk-radios__input");
+};
+
+const getFirstEnhanceableForm = () => {
+  const $element = getTextArea() || getRadio();
+  if (!$element) return null;
+
+  return $element.form;
+};
+
+const initWarnOnUnsavedChanges = ($form = getFirstEnhanceableForm()) => {
+  if (!$form) return;
+
+  let hasChanged = false;
+
+  $form.addEventListener("submit", () => {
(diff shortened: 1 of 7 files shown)


tvararu (approved):

app/webpacker/packs/warn-on-unsaved-changes.js
```diff
@@ -0,0 +1,42 @@
...
+  if (!$form) return;
+
+  let hasChanged = false;
+
+  $form.addEventListener("submit", () => {
+    window.onbeforeunload = null;
+  });
+
+  $form.addEventListener("change", () => {
+    hasChanged = true;
```
aliuk2012, line 26: I've never really liked this implementation for this warning because it does actually check whether the value has actually changed for the last time the page loaded. It would be better to do proper state management like store original value and compare the newer value to original)

Scenario:
User visits a page with existing data. They edit one text field (`hasChanged` is set true) but they later decide that it was the correct value and changes it back to the original value(`hasChanged` is STILL set to true). Now if they decide to leave the page they will get a prompt saying they changed data and could potentially lose data even though from their POV nothing has changed.

fofr: While that's ideal, I think having this is still better than no protection, and given this is a tried and tested script I don't want to be changing it much while we're MVPing.

aliuk2012: Totally understand and agree but just mentioning the limitation of this approach 😄

aliuk2012 (approved):

PR #335, opened 2020-08-27
Redirect support users, who are also RB users, to RB homepage

benilovj:
### Context

For support users who want to test functionality as an RB should
be redirected to the RB homepage upon login, otherwise there's no way for them
to navigate to that page. If they want to access support functionality,
they can use the nav links.

If a support user is also something else (e.g. an RB or a CC user too), redirect to that other thing when signing the user in.


aldavidson (approved):
🚢

PR #339, opened 2020-08-27
Show privacy notice to responsible body users

benilovj:
Show the privacy notice to first-time RB users.

### DIFF
--- a/app/controllers/responsible_body/home_controller.rb
+++ b/app/controllers/responsible_body/home_controller.rb
@@ -1,3 +1,10 @@
 class ResponsibleBody::HomeController < ResponsibleBody::BaseController
   def show; end
+
+  def privacy_notice; end
+
+  def seen_privacy_notice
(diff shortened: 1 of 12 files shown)


aldavidson (approved):
Looks OK to me - if I'm reading this right, it's still possible for an RB user who is shown the notice to get to the other pages without clicking Continue, but that may be a design choice (they'll get shown it again next time they log in)

benilovj:
> Looks OK to me - if I'm reading this right, it's still possible for an RB user who is shown the notice to get to the other pages without clicking Continue, but that may be a design choice (they'll get shown it again next time they log in)

yep i think that's ok for now

fofr:
Assume it's ok for RB users who are only in the connectivity pilot to see this too?

benilovj:
> Assume it's ok for RB users who are only in the connectivity pilot to see this too?

I think that's ok.

PR #342, opened 2020-08-27
Update en.yml

gemmaemoore:


benilovj:
I've moved this change onto #341 because they are related.

PR #343, opened 2020-08-27
Send slack events when RBs make choices about who orders

fofr:
Send notifications to Slack when each local authority or trust decides that schools or themselves should manage orders.

Examples:
```
CASTLE SCHOOL EDUCATION TRUST has devolved ordering to schools
CASTLE SCHOOL EDUCATION TRUST will manage orders centrally
```

### DIFF
--- a/app/models/who_will_order_event.rb
+++ b/app/models/who_will_order_event.rb
@@ -1,0 +1,27 @@
+class WhoWillOrderEvent < Event
+  def notifiable?
+    FeatureFlag.active?(:who_will_order_slack_notifications)
+  end
+
+  def message
+    I18n.t(
+      message_key,
+      scope: [:events],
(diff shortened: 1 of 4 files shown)


app/models/who_will_order_event.rb
```diff
@@ -0,0 +1,25 @@
+class WhoWillOrderEvent < Event
+  def notifiable?
+    FeatureFlag.active?(:who_will_order_slack_notifications)
+  end
+
+  def message
+    message_key = if schools_will_order?
```
aldavidson, line 7: Nitpicking,  but could this if statement be moved to a separate method?

```
def message_key
  If schools_will_order?
  (Etc)
end
```

aldavidson (approved):
Looks good! Just one suggestion above, not a blocker

PR #347, opened 2020-08-28
make email address case-insensitive in validation and signing in…

aldavidson:
### Context

[Trello card #555](https://trello.com/c/e5NVQLBI/555-add-a-responsible-body-user-should-not-allow-case-sensitive-email-addresses) - the default implementation of `:string`-type columns in Rails' Postgres adapter is case-sensitive. This means that the email_address field is case-sensitive.

After bulk-importing lots of users from a Computacenter spreadsheet of contacts, we found that 9 of them created duplicates of users who already existed, due to the spreadsheet giving capital letters in their email addresses. In one case, the responsible body listed in the spreadsheet for a user was different to the RB their existing record was linked to. This caused confusion for them when logging in.

* add 2 migrations - one to destroy any duplicate users, and another to add a case-insensitive unique constraint on `users.email_address`
* make the uniqueness validation on `user.email_address` case-insensitive
* force `email_address` to downcase before validating
* add specs to explicitly test for desired behaviour

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -16,7 +16,7 @@ class User < ApplicationRecord
 
   validates :email_address,
             presence: true,
-            uniqueness: true,
+            uniqueness: { case_sensitive: false },
             length: { minimum: 2, maximum: 1024 }
 
   include SignInWithToken
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -48,8 +48,6 @@ class User < ApplicationRecord
   end
 
   def force_email_address_to_lowercase!
-    # the 'self' scope is required, for some reason -
-    # - maybe some artefact of the callback flow?
     self.email_address = email_address.downcase if email_address.present?
   end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -49,8 +49,7 @@ class User < ApplicationRecord
 
   def force_email_address_to_lowercase!
     # the 'self' scope is required, for some reason -
-    # .email_address and @email_address both return nil
     # - maybe some artefact of the callback flow?
-    self.email_address = self.email_address.downcase if self.email_address.present?
+    self.email_address = email_address.downcase if email_address.present?
   end
 end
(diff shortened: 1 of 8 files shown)


app/models/user.rb
```diff
@@ -44,4 +46,10 @@ def needs_to_see_privacy_notice?
   def seen_privacy_notice?
     privacy_notice_seen_at.present?
   end
+
+  def force_email_address_to_lowercase!
+    # the 'self' scope is required, for some reason -
```
benilovj, line 51: I think the self scope is required otherwise Ruby defines it as a local variable, rather than calling the setter method.

app/services/session_service.rb
```diff
@@ -4,7 +4,7 @@ class TokenNotRecognised < StandardError; end
   class InvalidTokenAndIdentifierCombination < StandardError; end
 
   def self.send_magic_link_email!(email_address)
-    if (user = find_user_by_email(email_address))
+    if (user = find_user_by_email(email_address.downcase))
```
benilovj, line 7: The `downcase` shouldn't be necessary, because `find_user_by_email` applies `downcase` too:

```ruby
  def self.find_user_by_email(email_address)
    User.where('lower(email_address) = ?', email_address.downcase).first
  end
```

spec/features/signing_in_as_different_types_of_user_spec.rb
```diff
@@ -44,6 +44,14 @@
     end
   end
 
+  scenario 'signing in should not be case-sensitive in matching the email address (bug 555)' do
```
benilovj, line 47: We have a feature spec for this already: https://github.com/DFE-Digital/get-help-with-tech/blob/main/spec/features/session_behaviour_spec.rb#L126

app/services/session_service.rb
aldavidson: I've removed the `.downcase`, and changed the name of `find_user_by_email` to `find_user_by_lowercase_email` to be more explicit

benilovj (approved):

PR #357, opened 2020-08-28
Added key contact importer and rake task

tonyheadford:
### Context
[Trello card](https://trello.com/c/oeD0JAwt/553-support-process-for-bringing-rbs-into-the-devices-pilot)
Add importer for responsible body key contacts. Imports users from a CSV file, creates the user if not already in the service, sets the user as the key contact for the responsible body and emails the user using the 'Nominate contacts' Notify template.
Sets the `in_devices_pilot` flag on the RB.

Adds importer and data file hander.
Mailer for 'nominate contacts'
Rake task to import csv file manually from a URL
Front-end upload under `support/devices`

### Guidance to review
Run migration
`bundle exec rails db:migrate`

CSV headings for data file
image

Rake task can be run as `KEY_CONTACTS_FILE_URL="<location>" bundle exec rake import:key_contacts`

As a support user, navigate to `/support/devices` to upload via the front-end (I haven't messed with the nav links so no menu item for this as yet)


app/controllers/support/devices_controller.rb
```diff
@@ -0,0 +1,34 @@
+class Support::DevicesController < Support::BaseController
```
aldavidson, line 1: I found the name DevicesController confusing, as this doesn't create Devices - isn't this more like a `Support::Devices::KeyContactController` ? 
I think we're going to be using the Support::Devices namespace for a few more things soon

app/models/user.rb
```diff
@@ -4,6 +4,7 @@ class User < ApplicationRecord
 
   belongs_to :mobile_network, optional: true
   belongs_to :responsible_body, optional: true
+  has_many :key_responsible_bodies, class_name: 'ResponsibleBody', foreign_key: :key_contact_id, inverse_of: :key_contact
```
aldavidson, line 7: so a User can only `belong_to` one ResponsibleBody, but can be key_contact for many RBs? Is that right?
mixing up the one/many like that makes me nervous

Do we even need this relationship on user?

aldavidson (changes requested):
Most of this looks great, but I think the name of the controller is misleading and the key_contact relationship may be over complicated

app/controllers/support/devices_controller.rb
tonyheadford: Yes good call, wasn't sure what to call it while it was the only controller in the namespace

app/models/user.rb
tonyheadford: Yes a bit confusing, I don't __think__ we need it, just thought it might be useful. I'll remove it.

benilovj (approved):
Works a treat, :shipit:

PR #361, opened 2020-08-29
Devices guidance tweaks

benilovj:
### Context

Some absolute links were introduced during recent devices guidance changes. This makes it harder to test those pages because the links aren't relative and take you off to the production site.

- replace absolute links with relative links
- use Rails path helpers where possible (so that if content moves, the path doesn't break)
- fix some indentation issues that have crept in
- extract a view helper for generating links to devices guidance subpages


aldavidson (approved):
LGTM

PR #366, opened 2020-09-01
[557] Split support interface by pilot, and add performance/RBs pages for devices pilot

aldavidson:
### Context

The existing support interface was implicitly all based around the internet connectivity pilot. To allow us to have better insight into the devices pilot, we needed to refactor and split it up into an `internet` and a `devices` section.

* Move all the existing internet-related support UI under `/support/internet`
* Add new UI & models/methods for devices-related support UI
* Refactor the nav to differentiate the two pilots in `/support/`
* Make the two pilot-specific support/responsible body sections point to a single `support/users_controller`

### Guidance to review

There's certainly more that I _could_ have done on this - the nav is not quite as I would like it to be yet, and the `support/devices/responsible_bodies` section could be nicer. But this seemed like a good-enough point to get it reviewed & deployed. All suggestions for improvements welcome.

Screenshots:

Screen Shot 2020-09-01 at 10 27 40
Screen Shot 2020-09-01 at 10 27 30


benilovj (approved):
Looks good - nice work!

tonyheadford (approved):
LGTM

PR #367, opened 2020-09-01
[559] add "Check your school details" page as per prototype

aldavidson:
### Context

See [Trello card 559](https://trello.com/c/SvI6qUl2/559-check-your-school-details-view-only) - first view-only version of a school details page

* add school details page


app/controllers/school/details_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def details_rows
+    [
+      {
+        key: 'Provisional allocation',
+        value: t(:n_devices, scope: %w[school details], count: @school.std_device_allocation&.allocation.to_i),
+      },
+      {
+        key: 'Will your school need to order Chromebooks?',
+        value: @school.preorder_information&.will_need_chromebooks.to_s.capitalize,
```
benilovj, line 16:
```suggestion
        value: @school.preorder_information&.will_need_chromebooks&.to_s&.capitalize,
```
If you're using the `&` operator, you need to apply it to all subsequent chained calls otherwise it will blow up on the next call.

benilovj, line 16:
```suggestion
        value: @school.preorder_information&.will_need_chromebooks&.to_s&.capitalize,
```

app/views/school/details/show.html.erb
```diff
@@ -0,0 +1,41 @@
...
+      About allocations
+    </h2>
+    <p class="govuk-body">Allocations are based on:</p>
+    <ul class="govuk-list govuk-list--bullet">
+      <li>the number of children in years 3 to 10</li>
+      <li>free school meals data</li>
+      <li>an estimate of the number of devices a school already has</li>
+    </ul>
+    <p class="govuk-body">Numbers will be reassessed at the time of ordering based on availability and the extent of coronavirus restrictions.</p>
+    <p class="govuk-body">If you need to query your allocation, email <a href="mailto:COVID.TECHNOLOGY@education.gov.uk">COVID.TECHNOLOGY@education.gov.uk</a></p>
```
benilovj, line 39: can we use the `ghwt_contact_mailto` helper here?

app/views/school/details/show.html.erb
```diff
@@ -0,0 +1,41 @@
...
+
+    <p class="govuk-body">It’s possible to order a range of different devices (laptops, tablets, Chromebooks). If you are going to order Google Chromebooks, we’ll need 2 pieces of information to configure them:</p>
+    <ul class="govuk-list govuk-list--bullet govuk-!-margin-bottom-6">
+      <li>the domain your school uses for its Google services</li>
+      <li>the school’s recovery email address</li>
+    </ul>
+    <p class="govuk-body">We need this information so we can secure the devices.</p>
+
+    <p class="govuk-body">
+      If you need help finding these details, email <a href="mailto:COVID.TECHNOLOGY@education.gov.uk">COVID.TECHNOLOGY@education.gov.uk</a>
```
benilovj, line 27: can we use the `ghwt_contact_mailto` helper here?

benilovj (approved):

app/controllers/school/details_controller.rb
aldavidson: ....but `.to_s` converts nil to empty_string, meaning they're not needed

benilovj: In which case I'd add a `|| ""` clause to make the fall-through default really explicit...

app/views/school/details/show.html.erb
aldavidson: done

app/controllers/school/details_controller.rb
aldavidson: Isn't that just redundant? 

`nil.to_s == ""` has been in Ruby longer than I have (<2006), I'd argue that it's both more expected and more appropriate here than `&.` (which returns a nil you'd just have to handle elsewhere) and more elegant than `(some.chain.of.methods || "").some_other_method`

app/views/school/details/show.html.erb
aldavidson: done!

app/components/school/school_details_summary_list_component.rb
```diff
@@ -0,0 +1,84 @@
...
+      },
+      {
+        key: 'Type of school',
+        value: @school.type_label,
+      },
+      {
+        key: 'Who will order?',
+        value: "The #{(@school.preorder_information || @school.responsible_body).who_will_order_devices_label.downcase} orders devices",
+      },
+    ] + school_contact_row_if_contact_present + chromebook_rows_if_needed
```
fofr, line 30: We don't need:
- school contact row
- who will order row

The only extra rows from the prototype we need are the chromebook ones.

app/views/school/details/show.html.erb
```diff
@@ -0,0 +1,41 @@
...
+      <li>the school’s recovery email address</li>
+    </ul>
+    <p class="govuk-body">We need this information so we can secure the devices.</p>
+
+    <p class="govuk-body">
+      If you need help finding these details, email <%= ghwt_contact_mailto %>
+    </p>
+    <h2 class="govuk-heading-m">
+      About allocations
+    </h2>
```
fofr, line 31: We need to swap these two sections around.

app/components/school/school_details_summary_list_component.rb
```diff
@@ -0,0 +1,61 @@
...
+  def initialize(school:)
+    @school = school
+  end
+
+  def rows
+    [
+      {
+        key: 'Status',
+        value: render(SchoolPreorderStatusTagComponent.new(school: @school)),
+      },
```
fofr, line 17: We don't need the status either.

app/controllers/school/details_controller.rb
benilovj: I'm not going to die in a ditch over it, it's fine to leave as is, I just found it surprising/harder to parse than expected and would find the default more explicit hence flagged it here, but it's a matter of personal preference...

app/views/school/details/show.html.erb
aldavidson: done

app/components/school/school_details_summary_list_component.rb
aldavidson: it's gone

app/components/school/school_details_summary_list_component.rb
aldavidson: done

PR #378, opened 2020-09-02
[475] add a separate field for allocation cap

aldavidson:
### Context

See [Trello card 475](https://trello.com/c/am1Ir4o7/475-separate-the-concept-of-device-allocations-and-cc-caps) - we need to separate the concept of allocations (the maximum amount of devices a school will be able to order in the event of a lockdown) from caps (how many they can order right now, and thus by inference, whether there is a lockdown affecting them at the moment)

* add a `cap` field to `SchoolDeviceAllocation`
* make the `NotifyComputacenterOfCapUpdateJob` send the `cap` field instead of `allocation`

### DIFF
--- a/app/models/computacenter/outgoing_api/cap_update_request.rb
+++ b/app/models/computacenter/outgoing_api/cap_update_request.rb
@@ -1,18 +1,21 @@
 class Computacenter::OutgoingAPI::CapUpdateRequest
-  attr_accessor :endpoint, :username, :password, :allocation_ids
+  attr_accessor :endpoint, :username, :password, :allocation_ids, :timestamp
   attr_accessor :body, :payload_id, :response, :logger
 
   def initialize(args = {})
     @endpoint       = args[:endpoint] || setting(:endpoint)
     @username       = args[:username] || setting(:username)
     @password       = args[:password] || setting(:password)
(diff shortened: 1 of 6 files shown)


benilovj:
It's worrying that no tests break as a result of this change...

aldavidson:
> It's worrying that no tests break as a result of this change...

we have a test on `CapUpdateRequest` that `it 'renders the cap_update_request XML builder template'` but not on the contents of the builder template itself. Agreed I should probably add one

benilovj:
> > It's worrying that no tests break as a result of this change...
> 
> we have a test on `CapUpdateRequest` that `it 'renders the cap_update_request XML builder template'` but not on the contents of the builder template itself. Agreed I should probably add one

Can i suggest we do a webmock spec that goes around the entire machinery and asserts on the produced XML?

aldavidson:
> > > It's worrying that no tests break as a result of this change...
> > 
> > 
> > we have a test on `CapUpdateRequest` that `it 'renders the cap_update_request XML builder template'` but not on the contents of the builder template itself. Agreed I should probably add one
> 
> Can i suggest we do a webmock spec that goes around the entire machinery and asserts on the produced XML?

Done!

benilovj (approved):

PR #380, opened 2020-09-02
add CsvDataFileImporter, and 2 implementations for computacenter imports

aldavidson:
### Context

Relates to two Trello cards: [589-import-computacenter-shipto-identifiers-into-the-service](https://trello.com/c/Tz9AU70M/589-import-computacenter-shipto-identifiers-into-the-service) and [543-refactor-the-importer-services-based-on-tonys-new-abstractions](https://trello.com/c/oggGswho/543-refactor-the-importer-services-based-on-tonys-new-abstractions).

While working out a one-off import of computacenter IDs, it became clear that the simplest way to do it was actually to add a generic `CsvDataFileImporter` class that uses @tonyheadford's `CsvDataFile` abstraction. This makes the implementation code trivial, and also gives us a way to refactor most of the existing `(XYZ)Importer` classes to remove the boilerplate code and simplify them.

* add a `CsvDataFileImporter` class & spec
* add 2 new `CsvDataFile` subclasses for computacenter imports, each of which has an `import_record!` method for use with the  `CsvDataFileImporter` (plus specs)

### DIFF
--- a/app/models/computacenter/ship_to_data_file.rb
+++ b/app/models/computacenter/ship_to_data_file.rb
@@ -1,0 +1,14 @@
+class Computacenter::ShipToDataFile < CsvDataFile
+  def extract_record(row)
+    {
+      urn: row['URN'],
+      name: row['Name 1'],
+      ship_to: row['Ship to acct'],
+    }
+  end
+
+  def import_record!(record)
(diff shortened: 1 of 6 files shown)


tonyheadford (approved):
Nice!

benilovj (approved):

PR #381, opened 2020-09-02
564 schools wizard welcome page and 565-schools-wizard-privacy-notice

tonyheadford:
### Context
[Trello card 564](https://trello.com/c/gYDbcPgc/564-schools-welcome-page)
[Trello card 565](https://trello.com/c/gYDbcPgc/565-schools-wizard-privacy-notice)

Add welcome wizard for new school users with first 2 pages.

Adds `SchoolWelcomeWizard` model and `School::WelcomeWizardController` that navigate a school user through a series of pages and keeps track of progress.

### Guidance to review
The first time a school user signs in they will be redirected to the wizard instead of the 'You're signed in' page.
Unless the wizard has been completed, it will resume at the last page visited if the user signs out/signs in again.
Updates the `:privacy_notice_seen_at` timestamp when progressing from the privacy notice page
No, 'Back' links on the wizard pages, wasn't sure if they were needed?

Sign in token is cleared when moving steps (but it feels a bit icky here, I know @aldavidson had some ideas about refactoring sign-in)


benilovj:
I'm really like this, @tonyheadford! Some thoughts on the privacy notice:

- the privacy notice that was introduced in #339 for responsible bodies, and the content was designed to be applicable to schools too so we should reuse the page content (I guess maybe move the view template to the `shared` folder?)
- would probably be sensible to populate `privacy_notice_seen_at` for school users as well at the point when they click Continue. I guess that setting it would live in the `SchoolWelcomeWizard` model?

benilovj (approved):
Looks good to me! Worth @aldavidson taking a look too...

benilovj:
Looking at the spec failures, I wonder whether we want to apply the `has_completed_wizard` trait for `school_user`s by default... 🤷‍♂️

tonyheadford:
> Looking at the spec failures, I wonder whether we want to apply the `has_completed_wizard` trait for `school_user`s by default... 🤷‍♂️

Might be an idea as it's probably what we want the norm to be, I'll do that now

aldavidson (approved):
Nice implementation!

I think regarding clearing the sign-in token, the best approach is probably to start the wizard after the interstitial token-clearing - but that can be done separately as part of [the sign-in re-jig ticket](https://trello.com/c/aHrdHDJv/548-can-we-rejig-the-sign-in-process-to-fix-the-multiple-event-notifications-issue)

PR #384, opened 2020-09-03
Breadcrumb component

benilovj:
### Context

Breadcrumb markup is repetitive and hard to parse at a glance. We want to abstract it away to clean up view templates.

- Introduce a breadcrumb component and a view helper
- Apply it in the first view

### Guidance to review

Usage example:

```ruby
  <% breadcrumb([
    { t('page_titles.responsible_body_home') => responsible_body_home_path },
    title,
  ]) %>
<% end %>
```

Is the syntactic sugar sensible?

### DIFF
--- a/app/helpers/view_helper.rb
+++ b/app/helpers/view_helper.rb
@@ -30,7 +30,7 @@ module ViewHelper
     govuk_link_to body, devices_guidance_subpage_path(subpage_slug: slug), html_options
   end
 
-  def breadcrumb(items)
+  def breadcrumbs(items)
     render BreadcrumbComponent.new(items)
   end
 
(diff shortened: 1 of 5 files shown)


app/helpers/view_helper.rb
```diff
@@ -30,6 +30,10 @@ def link_to_devices_guidance_subpage(body, slug, html_options = {})
     govuk_link_to body, devices_guidance_subpage_path(subpage_slug: slug), html_options
   end
 
+  def breadcrumb(items)
```
aldavidson, line 33: just one suggestion about the name of this: the singular 'breadcrumb' to me implies that this is for one component of the breadcrumbs/breadcrumb trail. Can I suggest calling it 'breadcrumbs' or 'breadcrumb_trail' instead?

aldavidson (approved):
Looks great, and pretty much exactly what I was thinking. Just one suggestion about the name of the helper method (not a blocker tho)

app/helpers/view_helper.rb
benilovj: yep, good suggestion, fixed in 8d5acc7

PR #386, opened 2020-09-03
[240] Add and setup paper trail gem

asmega:
### Context

- https://trello.com/c/Gi48TlNC/240-put-in-an-auditing-gem-like-audited-or-papertrail-for-the-users-table
- We want to be able to audit changes to the user model so diffs/deltas can be generated
- The `paper_trail` gem should meet our requirements and seems better maintained compared to `audited`
- Below is example of a version from the gem which should be good enough for us to generate deltas

```
pry(main)> u.versions.last
=> #<PaperTrail::Version:0x00007f9516c6e500
 id: 9,
 item_type: "User",
 item_id: 1,
 event: "update",
 whodunnit: "User:1234",
 object:
  "---\nid: 1\n...YAML REPRESENTATION OF OBJECT",
 created_at: Thu, 03 Sep 2020 10:35:37 UTC +00:00,
 object_changes: "---\nupdated_at:\n- 2020-09-03 10:29:05.107686000 Z\n- 2020-09-03 10:35:37.829641000 Z\nis_support:\n- false\n- true\n">
```

### Questions

- `whodunnit` is a `paper_trail` terminology for the user that made then changes 
- Is the concept of `whodunnit` important? Can this be skipped altogether?
- I recall someone mention a CLI. if `whodunnit` is important should I add `whodunnit` to the CLI

- Add and setup `paper_trail` gem
- Hook up `whodunnit` user in the `ApplicationController`

### Guidance to review

- Create a user via the web interface
- Should see audit log for this user via `User.last.versions`
- ?? Make changes to a user via CLI, should see changes with `user.versions`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,6 +24,9 @@ gem 'mail-notify'
 # pagination
 gem 'pagy'
 
+# auditing of activerecord models
+gem 'paper_trail'
+
 # Use postgresql as the database for Active Record
 gem 'pg', '>= 0.18', '< 2.0'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -242,6 +242,9 @@ GEM
     notifications-ruby-client (5.3.0)
       jwt (>= 1.5, < 3)
     pagy (3.8.3)
+    paper_trail (11.0.0)
+      activerecord (>= 5.2)
+      request_store (~> 1.1)
     parallel (1.19.2)
     parser (2.7.1.4)
       ast (~> 2.4.1)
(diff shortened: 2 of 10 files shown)


asmega:
> Looks good! Just one suggestion - can we have versioning turned off in tests by default, like it describes on the [papertrail gem README](https://github.com/paper-trail-gem/paper_trail#7b-rspec) ?
> 
> Otherwise all fine

Done. It's now disabled by default and enabled for the relevant tests

aldavidson (approved):
💯

PR #388, opened 2020-09-03
543 refactor the CSV importer services

aldavidson:
### Context

[Trello card 543](https://trello.com/c/oggGswho/543-refactor-the-importer-services-based-on-tonys-new-abstractions) - we wrote several `XYZImporterService` objects to handle one-off imports of various CSV files. Since then, a new `CSVDataFile` class was introduced which abstracted away the CSV parsing & row-by-row processing, and a new `CsvDataFileImporter` class which abstracted the catching & tracking of errors, etc. This presents a good opportunity to refactor these services and pay down some Tech Debt

* Replace the `ImportComputacenterReferencesService`, `ImportLocalAuthorityGiasIdsService` and `ImportResponsibleBodyUsersFromComputacenterCsvService` objects with `CsvDataFile` subclasses and calls to a new single-method `CsvImportService`
* Add unit tests for the new `CsvDataFile` subclasses
* Move the `import_xyz_service_spec.rb` specs into a new `spec/integration` directory, tweaking where needed to use the new classes
* Tweak the `CsvDataFile` base class to allow opening CSV files from a remote URI as well as local file paths

### Guidance to review

The rake wrappers stay the same, so you should be able to just run the rake tasks as before:

```
bundle exec rake import:local_authority_gias_ids
bundle exec rake import:rb_users_from_cc_csv RB_USERS_CSV_URL=(some file URI)
bundle exec rake import:school_and_rb_computacenter_references CC_REFERENCES_FILE_URI=(some file URI)
```


app/models/csv_data_file.rb
```diff
@@ -35,7 +36,8 @@ def skip?(_row)
 private
 
   def read_file
-    CSV.foreach(@csv_path, headers: true, encoding: 'ISO8859-1:utf-8') do |row|
+    csv = URI.open(@csv_uri).read
+    CSV.parse(csv, headers: true, encoding: 'ISO8859-1:utf-8').select do |row|
```
tonyheadford, line 40: Doing it this way reads and parses the entire file in memory. When reading the schools data file (~57Mb) it appeared to be using 1Gb of memory to do this (are PaaS instances are 2Gb mem?), which is why I went with the download-in-chunks and `CSV.foreach` route.

tonyheadford (approved):
Looks great other than the worry about memory usage

app/models/csv_data_file.rb
tonyheadford: I don't think the `encoding` param does anything with `CSV.parse`

aldavidson: Good call - I've put this back to using `CSV.foreach` in combination with downloading any remote URI to a `Tempfile` (with your `RemoteFile.download`) if needed, and that seems to do the trick. I ran `rake import:schools` while monitoring RAM usage, and it never went over 185M (compared to 1.25G before)

PR #391, opened 2020-09-03
Show schools in support

benilovj:
### Context

Support users need to be able to see the allocations, caps and devices ordered for schools, when queries come in.

Add a schools table to the RB devices pilot support page.

### Guidance to review

image


aldavidson (approved):
This list is going to get long - some RBs have 500+ schools - but we can deal with that separately

PR #395, opened 2020-09-04
rejig the sign-in behaviour to only actually sign the user in once they click the button

aldavidson:
### Context

[Trello card 548](https://trello.com/c/aHrdHDJv/548-can-we-rejig-the-sign-in-process-to-fix-the-multiple-event-notifications-issue) (also many slack threads & Zendesk tickets) - For some users, we've been seeing multiple almost-simultaneous "A user from X has just signed in" notifications in Slack when they sign-in. We think this is due to their email software scanning links in emails for malicious behaviour before the user even sees them.

Rework the sign-in process slightly so that when the user clicks the magic link:
* it renders hidden fields for `token` & `identifier` in to the `DELETE` form 
* it does not actually sign the user in (and send the notification & clear the token, etc) until they click the button to sign in

Adjust the school wizard entry point slightly to fit
Adjust the specs to match & test the new behaviour

**NOTE**: I've altered the wording on the interstitial form page slightly, to make it clearer that they're not signed in until they click the button -

Screen Shot 2020-09-04 at 16 07 54

...but as far as the user is concerned that's the only change they'll see. My interaction design / content design skills are not the best, however, so I'm particularly keen to get @fofr 's input on this. Maybe we should add a `flash[:success]` message after clicking the button, to just give the user some obvious feedback that they really are signed in now?

### Guidance to review

Sign in as normal. Everything should Just Work (TM), including entry to the schools wizard, and re-entry partway through it if you sign out halfway through (@tonyheadford please double-check this)

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -40,7 +40,9 @@
     devices_guidance_index: "Get help with technology: devices"
     devices_guidance_how_to_order: "How to order laptops and tablets during coronavirus (COVID-19)"
     download_bt_hotspots: Download your BT hotspot log-ins
-    click_to_sign_in: You're signed in as %{organisation}
+    click_to_sign_in:
+      standard: You’re signed in
+      related_organisation: You're signed in as %{organisation}
     you_are_signed_in: You’re signed in
     devices_service_performance: Devices service performance
     internet_service_performance: Internet service performance
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1,4 +1,4 @@
-en:
+ en:
   service_name: Get help with technology
   page_titles:
     about_bt_wifi: Increasing children’s internet access through free BT hotspots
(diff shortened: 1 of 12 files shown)


fofr:
Let's go with:

```
You’re signed in as [name of rb or school]

[ Continue ]
```

I don't think the user's journey needs to change at all.

aldavidson:
> Let's go with:
> 
> ```
> You’re signed in as [name of rb or school]
> 
> [ Continue ]
> ```
> 
> I don't think the user's journey needs to change at all.

Sure, just one more question then - what should it show for Computacenter / MNO / Support users?

fofr:
@aldavidson For MNO's the name of the network. For Computacenter and Support what it shows now.

tonyheadford (approved):
LGTM

config/locales/en.yml
```diff
@@ -40,7 +40,9 @@ en:
     devices_guidance_index: "Get help with technology: devices"
     devices_guidance_how_to_order: "How to order laptops and tablets during coronavirus (COVID-19)"
     download_bt_hotspots: Download your BT hotspot log-ins
-    click_to_sign_in: Click the button below to sign in
+    click_to_sign_in:
+      standard: You’re signed in
+      related_organisation: You're signed in as %{organisation}
```
tonyheadford, line 45:
```suggestion
      related_organisation: You’re signed in as %{organisation}
```

tonyheadford (commented):
Wrong apostrophe in the new heading

PR #411, opened 2020-09-07
Use correct status for sums in dashboard

fofr:
`needs_information` is always 0, as the actual status is `needs_info`


benilovj:
I've added some specs to help prevent regressions...

benilovj (approved):

PR #412, opened 2020-09-07
[594] CSV table

asmega:
### Context

- Start point on figuring out how to generate data export for computacenter

- Add migration to form new table which will act as basis for future csv entries
- CSV generator which takes selected users and generates output csv

### Guidance to review

- Take some users and pass to csv generator
- `service = ComputacenterLedgerCsv.new(User.limit(3))`
- Generate the csv
- `service.to_csv`
- output csv should match agreed format

### DIFF
--- a/app/models/responsible_body.rb
+++ b/app/models/responsible_body.rb
@@ -9,6 +9,7 @@ class ResponsibleBody < ApplicationRecord
   has_many :schools
 
   extend Computacenter::ResponsibleBodyUrns::ClassMethods
+  include Computacenter::ResponsibleBodyUrns::InstanceMethods
 
   def humanized_type
     type.demodulize.underscore.humanize.downcase
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -83,4 +83,16 @@ class User < ApplicationRecord
       (is_computacenter? && 'Computacenter') || \
       (is_support? && 'DfE Support')
   end
+
+  def first_name
+    (full_name || '').strip.split(' ').first.to_s
+  end
+
+  def last_name
+    (full_name || '').strip.split(' ').last.to_s
+  end
+
+  def effective_responsible_body
+    responsible_body || school&.responsible_body
+  end
 end
--- a/app/models/computacenter/ledger.rb
+++ b/app/models/computacenter/ledger.rb
@@ -1,0 +1,82 @@
+require 'csv'
+
+module Computacenter
+  class Ledger
+    def initialize(users: nil)
+      @users = users
+    end
+
+    def to_csv
+      CSV.generate do |csv|
+        csv << headers
+
+        user_changes.each do |user_change|
+          csv << [
+            user_change.first_name,
+            user_change.last_name,
+            user_change.email_address,
+            user_change.telephone,
+            user_change.responsible_body,
+            user_change.responsible_body_urn,
+            user_change.cc_sold_to_number,
+            user_change.school,
+            user_change.school_urn,
+            user_change.cc_ship_to_number,
+            user_change.updated_at_timestamp.utc.strftime('%d/%m/%Y'),
+            user_change.updated_at_timestamp.utc.strftime('%R'),
+            user_change.updated_at_timestamp.utc.iso8601,
+            user_change.type_of_update,
+            user_change.original_email_address,
+          ]
+        end
+      end
+    end
+
+  private
+
+    def users
+      @users ||= User.who_can_order_devices.where('privacy_notice_seen_at IS NOT NULL').limit(1)
+    end
+
+    def user_changes
+      users.map do |user|
+        UserChange.new(
+          user_id: user.id,
+          first_name: user.first_name,
+          last_name: user.last_name,
+          email_address: user.email_address,
+          telephone: user.telephone,
+          responsible_body: user.effective_responsible_body.name,
+          responsible_body_urn: user.effective_responsible_body.computacenter_identifier,
+          cc_sold_to_number: user.effective_responsible_body.computacenter_reference,
+          school: user.school&.name,
+          school_urn: user.school&.urn,
+          cc_ship_to_number: user.school&.computacenter_reference,
+          updated_at_timestamp: user.created_at,
+          type_of_update: 'New',
+          original_email_address: nil,
+        )
+      end
+    end
+
+    def headers
+      [
+        'First Name',
+        'Last Name',
+        'Email',
+        'Telephone',
+        'Responsible Body',
+        'Responsible Body URN',
+        'CC Sold To Number',
+        'School',
+        'School URN',
+        'CC Ship To Number',
+        'Date of Update', # "24/08/2020" / "dd/mm/yyyy"
+        'Time of Update', # 24 hour clock
+        'Timestamp of Update', # ISO
+        'Type of Update', # New / Change / Remove
+        'Original Email',
+      ]
+    end
+  end
+end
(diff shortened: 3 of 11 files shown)


app/services/computacenter_ledger_csv.rb
```diff
@@ -0,0 +1,101 @@
...
+    when 'LocalAuthority'
+      "LEA#{responsible_body.gias_id}"
+    when 'Trust'
+      "t#{responsible_body.companies_house_number.to_i}" # QUESTION: is `to_i` needed? is it to remove padded zeros?
+    end
+  end
+
+  # TODO: will be moved to model
+  def first_name(user)
+    (user.full_name || '').strip.split(' ').first
```
asmega, line 49: is this ok?

app/services/computacenter_ledger_csv.rb
```diff
@@ -0,0 +1,101 @@
...
+private
+
+  def computacenter_identifier(responsible_body)
+    return nil if responsible_body.nil?
+
+    case responsible_body.type
+    when 'LocalAuthority'
+      "LEA#{responsible_body.gias_id}"
+    when 'Trust'
+      "t#{responsible_body.companies_house_number.to_i}" # QUESTION: is `to_i` needed? is it to remove padded zeros?
```
asmega, line 43: can you confirm this is correct?

benilovj: Yeah it's to remove the leading zero if one's there...

app/services/computacenter_ledger_csv.rb
```diff
@@ -0,0 +1,101 @@
...
+
+  def ledger_entries
+    users.map do |user|
+      ComputacenterUserLedger.new(
+        user_id: user.id,
+        first_name: first_name(user),
+        last_name: last_name(user),
+        email: user.email_address,
+        telephone: user.telephone,
+        responsible_body: user.responsible_body&.name,
```
benilovj, line 69: This isn't quite right – the responsible body should always present but is differently derived based on whether user is a school user or an RB users:

`responsible_body: (user.responsible_body || user.school.responsible_body).name`

asmega: ok this is now a method the user `derived_responsible_body`, not sure if that's the best name

db/migrate/20200907114036_create_computacenter_user_ledger.rb
```diff
@@ -0,0 +1,25 @@
+class CreateComputacenterUserLedger < ActiveRecord::Migration[6.0]
+  def change
+    create_table :computacenter_user_ledgers do |t|
+      t.integer :user_id
+      t.text :first_name
+      t.text :last_name
+      t.text :email
```
benilovj, line 7: Can we please change this for consistency?

```suggestion
      t.text :email_address
```

db/migrate/20200907114036_create_computacenter_user_ledger.rb
```diff
@@ -0,0 +1,25 @@
...
+      t.text :telephone
+      t.text :responsible_body
+      t.text :responsible_body_urn
+      t.text :cc_sold_to_number
+      t.text :school
+      t.text :school_urn
+      t.text :cc_ship_to_number
+      t.datetime :updated_at_timestamp
+      t.text :type_of_update
+      t.text :original_email
```
benilovj, line 17: Can we please change this for consistency?

```suggestion
      t.text :original_email_address
```

app/models/computacenter_user_ledger.rb
```diff
@@ -0,0 +1,2 @@
+class ComputacenterUserLedger < ApplicationRecord
```
benilovj, line 1: Ledger generally refers to the whole book of record, not a single record, so the naming could potentially be misconstrued. Maybe something like `ComputacenterUserChange` or `ComputacenterUserUpdateEntry`. Naming is hard...

app/models/computacenter_user_ledger.rb
```diff
@@ -0,0 +1,2 @@
...
+end
```
benilovj, line 2: `type_of_update` should probably be an enum defined here?

app/models/user.rb
```diff
@@ -83,4 +83,16 @@ def organisation_name
...
+  def first_name
+    (full_name || '').strip.split(' ').first.to_s
+  end
+
+  def last_name
+    (full_name || '').strip.split(' ').last.to_s
+  end
+
+  def derived_responsible_body
+    responsible_body || school.responsible_body
```
benilovj, line 96: Since we're defining this for all users (who can have neither responsible body nor school e.g. support users), we should guard for this:

```suggestion
    responsible_body || school&.responsible_body
```

app/services/backfill_computacenter_ledger.rb
```diff
@@ -0,0 +1,3 @@
+class BackfillComputacenterLedger
```
benilovj, line 1: not sure whether we need this initially

asmega: This is now gone

app/models/user.rb
asmega: done

app/models/computacenter_user_ledger.rb
asmega: Yea i thought about it for a while, I've now changed to `ComputacenterUserChange `

db/migrate/20200907114036_create_computacenter_user_ledger.rb
asmega: sure, maybe we can alias everything to `email` in the near future?

app/models/responsible_body.rb
```diff
@@ -38,6 +38,15 @@ def who_will_order_devices_label
...
 
+  def computacenter_identifier
+    case type
+    when 'LocalAuthority'
+      "LEA#{gias_id}"
+    when 'Trust'
+      "t#{companies_house_number.to_i}"
+    end
+  end
+
```
aldavidson, line 49: When we did a previous round of work involving these URN translations, we decided to put the code in a separate module (`app/models/computacenter/responsible_body_urns.rb`), both to keep the core model clean, and to help quantify how much of the code is purely for Computacenter integration. 

ResponsibleBody currently has the line: `extend Computacenter::ResponsibleBodyUrns::ClassMethods` - for consistency, could this method be moved to an `InstanceMethods` module in `app/models/computacenter/responsible_body_urns.rb`, which ResponsibleBody can then include?

app/services/computacenter_ledger_csv.rb
aldavidson: we've used `effective_xyz` elsewhere, so this could be `effective_responsible_body`

app/models/computacenter_user_change.rb
```diff
@@ -0,0 +1,3 @@
+class ComputacenterUserChange < ApplicationRecord
```
aldavidson, line 1: Where we have a `Computacenter` prefix, could this actually be in the `Computacenter::` namespace (& corresponding sub-directory) instead?

app/services/computacenter_ledger_csv.rb
```diff
@@ -0,0 +1,80 @@
+require 'csv'
+
+class ComputacenterLedgerCsv
```
aldavidson, line 3: As above, where we have a `Computacenter` prefix, could this actually be in the `Computacenter::` namespace (& corresponding sub-directory) instead?

Also, having a `to_csv` method on something which is already a `(something)Csv` feels awkward. 
How about calling this just `Computacenter::Ledger`? 
You'd then be initializing the ledger with some users - `ledger = Computacenter::Ledger.new(users: some_users)` - and generating CSV with `ledger.to_csv` ?

Naming things _is_ hard :)

aldavidson (changes requested):

aldavidson (approved):
👍

PR #413, opened 2020-09-07
[569] Add additional user flow to wizard

tonyheadford:
### Context
[Trello card](https://trello.com/c/EwtiVEL2/569-schools-wizard-user-wont-order-but-nominates-someone-else)
Gives the user the choice to add an additional user during the welcome wizard

Refactored school invite form to shared partial
User can add new school user via the wizard
### Guidance to review

School user must be the only (first) user for the school in the service
Selecting 'Yes' opens the invite user form
New user will be sent the School nominated contact email on creation
I've not implemented the "Back" link and I'm not sure how we'd handle this step at all if you'd already added a user and tried to go back - it would probably need to skip the page as the user would no longer be the only user for the school in the service - but additional content may be required.
The `id` of the new user is stored in the `SchoolWelcomeWizard.invited_user_id` attribute. I ended up jumping through hoops with this and ended up not using a relation to access this user with but it is accessible via the `invited_user` method.

image

### DIFF
--- a/app/controllers/school/welcome_wizard_controller.rb
+++ b/app/controllers/school/welcome_wizard_controller.rb
@@ -7,7 +7,7 @@ class School::WelcomeWizardController < School::BaseController
     if @wizard.update_step!(wizard_params)
       redirect_to next_step_path
     else
-      render @wizard.step
+      render @wizard.step, status: :unprocessable_entity
     end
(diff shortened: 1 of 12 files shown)


benilovj (approved):
Looks great...

PR #415, opened 2020-09-07
[571] schools wizard chromebooks

tonyheadford:
### Context
[Trello card](https://trello.com/c/5L2CpbJX/571-schools-wizard-chromebooks)
Ask if school needs Chromebooks

Adds page to wizard to ask if school needs Chromebooks
Assumes that if a school gets here that a `preorder_information`model has already been created - **I don't know if this is the case?**
Refactor of chromebook form object partials to conditionally add extra options and wording to cater for this view.

### Guidance to review
This page is only shown if `:will_need_chromebooks` is `nil` in the school's `preorder_information` model

image

### DIFF
--- a/app/views/shared/_chromebook_information_intro.html.erb
+++ b/app/views/shared/_chromebook_information_intro.html.erb
@@ -1,4 +1,7 @@
 <p class="govuk-body">
+<% if local_assigns[:show_different_devices_text] %>
+  It’s possible to order a range of different devices (laptops, tablets, Chromebooks).
+<%- end %>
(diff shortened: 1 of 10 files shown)


app/views/shared/_chromebook_information_intro.html.erb
```diff
@@ -1,4 +1,7 @@
 <p class="govuk-body">
+<% if local_assigns[:show_different_devices_text] %>
```
benilovj, line 2: I wonder whether we always show this microcopy, rather than making it conditional? @fofr

benilovj (approved):
👍  one non-blocking comment, that we can resolve in a follow-up PR if needed.

PR #422, opened 2020-09-08
[594] Add extra cols to computacenter_user_changes

asmega:
### Context

- Add extra columns to `computacenter_user_changes` as specified in the `v3` spreadsheet

- Add extra db columns to `computacenter_user_changes`
- CSV generator takes these extra columns into consideration

### Guidance to review

- Start service with users
- `service = Computacenter::Ledger.new(users: User.all)`
- Generate csv from service
- `service.to_csv`
- Should generate csv with extra columns


benilovj (approved):

aldavidson:
@asmega looks good, just note that in #423 I moved the `headers` method to a public class scope so that I could reference it in tests

PR #426, opened 2020-09-08
[594] Add service to backfill user_changes

asmega:
### Context

- Add service to backfill user changes table

- New service to backfill user changes table
- Should be idempotent so we can run multiple times to append to the table till we have a better mechanism in place
- We ignore users that have not read the privacy policy or can order devices 

### Guidance to review

- With some test users present in local database
- Call service with `Computacenter::BackfillLedger.new.call`
- Table should should ledger with new users dated in the past

### DIFF
--- a/app/models/computacenter/backfill_ledger.rb
+++ b/app/models/computacenter/backfill_ledger.rb
@@ -1,0 +1,43 @@
+require 'csv'
+
+module Computacenter
+  class BackfillLedger
+    def call
+      users.each do |user|
+        next if UserChange.where(user_id: user.id).exists?
+
+        UserChange.create!(
+          user_id: user.id,
+          first_name: user.first_name,
+          last_name: user.last_name,
+          email_address: user.email_address,
+          telephone: user.telephone,
+          responsible_body: user.effective_responsible_body.name,
+          responsible_body_urn: user.effective_responsible_body.computacenter_identifier,
+          cc_sold_to_number: user.effective_responsible_body.computacenter_reference,
+          school: user.school&.name,
+          school_urn: user.school&.urn,
+          cc_ship_to_number: user.school&.computacenter_reference,
+          updated_at_timestamp: user.created_at,
+          type_of_update: 'New',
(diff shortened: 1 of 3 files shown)


app/models/computacenter/backfill_ledger.rb
```diff
@@ -0,0 +1,43 @@
...
+          original_email_address: nil,
+          original_telephone: nil,
+          original_responsible_body: nil,
+          original_responsible_body_urn: nil,
+          original_cc_sold_to_number: nil,
+          original_school: nil,
+          original_school_urn: nil,
+          original_cc_ship_to_number: nil,
+        )
+      end
```
aldavidson, line 34: This code looks almost identical to the `Computacenter::Ledger.user_changes` method (similarly the `users` method below) - would it make more sense for this whole model to actually just be a `backfill` method on `Computacenter::Ledger` ?

aldavidson: you could then take the code which constructs a UserChange for a given User and make it either a class method on UserChange (something like `UserChange.for_user(user)`) or an instance method on user (something like `user.to_user_change`)

aldavidson (changes requested):

app/models/computacenter/backfill_ledger.rb
asmega: `Computacenter::Ledger#user_changes` will be swapped out in the near future. It will probably read some like
```ruby
def user_changes
  Computacenter::UserChange.all
end
```
It is a stop gap as that table isn't available yet so we're reading in users for now.

I can move to `UserChange.for_user(user)` or `user.to_user_change` but didn't want to pollute those classes. Also a `User` can't really be converted to a "change" as a user has a current state but a "change" is a change between states.

I feel like one we have a process in place we can probably delete this whole file

aldavidson: OK, that makes sense 👍

aldavidson (approved):

PR #427, opened 2020-09-08
Enable moving backwards in wizard

tonyheadford:
### Context
[Trello card](https://trello.com/c/oB609ypc/629-allow-users-to-use-the-browsers-back-button-in-the-school-wizard)

Pass wizard step to update method to enable back button to work as expected
Invite additional user only available to "first" school user
Chromebooks page only shown to "first" school user

### Guidance to review
Will need a good check of the flows through the wizard and using the back button to repeat steps.  Check that the user journey still makes sense.

### DIFF
--- a/app/controllers/school/welcome_wizard_controller.rb
+++ b/app/controllers/school/welcome_wizard_controller.rb
@@ -3,8 +3,10 @@ class School::WelcomeWizardController < School::BaseController
   before_action :resume_wizard, except: :next_step
 
   def next_step
-    clear_user_sign_in_token!
-    if @wizard.update_step!(wizard_params)
+    current_step = params.fetch(:step, @wizard.step)
+
+    # clear_user_sign_in_token!
+    if @wizard.update_step!(wizard_params, current_step)
       redirect_to next_step_path
     else
       render @wizard.step, status: :unprocessable_entity
--- a/app/models/school_welcome_wizard.rb
+++ b/app/models/school_welcome_wizard.rb
@@ -4,7 +4,6 @@ class SchoolWelcomeWizard < ApplicationRecord
   validates :step, presence: true
 
   enum step: {
-    welcome: 'welcome',
     privacy: 'privacy',
     allocation: 'allocation',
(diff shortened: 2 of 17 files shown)


fofr:
> Chromebooks page only shown to "first" school user
Why did we decide this?

tonyheadford:
> > Chromebooks page only shown to "first" school user
> 
> Why did we decide this?

I was trying to keep the movement through the wizard consistent, happy to change it if you want.  My thinking - the `:will_need_chromebooks` attribute has 3 states `yes`, `no` and `nil` (used for 'I don't know'). The first user has to make a choice.  That user can now go backwards and change that choice while in the wizard - previously the page was only shown if `:will_need_chromebooks` was `nil`.  However, now that we can go backwards, If we only show the page when that value is `nil` it means the user cannot get to it again while in the wizard - or rather, if they go backwards to the step before it and then try to step into it again, it would be skipped which makes the journey inconsistent?

fofr:
We need both:
- back/forward must work for users who see this page
- we need to show it to subsequent users if the value isn't set

fofr:
When a user signs in, can we decide what pages they see in the wizard? Rather than at the point of navigation.

tonyheadford:
I kind of do that already for the first user but I'm thinking if we extend
that to the chromebooks and store that in the wizard we can control
access/visibility forwards and backwards to that page

benilovj (approved):
Code changes look good, deferring to @fofr on the correct behaviour...

tonyheadford:
@fofr updated so now it records the decision on whether the user should be able to see the chromebooks page based on whether the questions has already been answered or not. Once that use has been given access to the chromebooks page, they can navigate forwards and backwards through the wizard and still see the page regardless of whether they have answered the question or not.

fofr:
Back and forward seem to work ok now.
If you use back or forward and then refresh the journey behaves strangely again (going to the latest not-seen page).

I think we need to rethink this because we shouldn't be breaking browser features like this, but for now we need to ship.

fofr (approved):

PR #428, opened 2020-09-08
Show 4G dongle allocations/caps/order counts in support

benilovj:
### Context

The support interface needs to soon support the ability to manage allocations, caps and order counts for both devices and dongles. We start this process by displaying the current settings for these bits.

- display the device allocations/caps/order counts in one column
- display the 4G dongle allocations/caps/order counts as well
- set a column width on the 'Name and URN' column

### Guidance to review

image


fofr (approved):
Design looks good

PR #429, opened 2020-09-08
Inviting schools onto the service via support

benilovj:
~**Includes #428, rebase once that PR is merged.**~ Done

### Context

We want to start bringing schools (who will be responsible for ordering devices) onto the service. We will initially do this via the support interface.

Add an 'invite' button for schools that are ready to be contacted. This button:

- adds the school contact as a user into the system
- emails them an invite
- updates the school's status
- only does something if it's possible to add the user (e.g. the user isn't already on the system for whatever reason)

Additionally, this PR updates the school statuses for the RBs so they're correct as the school users go through the welcome wizard.

### Guidance to review

image

image

image

### DIFF
--- a/app/models/preorder_information.rb
+++ b/app/models/preorder_information.rb
@@ -24,13 +24,15 @@ class PreorderInformation < ApplicationRecord
     set_defaults
   end
 
-  # Update this method as we add more fields (e.g. chromebook info)
-  # with reference to the prototype:
-  # https://github.com/DFE-Digital/increasing-internet-access-prototype/blob/master/app/views/responsible-body/devices/school/_status-tag.html
+  # If this method is added, we may need to update School::SchoolDetailsSummaryListComponent
   def infer_status
-    if school_will_order_devices?
-      school_contact.present? ? 'school_will_be_contacted' : 'needs_contact'
-    elsif orders_managed_centrally? && chromebook_information_complete?
+    if school_will_order_devices? && school_contact.nil?
+      'needs_contact'
+    elsif school_will_order_devices? && school_contacted_at.nil?
+      'school_will_be_contacted'
+    elsif school_will_order_devices? && school_contacted_at.present? && !chromebook_information_complete?
+      'school_contacted'
+    elsif chromebook_information_complete?
       'ready'
     else
       'needs_info'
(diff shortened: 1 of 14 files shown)


app/models/preorder_information.rb
```diff
@@ -24,13 +24,15 @@ def initialize(*args)
...
+    if school_will_order_devices? && school_contact.nil?
+      'needs_contact'
+    elsif school_will_order_devices? && school_contacted_at.nil?
+      'school_will_be_contacted'
+    elsif school_will_order_devices? && school_contacted_at.present? && !chromebook_information_complete?
+      'school_contacted'
+    elsif chromebook_information_complete?
       'ready'
     else
       'needs_info'
```
benilovj, line 38: You happy with this logic @fofr?

fofr: Do we also need:

```ruby
    elsif school_will_order_devices? && school_contacted_at.present? && chromebook_information_complete?
      'school_ready'
```

app/models/school.rb
```diff
@@ -67,4 +67,12 @@ def can_order_devices?(device_type = 'std_device')
     allocation = device_allocations.by_device_type(device_type).first
     allocation&.cap.to_i > allocation&.devices_ordered.to_i
   end
+
+  def invite_school_contact_if_possible!
```
aldavidson, line 71: just one minor nit-pick about this name - having `(...)_if_possible` combined with the bang feels inconsistent, like the bang implies it will raise an exception if it's not possible rather than return false (in line with the usual Rails `update` vs `update!` convention). I know there's the _other_ convention at play, that the bang means it's modifying internal state (which it is), so maybe it's just me :/

aldavidson (approved):
Looks great, only one comment about the method name but it's not a blocker. :shipit:

app/models/school.rb
benilovj: I'll drop the `if_possible!` from the method name because I agree it doesn't add much and looks strange.

What would you suggest doing about the bang? I could remove it since the method doesn't throw exceptions, and the name should be clear enough that it's state-modifying...

aldavidson: Yeah, TBH I think just `invite_school_contact` does the trick

benilovj: OK, as discussed on Slack, I've renamed the method to `invite_school_contact` in 7aa1c3a

app/models/preorder_information.rb
benilovj: Sure, added in 771f4c5

PR #434, opened 2020-09-09
Update guidance on requesting devices

fofr:
- Clarify when evidence is needed, relating to shielding and geographical regions
- Specify that we need to know the year groups
- Highlight that devices can be for any year group
- Move text into shared partial

https://trello.com/c/I2YoCXcE/624-clarify-guidance-on-shielding


emmajhf (approved):

tonyheadford (approved):
LGTM

PR #437, opened 2020-09-09
586 add validation for chromebook settings

tonyheadford:
### Context
[Trello card](https://trello.com/c/r20L12bz/586-add-validation-for-chromebook-settings)

Added G Suite domain validator
Added validations for Chromebook details in the`ChromebookInformationForm` model.
Validate `:school_or_rb_domain` with G Suite domain validator
Use EmailAddressValidator for `:recovery_email_address`
Added method to check for a service login which catches all of the domains that the MX check does and more. I've switched to using this for the validator.

### Guidance to review
The domain and email address validations are triggered when `:will_need_chromebooks` is set to 'yes'.
The domain must now be a valid G Suite domain.

Domains can be checked from the rails console with `Gsuite.is_gsuite_domain?('some-domain.com')` which returns `true` or `false` - this uses the service login check.

You can check if a domain has a google MX DNS record with `Gsuite.has_google_mx_record?('some-domain.com')`
You can check if a domain has a service login with `Gsuite.has_service_login?('some-domain.com')`

### DIFF
--- a/app/services/gsuite.rb
+++ b/app/services/gsuite.rb
@@ -24,7 +24,7 @@ module Gsuite
     if response.status.success?
       body = response.body.to_s
       logger.debug body if logger
-      body =~ /Sign in - Google Accounts/
+      !(body =~ /Sign in - Google Accounts/).nil?
     elsif response.status.redirect?
       # the redirects all seem to be to SAML providers for logins
       logger.debug response.body.to_s if logger
--- a/app/services/gsuite.rb
+++ b/app/services/gsuite.rb
@@ -3,11 +3,13 @@ require 'http'
 module Gsuite
   class DomainLookupError < StandardError; end
 
-  def self.is_gsuite_domain?(domain)
+  def self.is_gsuite_domain?(domain, logger = nil)
     response = HTTP.get("https://dns.google.com/resolve?name=#{domain}&type=MX")
     if response.status.success?
       payload = JSON.parse(response.body)
-      payload['Answer']&.any? { |h| h['data'] =~ /\.google\.com/i }
+      logger.debug(payload) if logger
+
+      payload['Answer']&.any? { |h| h['data'] =~ /\.google(mail)?\.com/i }
     else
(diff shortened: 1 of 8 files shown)


app/services/gsuite.rb
```diff
@@ -0,0 +1,15 @@
+require 'http'
+
+module Gsuite
+  class DomainLookupError < StandardError; end
+
+  def self.is_gsuite_domain?(domain)
+    response = HTTP.get("https://dns.google.com/resolve?name=#{domain}&type=MX")
+    if response.status.success?
+      payload = JSON.parse(response.body)
+      payload['Answer']&.any? { |h| h['data'] =~ /\.google\.com/i }
```
aldavidson, line 10: Should also look out for older gsuite domains for which the MX records point to googlemail.com

aldavidson (approved):
very nice! Just the comment about allowing googlemail.com, otherwise this is great 👍

app/services/gsuite.rb
tonyheadford: Good call, will add that in now

benilovj:
Build should pass after #440 is merged and this PR is rebased.

benilovj:
I've rebased this and going to merge it in, since we're doing a big push on getting schools on. It's a tradeoff but we will need to quickly fix any `PreorderInformation` records that are invalidated by this.

app/services/gsuite.rb
```diff
@@ -0,0 +1,36 @@
...
+      raise DomainLookupError, "Domain lookup failed: #{response.status}"
+    end
+  end
+
+  def self.has_service_login?(domain, logger = nil)
+    response = HTTP.get("https://www.google.com/a/#{domain}/ServiceLogin")
+    if response.status.success?
+      body = response.body.to_s
+      logger.debug body if logger
+      !(body =~ /Sign in - Google Accounts/).nil?
```
fofr, line 27: Just a word of warning on this - we're checking for text on a Google page design that gets updated _relatively_ often. Whereas checking for the error message that's shown might be more future proof – that error page is super old and hasn't changed in years.

PR #439, opened 2020-09-09
Remove 'will you order' step from the school welcome wizard

benilovj:
### Context

We want to avoid the risk of the school ending up without anyone
who can place orders. Since this is quite easy to do if the first
school user opts out of having a TechSource account and then doesn't
add anyone else.

It will still be possible for the first school user to opt out of having
the TechSource account by editing their account.

- Remove the 'Will you order' step from the school wizard
- Update the DB, in case any user happened to fall out of the wizard on that step


benilovj:
Build should pass after #440 is merged and this PR is rebased.

tonyheadford (approved):

PR #444, opened 2020-09-10
MNO requests should always have a responsible body

benilovj:
### Context

Had this branch knocking about for some time and wanted to get it in.

Enforce the existence of a responsible body on MNO requests at the DB level, for data consistency.

### Guidance to review

I've checked and there are no MNO requests without an RB on any of our environments, so no data migration is needed.


aldavidson (approved):
👍

PR #455, opened 2020-09-11
[594] Add guard to backfill

asmega:
### Context

- https://trello.com/c/n9stv3jD/594-user-changes-feed-for-computacenter

- Add guard incase user does not have a responsible body

### Guidance to review

- Backfill a user without a responsible body, should not raise an error


asmega:
turns out this isn't needed

PR #457, opened 2020-09-12
Support: show who will be invited from the school

benilovj:
### Context

Currently, when inviting a school, there's no way to see who you are inviting, in case you need to follow up with them for research.

- change the invite button to a link
- show an invitation confirmation page, which shows who will be invited

### Guidance to review

image

image

### DIFF
--- a/app/controllers/support/devices/schools_controller.rb
+++ b/app/controllers/support/devices/schools_controller.rb
@@ -4,6 +4,11 @@ class Support::Devices::SchoolsController < Support::BaseController
     @users = @school.users
   end
 
+  def confirm_invitation
+    @school = School.find_by!(urn: params[:school_urn])
+    @school_contact = @school.preorder_information&.school_contact
+  end
+
   def invite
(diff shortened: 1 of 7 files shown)


app/controllers/support/devices/schools_controller.rb
```diff
@@ -4,6 +4,11 @@ def show
     @users = @school.users
   end
 
+  def confirm_invitation
+    @school = School.find_by!(urn: params[:school_urn])
+    @school_contact = @school.preorder_information&.school_contact
```
asmega, line 9: if school contact is `nil` then `@school_contact` will be set to `nil`. This will cause an error in the view eg when `@school_contact.title.present` is called. I don't know if this will ever be nil though as I lack some knowledge there?

benilovj: Thanks. Fixed in 8892914

asmega (approved):

PR #458, opened 2020-09-12
Show school details in support

benilovj:
~~**This PR builds upon #457, so merge that one first and rebase this one.**~~ ✅ 

### Context

Support agents need to see how far individual schools have gotten in the pre-order journey.

### Changes proposed in this pull request

Display school details in support, but without any change links.
Provide a link to invite the school contact when that's appropriate.

### Guidance to review

Best reviewed commit-by-commit.

**When a school is ready to be invited**

image

**When the RB hasn't even made the decision to centralise or devolve orders**

image

PR #459, opened 2020-09-13
Introduce school order state

benilovj:
**Depends on #458, rebase this PR after that one is merged.**

### Context

Schools and responsible bodies need to know when they can start ordering devices, and whether that's for specific circumstances or local coronavirus restrictions. This will be tracked on a per-school basis.

- Introduce an order state on the school's pre-order information.
- Update the responsible body's school details summary table to show appropriate guidance in the "Can place orders?" section.

### Guidance to review

It's awkward having the order state on the **pre**order information, but it feels like that's the right place for the information. It's worth considering whether we want to rename `PreorderInformation`, to reflect its increased scope, as a subsequent change.

#### Screenshots

Is the microcopy correct? cc @emmajhf

**Cannot order**

image

**Can order under local lockdown**

image

** Can order for specific circumstances**

image

### DIFF
--- a/app/components/responsible_body/school_details_summary_list_component.rb
+++ b/app/components/responsible_body/school_details_summary_list_component.rb
@@ -12,36 +12,58 @@ class ResponsibleBody::SchoolDetailsSummaryListComponent < ViewComponent::Base
 
   def rows
     [
-      {
-        key: 'Status',
-        value: render(SchoolPreorderStatusTagComponent.new(school: @school)),
-      },
-      {
-        key: 'Who will order?',
-        value: "The #{(@school.preorder_information || @school.responsible_body).who_will_order_devices_label.downcase} orders devices",
-        change_path: responsible_body_devices_school_change_who_will_order_path(school_urn: @school.urn),
-        action: 'who will order',
-      },
-      {
-        key: 'Provisional allocation',
-        value: pluralize(@school.std_device_allocation&.allocation.to_i, 'device'),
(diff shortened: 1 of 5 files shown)


emmajhf:
@benilovj I think the microcopy is good. I have one suggested change to 'Can order for specific circumstances': I'd shorten it to 'Yes, for specific circumstances'.

benilovj:
> @benilovj I think the microcopy is good. I have one suggested change to 'Can order for specific circumstances': I'd shorten it to 'Yes, for specific circumstances'.

Thanks @emmajhf, I've incorporated your suggestion.

tonyheadford (approved):
Looks great.  I wonder whether the `order_state` would fit better on the `School` if you think of it more as the state of the school is in - has shielding pupils or wider restrictions - rather than directly  associated with ordering as such - might need to be renamed to `covid_state` or something better?
Also I think some of the copy you're using looks very similar to what I need so might be better off in the `en.yml`?

benilovj:
> Looks great. I wonder whether the `order_state` would fit better on the `School` if you think of it more as the state of the school is in - has shielding pupils or wider restrictions - rather than directly associated with ordering as such - might need to be renamed to `covid_state` or something better?
> Also I think some of the copy you're using looks very similar to what I need so might be better off in the `en.yml`?

I've moved the `order_state` to `schools`, as discussed.

So as to not block further work, I'll do a follow-up PR to extract the copy bits to `en.yml`.

benilovj:
Oh, I should maybe rename the field as you suggested @tonyheadford

PR #462, opened 2020-09-14
[502] Handle computacenter error on cap update

asmega:
### Context

- https://trello.com/c/FBZnVdwk/502-handle-computacenter-returning-200-response-on-failed-capupdates
- For cap updates the computacenter response can return a 200 when there is an error

- To ensure there is no error we also parse the response body
- If there is an error it will be raised

### Guidance to review

- n/a

### DIFF
--- a/app/models/computacenter/outgoing_api/cap_update_request.rb
+++ b/app/models/computacenter/outgoing_api/cap_update_request.rb
@@ -1,3 +1,5 @@
+require 'nokogiri'
+
 class Computacenter::OutgoingAPI::CapUpdateRequest
   attr_accessor :endpoint, :username, :password, :allocation_ids, :timestamp
   attr_accessor :body, :payload_id, :response, :logger
@@ -27,7 +29,7 @@ class Computacenter::OutgoingAPI::CapUpdateRequest
   def handle_response!
     response_body = @response.body.to_s
     @logger.debug("Response from Computacenter: \n#{response_body}")
-    unless @response.status.success?
+    unless success?
       raise(
         Computacenter::OutgoingAPI::Error,
         "Computacenter responded with #{@response.status}, response_body: #{response_body}",
@@ -48,5 +50,21 @@ class Computacenter::OutgoingAPI::CapUpdateRequest
 
   def renderer
     Computacenter::OutgoingAPI::BaseController
+  end
+
+private
+
+  def success?
+    @response.status.success? && xml_success?
+  end
+
+  def xml_response
+    @xml_response ||= Nokogiri::XML(@response.body)
+  end
+
+  def xml_success?
+    return true if xml_response.css('HeaderResult').blank?
+
+    xml_response.css('HeaderResult').attr('status').value == 'Success'
   end
 end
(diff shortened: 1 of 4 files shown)


benilovj (approved):

spec/models/computacenter/outgoing_api/cap_update_request_spec.rb
```diff
@@ -87,5 +87,20 @@
         expect { request.post! }.to raise_error(Computacenter::OutgoingAPI::Error)
       end
     end
+
+    context 'when the response contains an error' do
+      let(:response_body) do
+        '<CapAdjustmentResponse dateTime="2020-08-21T12:30:40Z" payloadID="11111111-1111-1111-1111-111111111111"><HeaderResult errorDetails="Non of the records are processed" piMessageID="11111111111111111111111111111111" status="Failed"/><FaledRecords><Record capAmount="9" capType="DfE_RemainThresholdQty|Std_Device" errorDetails="New cap must be greater than or equal to used quantity" shipTO="11111111" status="Failed"/></FaledRecords></CapAdjustmentResponse>'
```
aldavidson, line 93: There's a typo here: `<FaledRecords` should be `<FailedRecords`

app/models/computacenter/outgoing_api/cap_update_request.rb
```diff
@@ -49,4 +51,20 @@ def construct_body
...
+  end
+
+  def xml_response
+    @xml_response ||= Nokogiri::XML(@response.body)
+  end
+
+  def xml_success?
+    return true if xml_response.css('HeaderResult').blank?
+
+    xml_response.css('HeaderResult').attr('status').value != 'Failed'
```
aldavidson, line 68: I'd prefer this to be a positive check on `value == 'Success'` rather than `!= 'Failed'`

aldavidson (changes requested):

spec/models/computacenter/outgoing_api/cap_update_request_spec.rb
asmega: This was copied so it's possible this is a genuine error from the api we're working with

app/models/computacenter/outgoing_api/cap_update_request.rb
asmega: I've seen the api docs so I've updated this now

spec/models/computacenter/outgoing_api/cap_update_request_spec.rb
aldavidson: I checked with them over email - it was a typo in a previous version of their API, which has since been fixed. 

```ruby
[18] pry(main)> puts response.body.to_s
<CapAdjustmentResponse dateTime="2020-09-14T21:55:37Z" payloadID="bcd794dc-e73e-4527-9619-2d8070b77c75"><HeaderResult piMessageID="059df5cdf6d511eab0cc00000035c636" status="Success"/><FailedRecords/></CapAdjustmentResponse>
```

asmega: Good work. I've fixed the typo. I've also added an extra test case with the success sample above to ensure that works as expected

aldavidson: awesome! 👍

aldavidson (approved):
Looks good - thanks! 💯

PR #467, opened 2020-09-15
[672] add first-draft UI for support users to enable orders

aldavidson:
### Context

Trello card [672](https://trello.com/c/jcmxnwR1/672-support-users-can-set-the-order-state-cannot-order-can-order-specific-circumstances), as part of the larger ticket [664](https://trello.com/c/xUIUD6em/664-allow-disallow-ordering-and-set-caps-via-support)
 
Add just the initial user interface to allow support users to enable ordering of devices for schools.
This first version applies the updates from the form and redirects back to the school page - a subsequent PR will introduce the "Check and confirm" page.

Screenshots:

Change link and success message:
Screen Shot 2020-09-15 at 12 12 16

Form and validation errors:
Screen Shot 2020-09-15 at 12 12 58

### DIFF
--- a/app/controllers/support/devices/order_status_controller.rb
+++ b/app/controllers/support/devices/order_status_controller.rb
@@ -1,0 +1,35 @@
+class Support::Devices::OrderStatusController < Support::BaseController
+  before_action :set_school
+
+  def edit
+    @form = Support::EnableOrdersForm.new(enable_orders_form_params)
+  end
+
+  def update
+    @form = Support::EnableOrdersForm.new(enable_orders_form_params)
+    if @form.valid?
+      ActiveRecord::Base.transaction do
+        allocation = SchoolDeviceAllocation.find_or_initialize_by(school_id: @school.id, device_type: 'std_device')
+        # we only take the cap from the user if they chose specific circumstances
+        # for both other states, we need to infer a new cap from the chosen state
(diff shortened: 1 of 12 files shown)


tonyheadford (approved):
LGTM!

app/controllers/support/devices/order_status_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+  def update
+    @form = Support::EnableOrdersForm.new(enable_orders_form_params)
+    if @form.valid?
+      ActiveRecord::Base.transaction do
+        allocation = SchoolDeviceAllocation.find_or_initialize_by(school_id: @school.id, device_type: 'std_device')
+        # we only take the cap from the user if they chose specific circumstances
+        # for both other states, we need to infer a new cap from the chosen state
+        allocation.cap = allocation.cap_implied_by_order_state(order_state: @form.order_state, given_cap: @form.cap)
+        allocation.save!
+        @school.update!(order_state: @form.order_state)
```
benilovj, line 17: It feels like the controller is doing too much here. Could we extract this logic into something else (service? model?) in subsequent PRs?

aldavidson: I agree, and I was thinking that at the time I submitted this, but I wanted to get the Computacenter API call in there too and then see if that introduced any further complications.

That's now done (in #468) and I think there's a clear chunk of code I can pull out to a separate ... _something_  - I'll do that in a separate PR

PR #468, opened 2020-09-15
[679] call Computacenter API when cap is updated

aldavidson:
### Context

[Trello card 679](https://trello.com/c/E3iH8vu5/679-when-the-cap-is-updated-call-the-computacenter-api-synchronously-and-show-an-error-if-the-api-doesnt-respond-successfully-see-ht), part of [larger card 664](https://trello.com/c/xUIUD6em/664-allow-disallow-ordering-and-set-caps-via-support)

At the point a cap is changed, call the Computacenter API synchronously, and rollback the change if there is an error


app/controllers/support/devices/order_status_controller.rb
```diff
@@ -15,6 +15,12 @@ def update
         allocation.cap = allocation.cap_implied_by_order_state(order_state: @form.order_state, given_cap: @form.cap)
         allocation.save!
         @school.update!(order_state: @form.order_state)
+
+        request = Computacenter::OutgoingAPI::CapUpdateRequest.new(allocation_ids: [allocation.id])
+        request.post!
```
benilovj, line 20: As mentioned on the other PR, should we extract this logic into a service or model?

aldavidson: Yes, totally

aldavidson: @benilovj done!

app/services/cap_update_service.rb
```diff
@@ -0,0 +1,34 @@
...
+  def update_order_state!(order_state)
+    @school.update!(order_state: order_state)
+  end
+
+  def update_cap!(cap)
+    allocation = SchoolDeviceAllocation.find_or_initialize_by(school_id: @school.id, device_type: 'std_device')
+    # we only take the cap from the user if they chose specific circumstances
+    # for both other states, we need to infer a new cap from the chosen state
+    allocation.cap = allocation.cap_implied_by_order_state(order_state: @school.order_state, given_cap: cap)
+    allocation.save!
```
benilovj, line 26: This line blows up if the allocation is invalid (this issue already existed before this PR but it feels like now is a good time to fix it).

This can happen since #469 introduced some validations that are more sophisticated than what `Support::EnableOrdersForm` catches (because the form object is very simplistic).

I can see a few approaches to fix this:

1. Make the validations on `SchoolDeviceAllocation` and `Support::EnableOrdersForm` consistent. It feels like this approach would introduce unnecessary duplication between those models.
2. Move the construction/fetch of the allocation to `Support::EnableOrdersForm`, and validate that the allocation is valid inside `EnableOrdersForm`. This approach enables copying the validation errors from the allocation onto `EnableOrdersForm`, making it easy to display the error messages on the page.
3. Call `allocation.valid?` before trying to `save!` and skip the call to the Computacenter API if the return is false. That way the controller can also check for validity of the allocation and display the errors if it's not.

app/services/cap_update_service.rb
```diff
@@ -0,0 +1,34 @@
...
+    allocation
+  end
+
+  def notify_computacenter!(allocation_id)
```
benilovj, line 30: Could we call this something else please? e.g. `update_caps_on_computacenter` or similar?

I initially confused this with sending a Notify email to Computacenter, which is a follow-up story that we will have to add very shortly to this service.

app/services/cap_update_service.rb
```diff
@@ -0,0 +1,34 @@
+class CapUpdateService
```
benilovj, line 1: From its name, it sounds like this class should be generalisable to comms caps but isn't currently, since it's currently assuming `device_type: 'std_device'` inside `update_cap!`. How about passing in `device_type` via the initializer?

app/controllers/support/devices/order_status_controller.rb
```diff
@@ -9,18 +9,16 @@ def update
     @form = Support::EnableOrdersForm.new(enable_orders_form_params)
     if @form.valid?
       ActiveRecord::Base.transaction do
```
benilovj, line 11: Should the transaction be sitting outside the service or inside?

spec/features/support/devices/enabling_orders_for_a_school_spec.rb
```diff
@@ -42,23 +42,54 @@
...
             click_on 'Continue'
+            expect(mock_request).to have_received(:post!)
           end
 
-          # This will be the next PR
-          it 'takes me to the Check your answers page', pending: true do
-            expect(page).to have_text 'Check your answers and confirm'
-            expect(page).to have_text 'Yes, for specific circumstances'
-            expect(page).to have_text 'Up to 2 from an allocation of 50'
+          context 'when the Computacenter CapUpdate API does not raise an error' do
```
benilovj, line 58:
```suggestion
          context 'when the Computacenter CapUpdate API processes the update successfully' do
```

app/services/cap_update_service.rb
aldavidson: Sure, good idea 👍

app/controllers/support/devices/order_status_controller.rb
aldavidson: That's a really good question, and I thought about it a lot while coding this - but I'm pretty sure it should be outside the service. 

The service doesn't have any knowledge of the context in which it's called, only the caller does - in this case the controller, but it could also be called from some other larger process like a rake task.  The right thing to do is to let the exception bubble out and the appropriate remedial action be taken by the invoking context. Related, on some previous apps I've done things like enforce transactions for all non-GET requests by adding an `around_filter` in ApplicationController

The only case for putting it inside the service I can see would be if the service could be called remotely (like a REST API) - but then that's effectively at controller level anyway.

app/services/cap_update_service.rb
aldavidson: Done in e004e03c

app/services/cap_update_service.rb
aldavidson: On standup/dev huddle just now, we agreed to deal with this in a separate PR

aldavidson: 1. feels like the best approach to me. Maybe we can refactor the validation to a separate validator class, and then call it from both places

2. I did think about _not_ introducing a service object, but instead moving the save functionality to the form object like they have in Apply. I decide against, on the grounds of generally preferring form objects to represent a form and having any complex back-end interactions handled at the service layer  (or model layer directly if they're not so complex)

3. `save!` will raise an error if the record is invalid, then the error will bubble out to the controller and the transaction will be rolled back. The question then is how best for the form to indicate the validation error to the user - but if we do 1., then that problem is solved.

The only way I can see that the allocation could be invalid is if the user chooses the specific circumstances, and then enters a cap that's greater than the allocation - that's a problem with the form data, and should be validated at the form level.

benilovj (approved):

app/services/cap_update_service.rb
benilovj: I'm fairly relaxed about how we solve this, if we go for 1 and use a validator that's probably going to work well. May be worth combining the several validation rules into one custom validator, since they're all very similar anyway.

PR #469, opened 2020-09-15
[664] Cap and Allocation validations

asmega:
### Context

- https://trello.com/c/xUIUD6em/664-allow-cap-changes-via-support
- Add backend validations to ensure caps, allocations and order_state are kept in valid states

- Validations ensure that devices_ordered <= cap <= allocation
- Changing order_state we check the cap and allocation values

### Guidance to review

- `SchoolDeviceAllocation.devices_ordered` must be lte `SchoolDeviceAllocation.cap`
- `SchoolDeviceAllocation.cap` must be lte `SchoolDeviceAllocation.allocation`
- Changing `School.order_state` to `can_order` can only happen if there is a non-zero cap
- Changing `School.order_state` to `can_order_for_specific_circumstances` can only happen if there is a non-zero allocation

### DIFF
--- a/app/models/school_device_allocation.rb
+++ b/app/models/school_device_allocation.rb
@@ -3,6 +3,9 @@ class SchoolDeviceAllocation < ApplicationRecord
   belongs_to  :created_by_user, class_name: 'User', optional: true
   belongs_to  :last_updated_by_user, class_name: 'User', optional: true
 
+  validate :validate_cap_gte_devices_ordered
+  validate :validate_cap_lte_allocation
+
   enum device_type: {
     'coms_device': 'coms_device',
     'std_device': 'std_device',
@@ -28,6 +31,20 @@ class SchoolDeviceAllocation < ApplicationRecord
       self.cap = allocation.to_i
     else # specific circumstances
       given_cap
+    end
+  end
+
+private
+
+  def validate_cap_lte_allocation
+    if cap > allocation
+      errors.add(:cap, :lte_allocation)
+    end
+  end
+
+  def validate_cap_gte_devices_ordered
+    if cap < devices_ordered
+      errors.add(:cap, :gte_devices_ordered)
     end
   end
 end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -369,6 +369,11 @@
               blank: Select whether the contract is pay monthly or pay as you go (PAYG)
             agrees_with_privacy_statement:
               inclusion: The privacy statement must be shared with the account holder
+        school_device_allocation:
+          attributes:
+            cap:
+              lte_allocation: can’t be greater than allocation
+              gte_devices_ordered: can’t be less than devices ordered
         user:
           attributes:
             full_name:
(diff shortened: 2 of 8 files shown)


app/models/school_device_allocation.rb
```diff
@@ -19,4 +22,18 @@ def self.by_computacenter_device_type(cc_device_type)
...
+
+  def validate_cap_lte_allocation
+    if cap > allocation
+      errors.add(:cap, "can't be greater than allocation")
+    end
+  end
+
+  def validate_cap_gte_devices_ordered
+    if cap < devices_ordered
+      errors.add(:cap, "can't be less than devices ordered")
```
benilovj, line 36: can we use [a custom key](https://stackoverflow.com/a/18185133) to push this error message into `en.yml`?

app/models/school_device_allocation.rb
```diff
@@ -19,4 +22,18 @@ def self.by_computacenter_device_type(cc_device_type)
   def computacenter_cap_type
     Computacenter::CapTypeConverter.to_computacenter_type(device_type)
   end
+
+private
+
+  def validate_cap_lte_allocation
+    if cap > allocation
+      errors.add(:cap, "can't be greater than allocation")
```
benilovj, line 30: can we use [a custom key](https://stackoverflow.com/a/18185133) to push this error message into `en.yml`?

app/models/school.rb
```diff
@@ -85,4 +88,18 @@ def invite_school_contact
...
+private
+
+  def validate_can_order_against_cap
+    if can_order? && device_allocations.all? { |allocation| allocation.cap.zero? }
+      errors.add(:order_state, 'cannot change as all caps are zero')
+    end
+  end
+
+  def validate_can_order_for_specific_circumstances_against_allocation
+    if can_order_for_specific_circumstances? && device_allocations.all? { |allocation| allocation.allocation.zero? }
```
asmega, line 101: I'm not sure if this line is correct. Can someone confirm?

app/models/school_device_allocation.rb
asmega: done

app/models/school_device_allocation.rb
asmega: done

app/models/school.rb
benilovj: Yes i think this is right.

app/models/school.rb
```diff
@@ -85,4 +88,18 @@ def invite_school_contact
       false
     end
   end
+
+private
+
+  def validate_can_order_against_cap
+    if can_order? && device_allocations.all? { |allocation| allocation.cap.zero? }
```
asmega, line 95: Also this line. Can someone confirm this is correct?

benilovj (approved):

benilovj:
@asmega I think I've figured out where we went wrong. We've got circular dependencies between the `School` and `SchoolDeviceAllocation` models – it should only work one way. The `SchoolDeviceAllocation` is dependent on the school, so that's where all the validations should sit IMHO.

benilovj:
On second thoughts, I think the consistency check between school order state and allocations and caps being non-zero is more trouble than it's worth. 

Since we're going down the service route on the other PR, where the order state and caps will be changed transactionally together, I think we can safely skip this validation. I've amended the commit to remove it.

benilovj (dismissed):

benilovj (approved):

benilovj:
I think the validations are OK, but the `OrderStatusController` needs changes. Since there's another PR #468 that's making changes on that code, I think I'll leave my comments there instead.

PR #470, opened 2020-09-15
[635] ordering devices centrally

tonyheadford:
### Context
[Trello card](https://trello.com/c/43443pu0/635-ordering-devices-centrally)
Add ordering devices for responsible body
See [https://increasing-access-history.herokuapp.com/ordering-devices-centrally/](https://increasing-access-history.herokuapp.com/ordering-devices-centrally/)

Add 'order devices' page for RBs in devices pilot
The order devices page is different depending on:
Any schools can order
Any centrally managed schools can order
Any schools that will order devices can order

### Guidance to review
There are a couple of changes to the design that have had to be worked in.  The designs assume that we are tracking allocation amounts i.e. 'You've ordered 1 of 4 devices' which isn't a thing currently so that text is missing.
@emmajhf has provided some copy for when there are schools that will order devices that can order for specific circumstances and/or for local restrictions.
The final page in the design has not been added as it also relies on the tracking of allocation amounts - I'm also not sure at which point you'd want to switch from that page back to the 'Cannot order devices yet' page, if at all, or whether once an order has been placed you would never see the 'Cannot order devices yet' page again.

### DIFF
--- a/app/models/responsible_body.rb
+++ b/app/models/responsible_body.rb
@@ -90,33 +90,18 @@ class ResponsibleBody < ApplicationRecord
   end
 
   def is_ordering_for_schools?
-    schools.that_are_centrally_managed.count.positive?
+    schools.that_are_centrally_managed.any?
   end
 
   def has_centrally_managed_schools_that_can_order_now?
-    schools.that_are_centrally_managed.that_can_order_now.count.positive?
+    schools.that_are_centrally_managed.that_can_order_now.any?
   end
 
   def has_schools_that_can_order_devices_now?
-    schools.that_will_order_devices.that_can_order_now.count.positive?
+    schools.that_will_order_devices.that_can_order_now.any?
   end
 
   def has_any_schools_that_can_order_now?
-    schools.that_can_order_now.count.positive?
-  end
-
-  def count_of_schools_we_order_for_link_text
-    count = schools.that_are_centrally_managed.count
-    "#{count} #{'school'.pluralize(count)} you will be placing orders for"
-  end
-
-  def unmanaged_schools_that_can_order_link_text
-    count = schools.that_will_order_devices.can_order.count
-    "#{count} #{'school'.pluralize(count)} #{'has'.pluralize(count)} local coronavirus restrictions"
-  end
-
-  def unmanaged_schools_that_can_order_for_specific_circumstances_text
-    count = schools.that_will_order_devices.can_order_for_specific_circumstances.count
-    "#{count} #{'school'.pluralize(count)} can order devices for specific circumstances because their #{'request'.pluralize(count)} #{'has'.pluralize(count)} been approved."
+    schools.that_can_order_now.any?
   end
 end
--- a/app/models/responsible_body.rb
+++ b/app/models/responsible_body.rb
@@ -94,19 +94,15 @@ class ResponsibleBody < ApplicationRecord
   end
 
   def has_centrally_managed_schools_that_can_order_now?
-    schools.that_are_centrally_managed.that_can_order_std_devices_now.count.positive?
-  end
-
-  def schools_that_can_order_devices_now
-    schools.that_order_devices.that_can_order_std_devices_now
+    schools.that_are_centrally_managed.that_can_order_now.count.positive?
   end
 
   def has_schools_that_can_order_devices_now?
-    schools_that_can_order_devices_now.count.positive?
+    schools.that_will_order_devices.that_can_order_now.count.positive?
   end
 
   def has_any_schools_that_can_order_now?
-    schools.that_can_order_std_devices_now.count.positive?
+    schools.that_can_order_now.count.positive?
   end
 
   def count_of_schools_we_order_for_link_text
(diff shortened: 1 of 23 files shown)


app/models/responsible_body.rb
```diff
@@ -88,4 +88,35 @@ def self.with_completed_preorder_info_count
       ",
     )
   end
+
+  def is_ordering_for_schools?
+    schools.that_are_centrally_managed.count.positive?
```
benilovj, line 93: Would
```suggestion
    schools.that_are_centrally_managed.any?
```
achieve the same thing?

app/models/school.rb
```diff
@@ -73,9 +85,13 @@ def next_school_in_responsible_body_when_sorted_by_name_ascending
...
 
-  def can_order_devices?(device_type = 'std_device')
+  def has_devices_available_to_order?(device_type = 'std_device')
+    available_devices_count(device_type).positive?
+  end
+
+  def available_devices_count(device_type = 'std_device')
     allocation = device_allocations.by_device_type(device_type).first
-    allocation&.cap.to_i > allocation&.devices_ordered.to_i
+    allocation&.cap.to_i - allocation&.devices_ordered.to_i
```
benilovj, line 94: I wonder whether this logic should be on the allocation and called from here...

app/models/school.rb
```diff
@@ -85,4 +101,21 @@ def invite_school_contact
       false
     end
   end
+
+  def what_to_order_text
```
benilovj, line 105: Again, not sure about putting this logic on the model – I'd be ok with this living in the view.

app/views/responsible_body/devices/orders/specific_circumstances.html.erb
```diff
@@ -0,0 +1,18 @@
+<%= render partial: 'heading' %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">
+      <span class="govuk-caption-xl"><%= t('page_titles.order_devices.title') %></span>
+      <%= t('page_titles.order_devices.order_for_specific_circumstances') %>
+    </h1>
+
+    <p class="govuk-body">Now you’ve emailed us to <%= govuk_link_to 'request devices for specific circumstances', '#' %>, you can go ahead with your order.</p>
```
benilovj, line 10: Missing link...

app/models/responsible_body.rb
```diff
@@ -88,4 +88,35 @@ def self.with_completed_preorder_info_count
...
+
+  def has_schools_that_can_order_devices_now?
+    schools.that_will_order_devices.that_can_order_now.count.positive?
+  end
+
+  def has_any_schools_that_can_order_now?
+    schools.that_can_order_now.count.positive?
+  end
+
+  def count_of_schools_we_order_for_link_text
```
benilovj, line 108: Not sure about this method being on the model. Actually keeping the logic in the view allows you to use `https://apidock.com/rails/ActionView/Helpers/TextHelper/pluralize` which makes things a bit more readable:
```
    <p class="govuk-body">None of the
      <%= govuk_link_to '', responsible_body_devices_schools_path do %>
        <%= pluralize(@responsible_body.schools.that_are_centrally_managed.count, 'school') %> you will be placing orders for
      <% end %>
      have local coronavirus restrictions.
    </p>
```

app/models/responsible_body.rb
tonyheadford: Good spot!

app/models/school.rb
tonyheadford: Yes, good call

app/models/responsible_body.rb
tonyheadford: Think that one should also be handling 'has/have' to in the pluralisation and once it's more that one item it gets a bit more messy in the view - I agree these should't be on the model, would moving it into the `view_helper` be a better idea?

benilovj: > would moving it into the view_helper be a better idea?

I'm fine with that.

app/models/school.rb
aldavidson: ^ Seconded 👍  Let's try to keep presentation code out of the models, and either in the views (if it's simple) or helpers/components (if it's more complex)

app/models/responsible_body.rb
tonyheadford: Actually most of these were reasonable to do inline with just one going in the view helper

app/views/responsible_body/devices/orders/specific_circumstances.html.erb
tonyheadford: Good spot

app/models/school.rb
tonyheadford: Moved this one to the view helper

benilovj (approved):
LGTM! There's quite a lot going on here so I might have missed stuff but I think we best get this on dev/staging and see it with real-ish data.

PR #474, opened 2020-09-16
Make the connectivity pilot RB report more useful

benilovj:
# Context

When the connectivity pilot RB report in Support was conceived, we only had a limited number of RBs with users. Now that the devices work is also in the service, that assumption is no longer the case, so we need to filter the connectivity RB list
by participation in the connectivity pilot.

Only show RBs in the connectivity pilot for the support > internet > responsible bodies page.


aldavidson (approved):
LGTM

PR #475, opened 2020-09-16
[683] Who to contact form updates existing contact

asmega:
### Context

- https://trello.com/c/w2vkKQQu/683-rbs-cant-update-a-headteacher-contact
- Users are trying to update contacts but the implementation tries to create a new record with duplicated data which fires a uniqueness constraint therefore raising an error

- On this form if another contact is added but uses an email of an existing contact it will update the existing contact

### Guidance to review

- I'm not sure how to actually navigate to the concerned page. This http://localhost:3000/responsible-body/devices/schools/100000/who-to-contact worked for me though
- Ensure a contact already exists
- Enter email address of existing contact
- All other inputs should be something else
- Existing contact should now be updated with the just entered data

### DIFF
--- a/app/controllers/responsible_body/devices/who_to_contact_controller.rb
+++ b/app/controllers/responsible_body/devices/who_to_contact_controller.rb
@@ -18,7 +18,7 @@ class ResponsibleBody::Devices::WhoToContactController < ResponsibleBody::Device
       render :new, status: :unprocessable_entity
     else
       chosen_contact = @form.chosen_contact
-      chosen_contact.save! unless chosen_contact.persisted?
+      chosen_contact.save!
       @school.preorder_information.update!(school_contact: chosen_contact)
       flash[:success] = I18n.t(:success, scope: %i[responsible_body devices schools who_to_contact create], email_address: chosen_contact.email_address)
       redirect_to responsible_body_devices_school_path(@school.urn)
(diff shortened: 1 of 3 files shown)


app/controllers/responsible_body/devices/who_to_contact_controller.rb
```diff
@@ -18,8 +18,8 @@ def create
       render :new, status: :unprocessable_entity
     else
       chosen_contact = @form.chosen_contact
-      chosen_contact.save! unless chosen_contact.persisted?
```
asmega, line 21: i'm not sure why this guard is here. is it to guard some callback or notification?

app/controllers/responsible_body/devices/who_to_contact_controller.rb
```diff
@@ -18,8 +18,8 @@ def create
...
-      @school.preorder_information.update!(school_contact: chosen_contact)
+      chosen_contact.save!
+      @school.preorder_information&.update!(school_contact: chosen_contact)
```
asmega, line 22: My test school didn't have a preorder. Should I have given it one? Otherwise this will fix potentially an issue

benilovj: Yeah, you won't be able to get to this screen without a `PreorderInformation`.

asmega: ok, I've added preorder info on the test and removed the guard

app/controllers/responsible_body/devices/who_to_contact_controller.rb
benilovj: I don't think there is a really good reason for it being here. I probably added to avoid saving the headteacher contact again but then if it's already saved, the operation is a noop...

I think it's fine to remove.

benilovj (approved):

PR #476, opened 2020-09-16
678 enable orders confirmation screen

aldavidson:
### Context

[Trello card 678](https://trello.com/c/tvNJV6Cd/678-support-users-have-a-confirmation-screen-which-lets-them-see-and-adjust-the-changes-that-will-be-made-do-after-the-first-interac), final part of [larger ticket 664](https://trello.com/c/xUIUD6em/664-allow-disallow-ordering-and-set-caps-via-support)

* Introduce a confirmation page before actually applying changes to the ordering state of a school
* Factor-out a separate Validator object from the `SchoolDeviceAllocation` validations introduced in #469 
* Make the `EnableOrdersForm` adjust the given cap according to order status before validating, as the given cap should only be applied if the user selects the 'specific circumstances' option
* Add specs for the confirmation journey


app/views/support/devices/order_status/confirm.html.erb
```diff
@@ -0,0 +1,36 @@
...
+      order_state: @form.order_state,
+      cap: @form.cap,
+      allocation: @allocation,
+      change_path: support_devices_school_enable_orders_path(@school.urn, pass_through_params)) %>
+
+    <h2 class="govuk-heading-m govuk-!-margin-top-8">What happens next</h2>
+    <p class="govuk-body">We will:</p>
+    <ul class="govuk-list govuk-list--bullet">
+      <li>email everyone at <%= @school.name %> to tell them they <%= t(@form.order_state, scope: %i[support order_status confirm order_states]) %></li>
+      <li>tell their <%= @school.responsible_body.humanized_type %> <%= @school.responsible_body.name %> that this school <%= t(@form.order_state, scope: %i[support order_status confirm order_states]) %></li>
```
benilovj, line 23: I don't think we can include this yet because these things aren't automated. So could we please:

- remove this copy
- cut some cards into the Ready column, to remind us to add in these notifications

app/controllers/support/devices/order_status_controller.rb
```diff
@@ -21,6 +27,12 @@ def update
     render :edit, status: :unprocessable_entity
   end
 
+  # GET /support/devices/schools/:urn/enable-orders/confirm
```
benilovj, line 30: What's the easiest way to prevent this from falling over locally because the CC API is not set up? Is it worth having a guard on the CC call so it's only enabled in the environments where it's configured?

benilovj (changes requested):

app/controllers/support/devices/order_status_controller.rb
tonyheadford: Would a FeatureFlag be easiest like the email notifications?

app/views/support/devices/order_status/confirm.html.erb
aldavidson: I've removed the copy in 96b3d4fd  and added card 691](https://trello.com/c/PPq5XlWb/691-add-automated-notifications-to-the-school-and-its-rb-when-the-cap-orderstate-are-changed-via-support)

app/controllers/support/devices/order_status_controller.rb
aldavidson: A FeatureFlag makes most sense to me - I've added it in fcb5451f

benilovj (approved):

PR #477, opened 2020-09-16
[665] Added mail notification of cap changes to computacenter

tonyheadford:
### Context
[Trello card](https://trello.com/c/SsxJkncU/665-an-email-is-sent-to-computacenter-when-a-cap-is-updated)
An email notification is sent to Computacenter when a cap changes.
This is enabled using the feature flag `:notify_computacenter_of_cap_changes`

Addition of mailer for Computacenter with emails for devices and comms cap changes
Addition of feature flag to control whether email notification are sent

### Guidance to review
`CapUpdateService#update!` will trigger an email if the feature flag is activated.  This also attempts to send a SOAP request to computacenter (which failed on my dev environment) so may not be a safe way to test this?

### DIFF
--- a/app/mailers/computacenter_mailer.rb
+++ b/app/mailers/computacenter_mailer.rb
@@ -1,0 +1,48 @@
+class ComputacenterMailer < ApplicationMailer
+  def notify_of_devices_cap_change
+    setup_params
+
+    template_mail(
+      devices_cap_change_template_id,
+      to: recipient,
+      personalisation: personalisation,
+    )
+  end
+
+  def notify_of_comms_cap_change
+    setup_params
+
+    template_mail(
+      comms_cap_change_template_id,
+      to: recipient,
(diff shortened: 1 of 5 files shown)


benilovj:
> This also attempts to send a SOAP request to computacenter (which failed on my dev environment) so may not be a safe way to test this?

I just had [the same issue](https://github.com/DFE-Digital/get-help-with-tech/pull/476#discussion_r489560994) when trying to test #476. I wonder whether we want to have a way to see if CC is enabled on the particular environment (maybe by checking whether `Settings.computacenter.outgoing_api.endpoint` is set, or similar), and then using that to decide whether to do outbound Computacenter-related stuff (e.g. API calls, this email)?

That would mean we could avoid having an extra the feature flag...  just thinking out loud...

benilovj (approved):

PR #482, opened 2020-09-17
[684] Support can set current school contact

asmega:
### Context

- https://trello.com/c/mQs5Q3Jl/684-no-ability-to-edit-school-contacts-for-support-users
- Allow support to be able to select a contact and set the as the current contact

- In support added a secondary button to be able to set the contact on the page as the current contact. Only visible if the contact is not the current contact

image

### Guidance to review

- Find school with multiple contacts
- Go go edit contact of the current contact, should not see the `Set as current contact` button as they are already the current contact
- Go to another contact
- Set them as the current contact
- They should now be the current contact, visible on school show page

### DIFF
--- a/app/controllers/support/devices/contacts_controller.rb
+++ b/app/controllers/support/devices/contacts_controller.rb
@@ -16,6 +16,15 @@ class Support::Devices::ContactsController < Support::BaseController
     end
   end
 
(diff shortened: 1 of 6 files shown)


aldavidson (approved):
Looks fine to me :shipit:

PR #483, opened 2020-09-17
Add school order devices page

tonyheadford:
### Context
[Trello card](https://trello.com/c/6SoAV292/562-schools-order-devices-in-lockdown)
Order devices page for schools

Add order devices page with link to TechSource

### Guidance to review
The trigger for displaying the order devices page is `School#can_order_devices?` which takes the `order_state` and whether there are any devices available to order into account - should this only consider the `order_state` so a school can see when they've used their allocation rather than see the 'You can't order devices yet' page once they've ordered their devices?

Currently kept the `You’ve ordered 0 of 2 devices` text although we don't track it and we decided not to show it on the RB order devices pages. We can only show "`devices_ordered` of `cap`" which may end up being confusing.
I extracted the text around the tech source link from the RB journey into a partial for re-use here - the wording is slightly different maybe a content review to make it consistent across the journeys is needed at some point.

image

### DIFF
--- a/app/controllers/school/devices_controller.rb
+++ b/app/controllers/school/devices_controller.rb
@@ -1,9 +1,5 @@
 class School::DevicesController < School::BaseController
   def order
-    # replace with if conditon below when we've developed the other
-    # branch of the user journey
-    # if false && @school.can_order_devices?
-    # else
     if @school.can_order_devices?
       render :order
     else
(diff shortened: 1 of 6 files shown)


app/controllers/school/devices_controller.rb
```diff
@@ -4,8 +4,11 @@ def order
     # branch of the user journey
     # if false && @school.can_order_devices?
     # else
```
asmega, line 6: i assume we can now delete this commented code as you've implemented it?

asmega (approved):

app/controllers/school/devices_controller.rb
tonyheadford: Good spot

PR #484, opened 2020-09-17
[682] RBs can update the school contact

asmega:
### Context

- https://trello.com/c/afX5ZJbs/682-no-ability-to-edit-school-contacts-rb-user
- Allow RBs to update the school contact

- Adds change link to edit contact on school show page
- If headteacher contact present radios will be visible, otherwise no radios will be visible
- It will preselect headteacher contact or someone else radio depending on the current contact on the preorder info record
- It will populate the someone else fields if an existing contact for the school exists
- Changing the radio will point the preorder info record to the correct contact

#### Add change link
image

#### Edit contact page
image

### Guidance to review

There are quite a few scenarios to play with

- A school with and without a headteacher contact
- A school with and without a preorder info record
- A school with and without a contact record set on the preorder record
- The preorder record set as the headteacher contact or another contact
- Updating the record between headteacher and other contact
- Updating the record on the second contact
- Updating with the someone else contact with the headteacher email address should update headteacher contact

### DIFF
--- a/app/components/responsible_body/school_details_summary_list_component.rb
+++ b/app/components/responsible_body/school_details_summary_list_component.rb
@@ -94,6 +94,8 @@ private
       [{
         key: 'School contact',
         value: contact_lines.map { |line| h(line) }.join('<br>').html_safe,
+        change_path: responsible_body_devices_school_who_to_contact_edit_path(school_urn: @school.urn),
+        action: 'Change',
       }]
     else
       []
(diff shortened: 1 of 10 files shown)


tonyheadford (commented):
Looks good other than when you change the headteacher contact (using their email in the other part  to update details) you then no longer get the option to select headteacher or have a different contact you can only edit the single contact - is that the intention?

asmega:
> Looks good other than when you change the headteacher contact (using their email in the other part to update details) you then no longer get the option to select headteacher or have a different contact you can only edit the single contact - is that the intention?

@tonyheadford good spot. I've fixed the issue and added an extra test

tonyheadford (approved):
Yep works great now

PR #485, opened 2020-09-17
[667] add a first-draft UI to allow support users to adjust allocations

aldavidson:
### Context

[Trello card 667](https://trello.com/c/IdY4UmgF/667-support-users-can-set-allocations-for-devices)

Add first-draft UI for adjusting device allocations. There was no design or prototype to work off, so all design suggestions very welcome.

Screens:

Screen Shot 2020-09-17 at 17 19 33

Screen Shot 2020-09-17 at 17 19 53

### DIFF
--- a/app/components/support/school_details_summary_list_component.rb
+++ b/app/components/support/school_details_summary_list_component.rb
@@ -6,7 +6,7 @@ private
   end
 
   def allocation_row
-    super.except(:action_path, :action)
+    super.except(:action_path, :action).merge(change_path: support_devices_school_allocation_edit_path(school_urn: @school.urn))
   end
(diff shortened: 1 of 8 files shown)


emmajhf:
@aldavidson, I think this looks good.

Tiny thing: can we say 'Change device allocation' rather than 'Adjust'. This is consistent with the link text.

Also, you might want to add a suggested action to the error message, eg 'Enter an allocation that's greater than the current cap'. (Not sure if it can be the same. If it is, we'd say: 'Enter an allocation that's greater than, or the same as, the current cap'.)

benilovj (approved):

PR #492, opened 2020-09-18
[693 / 694] RB users and orders_devices

asmega:
### Context

- https://trello.com/c/9L3IQkh4/694-adding-or-editing-rb-users-should-force-setting-whether-the-user-needs-a-techsource-account-or-not
- https://trello.com/c/upShx5aS/693-all-existing-rb-users-should-be-eligible-for-a-techsource-account

- New RB users have `orders_devices` set to `true`
- Migration to set all existing RB users `orders_devices` to true
- Migration will fire activerecord lifecycle callbacks, ie. `PaperTrail::Version` records will be created and potentially `Computacenter::UserChange` records will be created

### Guidance to review

- Create new RB user
- New user `orders_devices` should be `true`
- All existing RB users `orders_devices` should be `true`

### DIFF
--- a/db/migrate/20200918101304_data_rb_users_techsource_account.rb
+++ b/db/migrate/20200918101304_data_rb_users_techsource_account.rb
@@ -1,0 +1,9 @@
+class DataRbUsersTechsourceAccount < ActiveRecord::Migration[6.0]
+  def up
+    User.responsible_body_users.update(orders_devices: true)
+  end
+
+  def down
+    # this migration cannot be reversed
+  end
+end
--- a/spec/controllers/responsible_body/users_controller_spec.rb
+++ b/spec/controllers/responsible_body/users_controller_spec.rb
@@ -47,4 +47,32 @@ RSpec.describe ResponsibleBody::UsersController do
       end
     end
   end
+
+  describe '#create' do
+    before do
+      sign_in_as rb_user
+    end
+
+    def perform_create!
+      post :create, params: {
+        id: local_authority.id,
+        user: {
+          full_name: 'John Doe',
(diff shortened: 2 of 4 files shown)


db/migrate/20200918101304_data_rb_users_techsource_account.rb
```diff
@@ -0,0 +1,9 @@
+class DataRbUsersTechsourceAccount < ActiveRecord::Migration[6.0]
+  def up
+    User.responsible_body_users.update(orders_devices: true)
```
benilovj, line 3: this will trigger the code that generates `UserChange` entries, right?

spec/controllers/responsible_body/users_controller_spec.rb
```diff
@@ -47,4 +47,33 @@
       end
     end
   end
+
+  describe '#create' do
+    before do
+      FeatureFlag.activate(:rbs_can_manage_users)
```
benilovj, line 53: I don't think you should need this flag for creating users?

benilovj (approved):

db/migrate/20200918101304_data_rb_users_techsource_account.rb
asmega: it will. is that what we want?

benilovj: yes, definitely.

spec/controllers/responsible_body/users_controller_spec.rb
asmega: good spot, i've now removed this line

PR #499, opened 2020-09-18
[697] New RB users assigned to a school

asmega:
### Context

- https://trello.com/c/hlBiVAGh/697-responsible-bodies-with-only-one-school-should-see-the-school-journey
- On support adding user to RB if RB only has one school and this is the first user set them up as a hybrid user 

- Hybrid users belongs to both an RB and a school
- Hybrid setup also orchestrates a few things to put things in the correct state 

#### User experience

- Login as support user
- Invite a new RB user - there must be no other RB users + RB must only have one school
- Logout and login as invited user
- Invited user should go thru school journey and no touch any RB screen
- Wizard should work and after sent user to school
- Logout and log back in
- Should be sent to school screens

#### Backend checks

- Preorder should be made for the school
- Preorder set `who_will_order_devices` to `school`
- Preorder should set `status` to `school_contacted`
- The school contacts should have one contact which is this hybrid user
- The preorder contact should be this school contact

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -120,7 +120,11 @@ class User < ApplicationRecord
     school = responsible_body.schools.first
 
     update(school: school)
-    school.create_preorder_information!(who_will_order_devices: 'school')
+    contact = school.contacts.create!(email_address: email_address,
+                                      full_name: full_name,
+                                      role: :contact,
+                                      phone_number: telephone)
+    school.create_preorder_information!(who_will_order_devices: 'school', school_contact: contact)
   end
 
 private
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -107,6 +107,10 @@ class User < ApplicationRecord
     responsible_body || school&.responsible_body
   end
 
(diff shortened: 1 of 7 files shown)


app/models/user.rb
```diff
@@ -107,6 +107,28 @@ def effective_responsible_body
     responsible_body || school&.responsible_body
   end
 
+  def hybrid?
+    school_id && responsible_body_id
+  end
+
+  def hybrid_setup!
```
asmega, line 114: no idea where to put this so I've shoved it here. probably not the right place...

tonyheadford: I would be tempted to push this out into a service.

tonyheadford (approved):
LGTM - I _think_ I understand the requirements. I do think the hybrid stuff should probably be refactored into a service.

app/models/user.rb
asmega: Ok. I think we can do that next time round. I've left everything in one place so it should be trivial to extract out

PR #500, opened 2020-09-20
Invite school contacts as soon as they're picked by the RB

benilovj:
### Context

While we were gaining confidence in the school journey, we didn't send out the school invites automatically after responsible bodies picked school contacts. Instead, they were waiting to be transitioned as a dev task/through support. We now have enough confidence in the school journey to start sending these out automatically.

There is still an edge case when inviting the school contact fails: when the school contact is an existing user in the system (because we can't handle multi-org yet). Those schools will remain in the 'school will be contacted' state.

- Invite the school contact as soon as they're picked by the RB
- Leave schools that can't be invited in the 'school will be contacted' state
- Make it really explicit that in `WhoToContactController`, the `create` and `edit` actions do the same thing.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -436,7 +436,8 @@
       schools:
         who_to_contact:
           create:
-            success: "Saved. We will email %{email_address} shortly"
+            failure: "Saved. We will email %{email_address} shortly"
+            success: "Saved. We’ve emailed %{email_address}"
     users:
       create:
         success: "We’ve sent an invitation to %{email_address}"
--- a/app/controllers/responsible_body/devices/who_to_contact_controller.rb
+++ b/app/controllers/responsible_body/devices/who_to_contact_controller.rb
@@ -6,24 +6,7 @@ class ResponsibleBody::Devices::WhoToContactController < ResponsibleBody::Device
   end
 
   def create
-    @form = ResponsibleBody::Devices::WhoToContactForm.new({
-      school: @school,
-    }.merge(who_to_contact_form_params))
-
-    if @form.invalid?
-      render :new, status: :unprocessable_entity
-    else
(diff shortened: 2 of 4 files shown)


config/locales/en.yml
```diff
@@ -436,7 +436,8 @@
       schools:
         who_to_contact:
           create:
-            success: "Saved. We will email %{email_address} shortly"
+            failure: "Saved. We will email %{email_address} shortly"
```
fofr, line 439: What are the failure cases here?

benilovj: It's this case:

https://github.com/DFE-Digital/get-help-with-tech/pull/500/commits/1d28c9a16cc75f83429ff1257c04c96d7b1f2a96

when the contact about to be invited is already a user.

aldavidson: I don't think I'm really clear about the intended behaviour here - if I've followed this correctly, then when the school invites another person who is _not_ already on the system, it should:

1. Add a User record for them 
2. Send them an invite email
3. Update the preorder information status to "School contacted"

But if the person _is_ already on the system, it will:

1. Try-but-fail to add a user record
2. Not send an invite email
3. Update the preorder_information status to "School will be contacted shortly"

So we're then left with a situation requiring manual intervention from us. Do we know yet how are we going to detect these cases, and how we're going to resolve them?

benilovj: > So we're then left with a situation requiring manual intervention from us.

yep.

> Do we know yet how are we going to detect these cases, 

They will be the ones remaining in "school will be contacted" once every other school is on.

> and how we're going to resolve them?

I think we can onboard them once we support multi-org.

aldavidson (approved):

PR #501, opened 2020-09-21
698 refactor user creation to a service

aldavidson:
### Context

See [Trello card 698](https://trello.com/c/Yn0xKhWJ/698-refactor-user-creation-to-a-service)

* Add a separate `CreateUser` service, with methods `invite_responsible_body_user` and `invite_school_user`

### Guidance to review

All behaviour should be unchanged

### DIFF
--- a/app/controllers/responsible_body/users_controller.rb
+++ b/app/controllers/responsible_body/users_controller.rb
@@ -12,14 +12,12 @@ class ResponsibleBody::UsersController < ResponsibleBody::BaseController
   end
 
   def create
-    @rb_user = @responsible_body.users.new(
-      user_params.merge(responsible_body_id: @responsible_body.id,
-                        orders_devices: true),
+    @rb_user = CreateUserService.invite_responsible_body_user(
+      user_params.merge(responsible_body_id: @responsible_body.id),
     )
-
-    if @rb_user.valid?
-      @rb_user.save!
-      InviteResponsibleBodyUserMailer.with(user: @rb_user).invite_user_email.deliver_later
+    # If anything goes wrong, the service will return a non-persisted user
+    # object so that we can inspect the errors
+    if @rb_user.persisted?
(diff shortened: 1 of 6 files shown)


app/controllers/responsible_body/users_controller.rb
```diff
@@ -12,14 +12,12 @@ def new
...
     )
-
-    if @rb_user.valid?
-      @rb_user.save!
-      InviteResponsibleBodyUserMailer.with(user: @rb_user).invite_user_email.deliver_later
+    # If anything goes wrong, the service will return a non-persisted user
+    # object so that we can inspect the errors
+    if @rb_user.persisted?
       flash[:success] = I18n.t(:success, scope: %i[responsible_body users create], email_address: @rb_user.email_address)
       EventNotificationsService.broadcast(InviteEvent.new(user: @user))
```
benilovj, line 22: I'd probably vote for bringing the slack event into the service as well – in my eyes, it's in the same category as the email notification to the user.

aldavidson: I originally thought that too, but the InviteEvent is to show user activity - that a user from that RB has invited someone else - and if we put it into the service, then it will be broadcast from the support UI too. Which would be misleading.

aldavidson: Yep, the InviteEvent message says -
`"A user from %{organisation} has invited someone to the service"`
- which is not the case when it's done from the support UI. 
So are we OK to leave this in the controller  @benilovj  ?

benilovj: Yep in the interest of expediency, let's leave it as you've got it. Ideally we'd have the concept of **who** invited the user and use appropriate messaging in slack, but that can be introduced later.

benilovj (approved):

PR #466, opened 2020-09-15
Add govuk components

aliuk2012:
### Changes proposed in this pull request
- Adds [govuk-component gem](https://github.com/DFE-Digital/govuk-components) - [(Demo)](https://dfe-digital.github.io/govuk-components/)
- Replaces tag component with govuk-component version
- Replaces summary list component with govuk-component version
- Replaces breadcrumbs component with govuk-component version
- Replaces govuk_link_to helper with govuk-component version

There is plenty more that could be replaced but I'll leave that for another PR, otherwise this is going to be an even larger PR to review.

### Guidance to review

Attempt to view different pages in the review app https://ghwt-review-pr-466.herokuapp.com/

PR #504, opened 2020-09-21
Add govuk components and replace one component

aliuk2012:
### Context
Part of #466 

- Adds [govuk-component gem](https://github.com/DFE-Digital/govuk-components) - [(Demo)](https://dfe-digital.github.io/govuk-components/)
- Replaces tag component with govuk-component version

There is plenty more that could be replaced but I'll leave that for another PR, otherwise this is going to be an even larger PR to review.


benilovj (approved):
Nice!

PR #509, opened 2020-09-21
Replaces custom breadcrumb component with govuk-component

aliuk2012:
### Context

Apologies for the size of this PR, as you will see from the file changes, this will standardise our approach to breadcrumbs. Currently, there is a mixture of hardcoded markup in the view or a view helper method which uses a custom view component under the hood.

```
 <% breadcrumbs([{ "Home" => school_home_path }, title ]) %>
```

part of #466 

- Replaces custom breadcrumb component with govuk-component version. All that's needed is to pass the component a hash with key name being the link copy and the value being a path for that breadcrumb link. if the value is null then it will just output the key name but not be considered a link.


app/views/responsible_body/devices/home/show.html.erb
```diff
@@ -1,10 +1,10 @@
 <%- title = t('page_titles.responsible_body_devices_home') %>
 <%- content_for :title, title %>
 <%- content_for :before_content do %>
-  <% breadcrumbs([
-    { 'Home' => responsible_body_home_path },
-    title,
-  ]) %>
+  <%= render GovukComponent::Breadcrumbs.new(breadcrumbs:{
```
benilovj, line 4: do we need to get rid of the syntactic sugar? (`breadcrumbs` is much punchier than `render GovukComponent::Breadcrumbs.new(breadcrumbs:`), and the `"#{title}": ''` could also be cleaner?

aliuk2012: @benilovj I've taking another stab at replacing the breadcrumbs. I've kept the view helper method as we discussed. I think this is good for another review and potential merge.

benilovj (approved):

PR #512, opened 2020-09-21
generate UserChange of type "Remove" when user.orders_devices is updated from true to false

aldavidson:
### Context

[Trello card 708](https://trello.com/c/5w0iHfAu/708-missing-cc-user-change-record-for-users-who-dont-order-anymore) 

When the `orders_devices` field is updated from true to false, detect this and make the corresponding `UserChange` record into a 'Remove' type


benilovj (approved):
Looks good to me.

asmega (approved):

PR #513, opened 2020-09-21
Swap out govuk_link_to with govuk-component version

aliuk2012:
- replaces custom govuk_link_to with govuk-component govuk_link_to helper method ([source](https://github.com/DFE-Digital/govuk-components/blob/master/app/helpers/govuk_link_helper.rb#L2-L4))

There is only one major difference between govuk-component helper method implementation and the one in the service. The difference is that govuk-component helper method does not support passing in a block. I don't see this as a blocker as I could only see one instance where a block was used and I assume the reason it was a block was because it was using html markup and rails would have been escaping it. This could have been overcome by using html_safe.

### DIFF
--- a/app/views/responsible_body/internet/mobile/manual_requests/confirm.html.erb
+++ b/app/views/responsible_body/internet/mobile/manual_requests/confirm.html.erb
@@ -14,9 +14,9 @@
           <%= @extra_mobile_data_request.account_holder_name %>
         </dd>
         <dd class="govuk-summary-list__actions">
-          <%= govuk_link_to('', new_responsible_body_internet_mobile_manual_request_path( anchor: 'extra-mobile-data-request-account-holder-name-field') ) do %>
-            Change<span class="govuk-visually-hidden"> Account holder name</span>
-          <%- end %>
+          <%= govuk_link_to('Change<span class="govuk-visually-hidden"> Account holder name</span>'.html_safe,
+                            new_responsible_body_internet_mobile_manual_request_path(
+                              anchor: 'extra-mobile-data-request-account-holder-name-field') ) %>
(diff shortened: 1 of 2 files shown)


app/views/responsible_body/internet/mobile/manual_requests/confirm.html.erb
```diff
@@ -14,9 +14,9 @@
           <%= @extra_mobile_data_request.account_holder_name %>
         </dd>
         <dd class="govuk-summary-list__actions">
-          <%= govuk_link_to('', new_responsible_body_internet_mobile_manual_request_path( anchor: 'extra-mobile-data-request-account-holder-name-field') ) do %>
-            Change<span class="govuk-visually-hidden"> Account holder name</span>
-          <%- end %>
+          <%= govuk_link_to('Change<span class="govuk-visually-hidden"> Account holder name</span>'.html_safe,
```
fofr, line 17: (We should push the block logic upstream to govuk-components, so that `govuk_link_to` continues to act like the native `link_to` does.)

aliuk2012: Good point, I'll take that up with govuk-components but I don't see that being a blocker for us to get this merged.

fofr:
For reference, switching to this implementation:
https://github.com/DFE-Digital/govuk-components/blob/master/app/helpers/govuk_link_helper.rb#L2-L4

fofr (approved):
Looks good.

benilovj:
I've raised the issue of `govuk_link_to` not taking blocks here: https://github.com/DFE-Digital/govuk-components/issues/65

PR #514, opened 2020-09-21
[647] school users can edit their users

tonyheadford:
### Context
[Trello card](https://trello.com/c/nW8wIPl7/647-school-users-can-edit-their-users)
Enable school users to update their users

Add edit and update functionality to the users controller for schools
Add 'Change' links to school user list (they are all the same  - should there only be one per-user?)
Added 'Orders devices' to the list as this seemed like it should be present?

### Guidance to review
As a school user you can update the details for the users associated with the school.

image

image


aldavidson (approved):
LGTM

PR #517, opened 2020-09-21
Single school journey fixes and improvements

benilovj:
### Context

The data for single-school responsible bodies should be consistent with the other RBs.

- set 'who will order devices' for single-school RBs
- don't explicitly override the PreorderInformation status, let it be derived automatically

### Guidance to review

Review each commit separately.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -23,9 +23,9 @@ RSpec.describe User, type: :model do
       expect(contact.phone_number).to eql(user.telephone)
     end
 
-    it 'marks preorder#status as school_contacted' do
+    it 'marks preorder#status as school_will_be_contacted' do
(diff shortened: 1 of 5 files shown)


spec/models/user_spec.rb
```diff
@@ -25,7 +25,7 @@
 
     it 'marks preorder#status as school_contacted' do
```
aldavidson, line 26: if the expected value has changed, can you update the example description to match?

aldavidson (changes requested):

spec/models/user_spec.rb
benilovj: Yes, good catch, fixed and force pushed.

aldavidson (approved):

PR #518, opened 2020-09-21
Fix broken links in support

benilovj:
### Context

Because the school details table in support inherits from the responsible body school details table, a few broken links have snuck in.

Remove the broken links that point to the RB interface.


aldavidson (approved):
LGTM 👍

PR #522, opened 2020-09-22
Single school RB onboarding: iteration

benilovj:
### Context

#499 was a first pass at allowing single-school responsible bodies onto the service. When trying to run it on production, I realised that the assumptions in that change we'd made were incorrect, namely many single-school RBs already have existing users from wave 1.

Iterate logic for onboarding single-school responsible bodies:

- Handle the case where an RB already has users
- Handle the case where the RB has no users but the school has a headteacher
- Remove the v1 logic as it's superseded

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -115,25 +115,6 @@ class User < ApplicationRecord
     school_id && responsible_body_id
   end
 
-  def hybrid_setup!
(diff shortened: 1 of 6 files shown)


tonyheadford (approved):
Looks great!

asmega (approved):

PR #529, opened 2020-09-22
Use dashes instead of underscores in support URL

benilovj:
### Context

OCD / consistency.

Use dashes instead of underscores in support URL.


tonyheadford (approved):
Good call

PR #530, opened 2020-09-22
Onboard single-school RBs with users who are an existing school contacts

benilovj:
### Context

Having onboarded a handful of single academy trusts, the onboarding service had trouble with a single academy trust where one of the trust users was the headteacher. When the service tried to create the headteacher as the school contact, it triggered the email uniqueness validation.

Only create a new school contact if one matching the user doesn't already exist.


tonyheadford (approved):
LGTM

PR #533, opened 2020-09-23
720 record cap update request timestamp and payload

aldavidson:
### Context

See [Trello card 720](https://trello.com/c/Fd0q7j4m/720-investigate-missing-cap-updates) - this will help us diagnose any future Cap Update issues

* When a successful CapUpdateRequest is made, record the timestamp & payload ID on each of its allocations

### DIFF
--- a/app/models/computacenter/outgoing_api/cap_update_request.rb
+++ b/app/models/computacenter/outgoing_api/cap_update_request.rb
@@ -36,7 +36,6 @@ class Computacenter::OutgoingAPI::CapUpdateRequest
       )
     end
 
-    SchoolDeviceAllocation.where(id: @allocation_ids).update_all(cap_update_request_timestamp: @timestamp, cap_update_request_payload_id: @payload_id)
     @response
   end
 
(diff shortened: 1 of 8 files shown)


app/models/computacenter/outgoing_api/cap_update_request.rb
```diff
@@ -36,6 +36,7 @@ def handle_response!
       )
     end
 
+    SchoolDeviceAllocation.where(id: @allocation_ids).update_all(cap_update_request_timestamp: @timestamp, cap_update_request_payload_id: @payload_id)
```
benilovj, line 39: It feels like we're breaking encapsulation a bit by putting this update here, rather than at the point where `post!` is called... In my mind, this class should concern itself with CC request/response logic but shouldn't really know too much about allocations/caps etc.

benilovj (approved):
Functionally fine, just one comment about where the update should live.

app/models/computacenter/outgoing_api/cap_update_request.rb
aldavidson: yep, fair point - done in cadfa4e

PR #539, opened 2020-09-23
[721] CC can bulk update users with TechSource accounts

asmega:
### Context

- https://trello.com/c/1ZdxZJaL/721-form-for-allowing-cc-to-confirm-the-creation-of-techsource-accounts

- Adds pages to Computacenter area so they can bulk update users who have TechSource accounts
- Add migration to add column on user to track if user has a TechSource accounts, defaults to false
- Pauls changes are also included in this PR https://github.com/DFE-Digital/get-help-with-tech/pull/541 which update the computacenter home page

image
image
image
image

### Guidance to review

- Login as a computacenter user
- Should see section at bottom to bulk update users with TechSource accounts
- On form entering blank or not one email per row raise validation error
- Submitting the form should update found users to have a TechSource account

### DIFF
--- a/app/form_objects/bulk_techsource_form.rb
+++ b/app/form_objects/bulk_techsource_form.rb
@@ -1,0 +1,27 @@
+class BulkTechsourceForm
+  include ActiveModel::Model
+
+  attr_accessor :emails
+
+  validates :emails, presence: true
+  validate :validate_one_email_per_line
+  validate :validate_valid_emails
+
+  def array_of_emails
+    emails.split("\r\n").map(&:strip).reject(&:blank?).map(&:downcase)
+  end
+
+private
+
+  def validate_one_email_per_line
+    unless array_of_emails.all? { |line| !line.match(/,| /) }
+      errors.add(:emails, 'Enter no more than one email address per line')
+    end
+  end
+
+  def validate_valid_emails
+    unless array_of_emails.all? { |line| line.match(/.*@.*\..*/) }
+      errors.add(:emails, 'Ensure all email addresses are valid')
+    end
+  end
+end
--- a/app/views/computacenter/home/show.html.erb
+++ b/app/views/computacenter/home/show.html.erb
@@ -4,30 +4,22 @@
       <%= t('page_titles.computacenter.home') %>
     </h1>
 
-    <p class="govuk-body">Use this service to:</p>
+    <h2 class="govuk-heading-m">
+      <%= govuk_link_to 'Download TechSource users', computacenter_user_ledger_path(format: :csv) %>
+    </h2>
 
-    <ul class="govuk-list govuk-list--bullet">
-      <li>
-        <span class="govuk-!-margin-bottom-4">download a CSV file of users who have permission to place orders on TechSource</span>
-        <br /><br />
-        <%= govuk_button_link_to 'Download TechSource users', computacenter_user_ledger_path(format: :csv), classes: 'govuk-!-margin-top-4' %>
-      </li>
-    </ul>
+    <p class="govuk-body">Download a CSV file of all users with permission to place orders on TechSource.</p>
 
-    <ul class="govuk-list govuk-list--bullet">
-      <li>
-        <span class="govuk-!-margin-bottom-4">download a CSV file of Google Chromebook details for schools</span>
-        <br /><br />
-        <%= govuk_button_link_to 'Download Chromebook details', computacenter_chromebooks_path(format: :csv), classes: 'govuk-!-margin-top-4' %>
-      </li>
-    </ul>
+    <h2 class="govuk-heading-m">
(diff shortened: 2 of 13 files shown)


db/migrate/20200923115939_add_techsource_account_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddTechsourceAccountToUsers < ActiveRecord::Migration[6.0]
+  def change
+    add_column :users, :has_techsource_account, :boolean, default: false
```
asmega, line 3: another option is use a time and set this to default to `NULL`

app/views/computacenter/home/show.html.erb
```diff
@@ -13,12 +13,21 @@
...
       </li>
     </ul>
+
+    <ul class="govuk-list govuk-list--bullet">
+      <li>
+        <span class="govuk-!-margin-bottom-4">update users that have TechSource accounts</span>
+        <br /><br />
+        <%= govuk_button_link_to 'Update TechSource users', computacenter_techsource_path, classes: 'govuk-!-margin-top-4' %>
+      </li>
+    </ul>
```
fofr, line 31: This list with buttons layout is weird, I'll do a design pass on this once this PR is merged to make it a little more friendly.

fofr:
This error message is a little odd when what's entered is invalid:
Screen Shot 2020-09-24 at 09 24 55

app/views/computacenter/techsource/new.html.erb
```diff
@@ -0,0 +1,21 @@
...
+    <h1 class="govuk-heading-xl">
+      <%= t('page_titles.computacenter.techsource') %>
+    </h1>
+
+    <p class="govuk-body">Enter email addresses of users that have a TechSource account, there should be one email address per line.</p>
+
+    <%= form_with model: @form, url: computacenter_techsource_path do |f| %>
+      <%= f.govuk_text_area :emails,
+                            label: { text: 'Email addresses' },
+                            hint_text: 'Enter one email address per line',
```
aldavidson, line 15: Following on from @fofr's comment, I think we can probably do without the hint text - it's just repeating what's in the paragraph above, and potentially in the error message below

db/migrate/20200923115939_add_techsource_account_to_users.rb
aldavidson: I think a boolean here is OK - it's consistent with the `is_support` and `is_computacenter` flags.

app/views/computacenter/home/show.html.erb
asmega: Your suggested changes have now been incorporated into this branch

app/services/bulk_techsource_service.rb
```diff
@@ -0,0 +1,34 @@
+class BulkTechsourceService
```
aldavidson, line 1: I think the code in this service is fine, but the name of the service and/or main method could be clearer - I had to look at the code to know what `BulkTechSourceService.call` does. Could it be something more descriptive, like `BulkTechSourceService.set_has_account_flag` ?

aldavidson (approved):
Just one comment about the name of the service/method - not a blocker though

app/form_objects/bulk_techsource_form.rb
```diff
@@ -0,0 +1,20 @@
+class BulkTechsourceForm
+  include ActiveModel::Model
+
+  attr_accessor :emails
+
+  validates :emails, presence: true
+  validate :validate_one_email_per_line
+
+  def array_of_emails
+    emails.split("\r\n").map(&:strip).reject(&:blank?)
```
benilovj, line 10:
```suggestion
    emails.split("\r\n").map(&:strip).reject(&:blank?).map(&:downcase)
```

app/services/bulk_techsource_service.rb
benilovj: Yeah, seconded. I'd prefer the name to be incredibly specific to what it's doing (e.g. `ConfirmTechsourceAccountCreatedService`)

benilovj (changes requested):
A couple of comments, I think the downcase is important otherwise I'd expect us to see case mismatches.

app/views/computacenter/techsource/new.html.erb
asmega: the hint text has now been removed. some surrounding copy has also been tweaked

app/services/bulk_techsource_service.rb
asmega: it's been renamed to the suggested one

app/form_objects/bulk_techsource_form.rb
asmega: done

benilovj (approved):

PR #543, opened 2020-09-24
Update order pages to match design, tweak TechSource guidance

fofr:
### Context

- Now the invitation email is not being sent, we need to be explicit about how to sign in to TechSource the very first time
- Update 'Cannot order yet' page to match design
- Use "Home" as first breadcrumb title

### DIFF
--- a/app/views/school/devices/cannot_order.html.erb
+++ b/app/views/school/devices/cannot_order.html.erb
@@ -2,7 +2,7 @@
 <%- content_for :title, title %>
 <%- content_for :before_content do %>
   <% breadcrumbs([
-    { t('page_titles.school_home') => school_home_path },
+    { 'Home' => school_home_path },
     'Order devices',
   ]) %>
 <%- end %>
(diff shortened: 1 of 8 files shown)


app/views/school/devices/cannot_order.html.erb
```diff
@@ -2,17 +2,22 @@
 <%- content_for :title, title %>
 <%- content_for :before_content do %>
   <% breadcrumbs([
-    { t('page_titles.school_home') => school_home_path },
-    title,
+    { 'Home' => school_home_path },
+    'Order devices',
```
benilovj, line 6: Can't we just change line 1 to:

```
<%- title = t('page_titles.school_order_devices') %>
```

and then leave this line and the h1 as is?

benilovj (approved):
Just one minor suggestion

app/views/school/devices/cannot_order.html.erb
fofr: No. The page title showing in `<title>` needs to remain the same. The "Order devices" in caption and breadcrumb is for orientation.

PR #549, opened 2020-09-25
718 refactor & simplify user_change generation

aldavidson:
### Context

A necessary precursor of [Trello card 718  - Handle multiple schools in an MVP way](https://trello.com/c/aYQOz3QB/718-spike-handle-multiple-schools-in-a-minimum-viable-way). This will enable us to trigger `Computacenter::UserChange` generation from a future `UserSchool` association. Also simplifies implementation of [700 - Send UserChanges to Computacenter's ServiceNow API](https://trello.com/c/mYNXeNHI/700-send-user-changes-to-computacenters-servicenow-api).

* Generate `Computacenter::UserChange` records directly, rather than from Papertrail versions
* Determine whether a UserChange is needed (and if so what type) by looking at existing UserChange records, rather than the previous state of the User record
* All expectations of the UserChange attributes stay the same as now, except for one - a 'Remove' type no longer has all of the current attributes blanked out (after a call with Computacenter, we believe this will make it easier for them to process)

Also note that UserChange generation in specs is now _independent_ of whether Papertrail's versioning is on or off. This only affects specs which are actively testing the UserChange generation, where I've had to explicitly delete existing UserChanges after creating users.

### DIFF
--- a/app/models/computacenter/user_change_generator.rb
+++ b/app/models/computacenter/user_change_generator.rb
@@ -1,12 +1,11 @@
 class Computacenter::UserChangeGenerator
-  attr_reader :user, :last_change_for_user
+  attr_reader :user
 
   def initialize(user)
     @user = user
   end
 
   def generate!
-    @last_change_for_user = Computacenter::UserChange.last_for(user)
     if change_needed?
       change = Computacenter::UserChange.new(consolidated_attributes)
       change.add_original_fields_from(last_change_for_user) if last_change_for_user.present?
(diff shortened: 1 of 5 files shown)


app/models/computacenter/user_change_generator.rb
```diff
@@ -1,194 +1,81 @@
...
-
-    case type_of_update
-    when 'New'
-      return unless after_user.relevant_to_computacenter?
-    when 'Change'
-      return unless after_user.relevant_to_computacenter?
-    when 'Remove'
-      return unless before_user.relevant_to_computacenter?
+  def generate!
+    @last_change_for_user = Computacenter::UserChange.last_for(user)
```
asmega, line 9: small non-blocking thing. can this be extracted to a memoized private method

aldavidson: done 👍

asmega (approved):

PR #550, opened 2020-09-25
[737] Support can search schools by URN

asmega:
### Context

- https://trello.com/c/0RnBhnLD/737-lockdown-dashboard

- Add schools to nav header which links to school search
- Able to search for schools via list of URNS
- This has not been added to `index` as it needs to handle the form and the `POST` and use of query params I feel would have been a fiddly path to go down

#### Header

image

#### Search form

image

#### Search results

image

### Guidance to review

- Login as support user
- Click on schools in the nav
- Paste a list of URNs into search form with valid and invalid URNs
- Should see search results


benilovj (approved):
Looks good. I'm happy for this to be merged as is, so we can get it out quicker, but could you please add a simple feature spec so we don't break this in future?

PR #551, opened 2020-09-25
718 mvp multiple schools

aldavidson:
### Context

See [Trello Card 718](https://trello.com/c/aYQOz3QB/718-spike-handle-multiple-schools-in-a-minimum-viable-way). We urgently need to support users who genuinely administer multiple schools, but the work involved in supporting that impacts lots of areas where the code and the interface implicitly assume that a user can have only one school. So we're iterating towards full support in several smaller steps which can be reviewed and merged independently.

This PR does the bare minimum needed to get the model changed to support multiple schools. There are no interface changes - those can come as a separate dedicated PR later.

* Replace `user belongs_to :school` with `user has_many :user_schools / has_many :schools, through: :user_schools`
* Add adapter methods `school / school=` and `school_id / school_id=` to `User`, which maintain compatibility with the rest of the code and just take the first record from the `schools` association

Further smaller PRs can take on the task of making bits of the code genuinely support multiple schools - for instance, the UserChangeGenerator still uses `user.school&.urn` etc. It will be a pretty straightforward change to extend that to `user.schools.map(&:urn).join('|')` and so on - but I wanted to keep _this_ PR as small as possible 

### Guidance to review

Please pay particular attention to the migration which populates the `user_schools` table from existing data. 
It would be a good idea to dump out a list of existing user_id / school_id pairs before and after running the migration to make sure that it has maintained all existing relationships correctly (for instance, with a command like `User.where.not(school_id: nil).each { |user| puts "#{user.id}\t#{user.school_id}" }`)

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -39,9 +39,6 @@ class User < ApplicationRecord
 
   include SignInWithToken
 
-  attribute :school_id, :integer
-  attribute :relevant_to_computacenter, :boolean
-
   after_save do |user|
     Computacenter::UserChangeGenerator.new(user).generate!
   end
@@ -111,36 +108,16 @@ class User < ApplicationRecord
     responsible_body || school&.responsible_body
   end
 
-  def relevant_to_computacenter
+  def relevant_to_computacenter?
     seen_privacy_notice? && orders_devices?
   end
-  alias :relevant_to_computacenter? :relevant_to_computacenter
 
   def hybrid?
     school_id && responsible_body_id
   end
 
-  def hybrid_setup!
-    return if responsible_body.blank?
-
-    one_school = responsible_body.schools.count == 1
-    only_user = responsible_body.users == [self]
-
-    return unless one_school && only_user
-
-    school = responsible_body.schools.first
-
-    update!(school: school)
-    responsible_body.update_who_will_order_devices('schools')
-    contact = school.contacts.create!(email_address: email_address,
-                                      full_name: full_name,
-                                      role: :contact,
-                                      phone_number: telephone)
-    school.preorder_information.update!(school_contact: contact)
-  end
-
-  # Wrapper methods to ease the transition from user belongs_to school,
-  # to user has_many schools
+  # Wrapper methods to ease the transition from 'user belongs_to school',
+  # to 'user has_many schools'
   def school
     schools.first
   end
@@ -151,7 +128,7 @@ class User < ApplicationRecord
 
   def school_id=(new_school_id)
     user_schools.delete_all
-    user_schools.create(school_id: new_school_id) if new_school_id
+    schools << School.find(new_school_id) if new_school_id
   end
 
   def school=(new_school)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -7,7 +7,8 @@ class User < ApplicationRecord
 
   belongs_to :mobile_network, optional: true
   belongs_to :responsible_body, optional: true
-  belongs_to :school, optional: true
+  has_many :user_schools, dependent: :destroy
+  has_many :schools, through: :user_schools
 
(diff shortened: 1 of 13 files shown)


app/models/user.rb
```diff
@@ -115,6 +116,26 @@ def hybrid?
...
+    schools.first
+  end
+
+  def school_id
+    school&.id
+  end
+
+  def school_id=(new_school_id)
+    user_schools.delete_all
+    schools << School.find(new_school_id) if new_school_id
```
asmega, line 131: i feel this guard should also guard the line above? otherwise a mistake might delete all the associations

app/models/user.rb
```diff
@@ -115,6 +116,26 @@ def hybrid?
...
+  end
+
+  def school=(new_school)
+    user_schools.delete_all
+    schools << new_school if new_school.present?
```
asmega, line 136: i feel this guard should also guard the line above? otherwise a mistake might delete all the associations

spec/factories/users.rb
```diff
@@ -53,7 +53,13 @@
     end
 
     factory :school_user do
-      school
+      transient do
+        school { build(:school) }
```
asmega, line 57: i guess this will go soon. but as there's also a `school` method i'm not sure if quirky behaviour might happen

asmega (approved):
- Looks fine, the guard change isn't needed but i think would be safer.
- I tested the migrations on a small dataset (couple dozen) and seemed fine.
- I lack knowledge in the deploy scripts and not sure how a rollback can take place but i guess we can roll forward instead for the down migration.

tonyheadford (approved):
Tried the migration and all seemed OK on my limited dataset.

app/models/user.rb
aldavidson: I found a few cases in the specs where we do things like `user.update!(school_id: nil)` to remove a user from a school - this construction is to support that case.

app/models/user.rb
aldavidson: Likewise, this supports the case of removing a user from a school by doing `user.update!(school: nil)` etc

spec/factories/users.rb
aldavidson: I had to try quite a few iterations of this before ending up with a way that worked! 
This lets the specs do:
* `build(:school_user)` to have a new school constructed automatically
* `build(:school_user, school: some_other_school_object)` to supply a particular school

etc

aldavidson:
Great, thanks @asmega and @tonyheadford ! I'd like to also have a once-over from @benilovj before merging, if possible

db/migrate/20200922130419_add_user_schools.rb
```diff
@@ -0,0 +1,37 @@
...
+      SELECT      id, school_id, NOW(), NOW()
+      FROM        users
+      WHERE       school_id IS NOT NULL
+    SQL
+    connection.execute populate_table_sql
+
+    add_index :user_schools, %i[user_id school_id], unique: true
+    add_index :user_schools, %i[school_id user_id], unique: true
+
+    remove_column :users, :school_id
```
benilovj, line 20: I'd suggest we do this in a subsequent change....

benilovj (approved):
Seems fine, one minor comment about not dropping `school_id` in the same change...

db/migrate/20200922130419_add_user_schools.rb
aldavidson: the only issue then is that ActiveRecord will read it as an attribute, and we may get conflicts with the overridden `school_id` method.

how about renaming it, to `legacy_school_id` or something similar?

benilovj: 👍  to rename...

aldavidson: done in cc596cf

PR #555, opened 2020-09-27
[Do not merge before #551] 718 allow user changes to show multiple schools

aldavidson:
### Context

Second part of [Trello card 718](https://trello.com/c/aYQOz3QB/718-spike-handle-multiple-schools-in-a-minimum-viable-way)
(Depends on #551 MVP Multiple Schools - DO NOT MERGE until that one is merged first)

* add multiple-school support to `Computacenter::UserChange` records

### Guidance to review

One edge case that came up in implementation was the question of what to do when the user has multiple schools from different responsible bodies. Which one should we choose for `effective_responsible_body` ? I've chosen to use the first - but I'm sure equally-valid arguments could be made for alternatives.

### DIFF
--- a/app/models/computacenter/user_change.rb
+++ b/app/models/computacenter/user_change.rb
@@ -39,5 +39,15 @@ module Computacenter
       )
       self
     end
+
+    def self.for_destroyed_user(user)
+    end
+
+    def self.for_new_user(user)
+    end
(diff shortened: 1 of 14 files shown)


asmega (approved):
Looks fine

PR #559, opened 2020-09-28
[751] TechSource confirmations handle extra cases

asmega:
### Context

- https://trello.com/c/8zERGQht/751-techsource-account-confirmation-fixes

- Handle a case where by a user has recently changed email. When sent an old email we do a look up with the user change table and find the current user.
- Handle case of a user that is no longer computacenter relevant by checking if they were last removed from the user change table

#### Scenario 1

- Change a users email address
- Visit http://localhost:3000/computacenter/techsource
- Enter the users old email address
- User should now have `techsource_account_confirmed_at` set

#### Scenario 2

- Delete a computercenter relevant user
- Visit http://localhost:3000/computacenter/techsource
- Enter the deleted users email address
- Should not show as unprocessed and should show email as processed

### DIFF
--- a/app/models/computacenter/user_change.rb
+++ b/app/models/computacenter/user_change.rb
@@ -17,6 +17,8 @@ module Computacenter
 
     enum type_of_update: { New: 0, Change: 1, Remove: 2 }
 
+    belongs_to :user, optional: true
+
     def self.last_for(user)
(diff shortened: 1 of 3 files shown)


benilovj (approved):
Seems fine to me

PR #560, opened 2020-09-28
[757] Support can add new school users

asmega:
### Context

- https://trello.com/c/T2A6JdPn/757-add-a-new-school-user-via-support

- New page in support to add new school users
- Tweaks to `CreateUserService` to remove overbearing overrides

### Guidance to review

- Login as support user
- Navigate to a school via RB
- Should be a button below users table to add new users
- Click this button
- Should be taken to page to add new user to the school
- Fill it in incorrectly
- Should see errors
- Fix mistakes and submit
- Should create this new user on the school
- Should send email to new user

### DIFF
--- a/app/controllers/support/devices/users_controller.rb
+++ b/app/controllers/support/devices/users_controller.rb
@@ -1,4 +1,20 @@
 class Support::Devices::UsersController < Support::BaseController
+  def new
+    @school = School.find_by(urn: params[:school_urn])
+    @user = @school.users.build
+  end
+
+  def create
+    @school = School.find_by(urn: params[:school_urn])
+    @user = @school.users.build(user_params)
+
+    if @user.save
+      redirect_to support_devices_school_path(urn: @school.urn)
+    else
+      render :new, status: :unprocessable_entity
+    end
+  end
+
   def edit
     @school = School.find_by(urn: params[:school_urn])
     @user = present(@school.users.find(params[:id]))
--- a/app/controllers/support/devices/users_controller.rb
+++ b/app/controllers/support/devices/users_controller.rb
@@ -6,9 +6,10 @@ class Support::Devices::UsersController < Support::BaseController
 
   def create
     @school = School.find_by(urn: params[:school_urn])
-    @user = @school.users.build(user_params)
+    user_attributes = @school.users.build(user_params).attributes
+    @user = CreateUserService.invite_school_user(user_attributes)
 
-    if @user.save
+    if @user.persisted?
       redirect_to support_devices_school_path(urn: @school.urn)
     else
       render :new, status: :unprocessable_entity
--- a/app/services/create_user_service.rb
+++ b/app/services/create_user_service.rb
@@ -8,7 +8,7 @@ class CreateUserService
   end
 
   def self.invite_school_user(user_params = {})
-    user = User.new(user_params.merge(override_params))
+    user = User.new(user_params.merge(approved_at: Time.zone.now))
     if user.save
       InviteSchoolUserMailer.with(user: user).nominated_contact_email.deliver_later
     end
--- a/app/views/support/devices/schools/show.html.erb
+++ b/app/views/support/devices/schools/show.html.erb
@@ -20,6 +20,8 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
+    <%= govuk_button_link_to t('page_titles.new_responsible_body_user'), new_support_devices_school_user_path(school_urn: @school.urn) %>
+
     <h2 class="govuk-heading-l">Users</h2>
 
     <% if @users.present? %>
--- a/app/views/support/devices/users/new.html.erb
+++ b/app/views/support/devices/users/new.html.erb
@@ -1,0 +1,24 @@
+<%- title = t('page_titles.new_school_user') %>
+<%- content_for :title, title_with_error_prefix(title, @user.errors.any?) %>
+<%- content_for :before_content do %>
+  <%= breadcrumbs([
+    { t('page_titles.support_responsible_bodies') => support_devices_responsible_bodies_path },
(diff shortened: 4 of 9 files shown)


app/services/create_user_service.rb
```diff
@@ -1,23 +1,17 @@
 class CreateUserService
   def self.invite_responsible_body_user(user_params = {})
-    user = User.new(user_params.merge(override_params))
+    user = User.new(user_params)
```
benilovj, line 3: I don't think we don't want to be making this change to this method. "All responsible body users get a TechSource account" is a part of the domain logic that correctly lived in the service layer. There isn't any use case for expositing either `ordered_devices` or `approved_at` outside the service.

app/services/create_user_service.rb
```diff
@@ -1,23 +1,17 @@
...
     if user.save
       InviteResponsibleBodyUserMailer.with(user: user).invite_user_email.deliver_later
     end
     user
   end
 
   def self.invite_school_user(user_params = {})
-    user = User.new(user_params.merge(override_params))
+    user = User.new(user_params)
```
benilovj, line 11: I think you've found a bug here... it looks like previously, even if a new school user was invited without a TechSource account, one was still being created for them. I'd suggest we fix the bug as part of this PR.

So for `self.invite_school_user` I think we want to let outside callers to set `orders_devices` but not `approved_at`.

app/views/support/devices/schools/show.html.erb
```diff
@@ -50,6 +50,8 @@
   <p class="govuk-body">None</p>
 <% end %>
 
+<%= link_to 'Add school user', new_support_devices_school_user_path(school_urn: @school.urn), class: 'govuk-button' %>
```
benilovj, line 53:
```suggestion
<%= govuk_button_link_to 'Add school user', new_support_devices_school_user_path(school_urn: @school.urn) %>
```

config/locales/en.yml
```diff
@@ -100,6 +100,7 @@
       can_order_devices: Yes, give them access to the TechSource website
       can_order_devices_hint: This is where they’ll place orders.
       cannot_order_devices: 'No'
+    new_school_user: New school user
```
benilovj, line 103:
```suggestion
    new_school_user: Invite a new user
```
This copy is consistent with the responsible body and school versions of this apge.

benilovj (changes requested):

config/locales/en.yml
asmega: done

app/views/support/devices/schools/show.html.erb
asmega: good spot, i've changed this now

app/services/create_user_service.rb
asmega: I see, I've changed `invite_responsible_body_user ` back to the original

app/services/create_user_service.rb
asmega: Made a tweak, should now work as you've described

app/controllers/support/devices/users_controller.rb
```diff
@@ -1,4 +1,21 @@
...
+
+  def create
+    @school = School.find_by(urn: params[:school_urn])
+    user_attributes = @school.users.build(user_params).attributes
+    @user = CreateUserService.invite_school_user(user_attributes)
+
+    if @user.persisted?
+      redirect_to support_devices_school_path(urn: @school.urn)
+    else
+      render :new
```
benilovj, line 15:
```suggestion
      render :new, status: :unprocessable_entity
```

app/views/support/devices/users/new.html.erb
```diff
@@ -0,0 +1,24 @@
+<%- title = t('page_titles.new_school_user') %>
+<%- content_for :title, title_with_error_prefix(title, @user.errors.any?) %>
+<%- content_for :before_content do %>
+  <%= breadcrumbs([
+    { t('page_titles.support_responsible_bodies') => support_devices_responsible_bodies_path },
+    { @school.responsible_body.name => support_devices_responsible_body_path(@school.responsible_body) },
+    { @school.name => support_devices_school_path(urn: @school.urn) },
+    'New school user',
```
benilovj, line 8:
```suggestion
    title,
```

app/views/support/devices/users/new.html.erb
```diff
@@ -0,0 +1,24 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">
+      <%= title %>
+    </h1>
+
+    <%= form_for @user, url: support_devices_school_users_path(school_urn: @school.urn) do |f| %>
+      <%= f.govuk_error_summary %>
+      <%= render partial: 'shared/school_user_form', locals: { form: f } %>
+      <%= f.govuk_submit 'Create school user' %>
```
benilovj, line 21:
```suggestion
      <%= f.govuk_submit 'Send invite' %>
```

benilovj (approved):
Looks fine now, just a couple of minor suggestions for design consistency

app/controllers/support/devices/users_controller.rb
asmega: done

app/views/support/devices/users/new.html.erb
asmega: done

app/views/support/devices/users/new.html.erb
asmega: done

benilovj:
Just one more thing. We want the add user button to be above the Users list, like so:

image

asmega:
> Just one more thing. We want the add user button to be above the Users list, like so:
> 
> image

Outside this ticket I think there are other places that will also need changing if we want this as the pattern eg. on the RB show page

PR #571, opened 2020-09-29
700 send user changes to Computacenters ServiceNow API

aldavidson:
### Context

[Trello card 700](https://trello.com/c/mYNXeNHI/700-send-user-changes-to-computacenters-servicenow-api)

* Add a class to encapsulate sending a ServiceNow Import API request for a given UserChange
* Add a job class to instantiate an API request with the latest UserChange for a given user, send it, and record the transaction against the UserChange
* Add a call to enqueue one of those jobs whenever a UserChange is generated, so long as the relevant endpoint is defined in Settings

### Guidance to review

The three additions above are done in order with the three commits in this PR - I suggest reviewing it commit-by-commit.
I've tested this against CC's test server, we don't have credentials for any other server yet, so no jobs will run on prod even if this merged and deployed.

### DIFF
--- a/app/models/computacenter/user_change_generator.rb
+++ b/app/models/computacenter/user_change_generator.rb
@@ -10,6 +10,7 @@ class Computacenter::UserChangeGenerator
       change = Computacenter::UserChange.new(consolidated_attributes)
       change.add_original_fields_from(last_change_for_user) if last_change_for_user.present?
       change.save!
+      NotifyComputacenterOfLatestChangeForUserJob.perform_later(@user.id) if Settings.computacenter.service_now_user_import_api.endpoint.present?
       change
     end
   end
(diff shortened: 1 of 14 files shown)


benilovj:
Looks fine at first glance but @tonyheadford and @asmega definitely should take a look.

app/models/computacenter/user_change_generator.rb
```diff
@@ -10,6 +10,7 @@ def generate!
       change = Computacenter::UserChange.new(consolidated_attributes)
       change.add_original_fields_from(last_change_for_user) if last_change_for_user.present?
       change.save!
+      NotifyComputacenterOfLatestChangeForUserJob.perform_later(@user.id) if Settings.computacenter.service_now_user_import_api.endpoint.present?
```
tonyheadford, line 13: Personal taste but I find this a bit jarring in the main logic flow and think it would be better as a separate private method e.g.`notify_computacenter` or the `Settings...` bit made into a private method e.g. `service_now_api_configured?`
I realise that moving it into the UserChange model as `save_and_notify_computacenter!` would feel wrong as well given your queueing against the `user_id` for good reason. Not a biggie, it just stopped me in my tracks where the rest of the changes flowed great.

tonyheadford (approved):
Looks great!

app/models/computacenter/user_change_generator.rb
aldavidson: Totally agree - done in f8fb0650

PR #572, opened 2020-09-29
Fix school status when school user added via support

benilovj:
### Context

Before #560, the only way to add the first user to a school was to invite a school contact. It's now possible for support users to add the first user directly. The school status isn't being updated correctly in this case.

Update the school status inference logic to correctly calculate the status, irrespective of how the first user was added.
Ensure that the school status is recalculated when a school user is added.

### DIFF
--- a/app/models/preorder_information.rb
+++ b/app/models/preorder_information.rb
@@ -27,13 +27,13 @@ class PreorderInformation < ApplicationRecord
 
   # If this method is added, we may need to update School::SchoolDetailsSummaryListComponent
   def infer_status
-    if school_will_order_devices? && school_contact.nil?
+    if school_will_order_devices? && !any_school_users? && school_contact.nil?
       'needs_contact'
-    elsif school_will_order_devices? && school_contacted_at.nil?
+    elsif school_will_order_devices? && !any_school_users? && school_contact.present?
       'school_will_be_contacted'
-    elsif school_will_order_devices? && school_contacted_at.present? && !chromebook_information_complete?
+    elsif school_will_order_devices? && any_school_users? && !chromebook_information_complete?
       'school_contacted'
-    elsif school_will_order_devices? && school_contacted_at.present? && chromebook_information_complete?
+    elsif school_will_order_devices? && any_school_users? && chromebook_information_complete?
       'school_ready'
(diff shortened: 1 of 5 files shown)


asmega (approved):
seems fine

app/models/preorder_information.rb
```diff
@@ -42,6 +42,10 @@ def infer_status
     end
   end
 
+  def refresh_status
```
aldavidson, line 45: minor niggle - can this method be `refresh_status!` as it both saves state, and can throw an exception?

app/models/preorder_information.rb
```diff
@@ -27,13 +27,13 @@ def initialize(*args)
...
   def infer_status
-    if school_will_order_devices? && school_contact.nil?
+    if school_will_order_devices? && !any_school_users? && school_contact.nil?
       'needs_contact'
-    elsif school_will_order_devices? && school_contacted_at.nil?
+    elsif school_will_order_devices? && !any_school_users? && school_contact.present?
       'school_will_be_contacted'
-    elsif school_will_order_devices? && school_contacted_at.present? && !chromebook_information_complete?
+    elsif school_will_order_devices? && any_school_users? && !chromebook_information_complete?
       'school_contacted'
```
aldavidson, line 35: If I'm reading this right, then we're no longer using the `school_contact` field to infer whether the school has been contacted or not. It's now primarily whether the school has any users or not. 

So is "school will be contacted" / "school_contacted" still the right set of names for these two states? (changing the names could of course be a separate PR)

aldavidson (approved):
No blockers, but one name-of-method question, and one slightly deeper question about the state logic

app/models/preorder_information.rb
benilovj: fair - done!

app/models/preorder_information.rb
benilovj: I think we will be able to retire the "school will be contacted" status – it's currently used to park:

- schools we've not contacted yet until the service can scale to handle the volume
- schools we can't contact yet because we can't add users to multiple schools yet

After we solve both, we should revisit the statuses and pare stuff down...

PR #578, opened 2020-09-30
[577] Tweak cannot order page 

asmega:
### Context

- https://trello.com/c/vaC4fRyk/577-order-devices-screen-for-people-without-access-to-techsource

- Change the cannot order page
- Change titles
- Additional copy on they cannot place orders if applicable

### Guidance to review

- Login as school user that cannot `orders_devices`
- Make the school order devices but cannot order at the moment 
- Visit order devices page
- Should see techsource copy at the bottom of the page
- Change user so `orders_devices` is `true` and `techsource_account_confirmed_at` is `1.second ago`
- Reload the page
- Should not see techsource copy

### DIFF
--- a/app/views/school/devices/cannot_order.html.erb
+++ b/app/views/school/devices/cannot_order.html.erb
@@ -1,4 +1,5 @@
-<%- title = t('page_titles.school_cannot_order_devices') %>
+<%- who_cannot_order = @user.orders_devices? ? 'school_user' : 'school' %>
+<%- title = t("page_titles.#{who_cannot_order}_cannot_order_devices") %>
 <%- content_for :title, title %>
 <%- content_for :before_content do %>
   <% breadcrumbs([
@@ -21,5 +22,15 @@
     <p class="govuk-body">You can still <%= govuk_link_to 'request devices for disadvantaged children', school_request_devices_path %> from any year group who:</p>
 
     <%= render partial: 'shared/devices/who_to_request_for_list' %>
+
+    <% unless @user.orders_devices? %>
+      <h2 class="govuk-heading-m">
+        You won’t be able to place orders yourself
+      </h2>
+
+      <p class="govuk-body">You do not have a TechSource account. Someone else will need to place your school’s orders.</p>
+
+      <p class="govuk-body"><%= govuk_link_to "Go to manage users to see who can place orders, or give yourself access", school_users_path %>.</p>
+    <% end %>
   </div>
 </div>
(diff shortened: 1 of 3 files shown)


app/views/school/devices/cannot_order.html.erb
```diff
@@ -21,5 +21,15 @@
     <p class="govuk-body">You can still <%= govuk_link_to 'request devices for disadvantaged children', school_request_devices_path %> from any year group who:</p>
 
     <%= render partial: 'shared/devices/who_to_request_for_list' %>
+
+    <% unless @user.capable_of_placing_orders? %>
```
fofr, line 25: This isn't quite right, this message should only show to users who will never have a TechSource account, not for those who will have an account that just isn't set up yet. ie When they were invited they were not given permission to order.

fofr (changes requested):

app/views/school/devices/cannot_order.html.erb
```diff
@@ -3,7 +3,7 @@
 <%- content_for :before_content do %>
   <% breadcrumbs([
     { 'Home' => school_home_path },
-    'Order devices',
+    title,
```
fofr, line 6: The breadcrumb here is intentionally "Order devices" to match the link that was clicked from the homepage. It's for orientation.

app/views/school/devices/cannot_order.html.erb
fofr: For users who have a pending account, we need to show this design:
https://ghwt-design-history.herokuapp.com/ordering-devices-school/#school-can-order-but-your-techsource-account-is-not-ready-yet

asmega: done, I wasn't aware of the other design

app/views/school/devices/cannot_order.html.erb
asmega: This has been changed back. I picked it up from here https://github.com/DFE-Digital/get-help-with-tech-prototype/pull/14

config/locales/en.yml
```diff
@@ -82,7 +82,7 @@
       responsible_body: Orders will be placed centrally
     privacy_notice: Privacy notice
     school_home: Get devices for your school
-    school_cannot_order_devices: You cannot order devices yet
+    school_cannot_order_devices: Your school cannot order devices yet
```
fofr, line 85: We need both titles, one for order users, and one for non order users:

See these two:
https://ghwt-design-history.herokuapp.com/ordering-devices-school/#you-cannot-order-devices-yet
https://ghwt-design-history.herokuapp.com/ordering-devices-school/#your-school-cannot-order-devices-yet-non-order-user

asmega: Thanks, should be sorted now

benilovj (approved):
Looks fine from a code point of view, will defer to @fofr for design approval.

fofr (approved):

PR #575, opened 2020-09-30
[768] stage gias school data

tonyheadford:
### Context
[Trello card](https://trello.com/c/ZUUxMTK7/768-stage-gias-school-data)
Import school data from GIAS into a staging table.  Make this a daily process.
 
### Changes proposed in this pull request
Remove current school import process
Import school data into staging table for schools
Import school links data into the staging table
Add job for importing the school and link data
Add `sidekiq-scheduler` and a schedule for running the job daily (currently arbitrarily at 5:15am but really we should understand when the GIAS data is updated each day)
Added the sidekiq-scheduler integration to the sidekiq dashboard page.

### Guidance to review
This is the first step for updating school data.  This does not make any changes to the existing school data yet, but imports any relevant changes into the staging table each day.
Running the `bundle exec rake import:schools` task now runs the GIAS to staging process.

PR #583, opened 2020-10-01
Moved school links to separate model and added creates and simple updates to schools

tonyheadford:
### Context
[Trello card](https://trello.com/c/EhwQo0Yd/769-gias-updates-handle-simple-updates-for-schools)
Handle simple updates to schools - creating schools that don't exist and simple attribute changes

Fixes issue with multiple links per school i.e. a school closes and the pupils are moved to more than one other school.
Keeps track of the time of the last update and only applies changes to staging since that date.
Imports new schools from staging
Applies attribute changes to existing schools

### Guidance to review
This change will enable the importing of schools (from staging) and updating existing schools that have attribute changes in staging.  The previous PR #575 updated staging daily from GIAS.

__Note:__ Handling URN changes / closures will follow in subsequent PRs

### DIFF
--- a/app/services/school_update_service.rb
+++ b/app/services/school_update_service.rb
@@ -1,0 +1,59 @@
+class SchoolUpdateService
+  RB_NAME_MAP = {
+    'Bristol, City of' => 'City of Bristol',
+    'Dorset' => 'Dorset Council',
+    'Herefordshire, County of' => 'Herefordshire',
+    'Kingston upon Hull, City of' => 'Kingston upon Hull',
+  }.freeze
+
+  def update_schools
+    # look at the schools that have changed since the last update
+    last_update = Staging::DataUpdateRecord.last_update_for(:schools)
+
+    # simple updates for schools that are open
+    Staging::School.updated_since(last_update).open.each do |staged_school|
+      school = School.find_by(urn: staged_school.urn)
+      if school
+        update_school(school, staged_school)
+      else
+        create_school(staged_school)
+      end
+    end
+
+    Staging::DataUpdateRecord.updated!(:schools)
+  end
+
+private
+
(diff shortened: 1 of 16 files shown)


app/services/school_update_service.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+private
+
+  def update_school(school, staged_school)
+    # update school details
+    attrs = staged_attributes(staged_school)
+    school.update!(attrs)
+  rescue ActiveRecord::RecordInvalid => e
+    Rails.logger.error(e.record.errors)
```
asmega, line 33: would it be useful if these were also sent to sentry?

aldavidson: related - do we care about updates failing? This is fine if we don't, but I don't know enough about the use case for this to say whether we do or not... @tonyheadford ?

aldavidson (approved):
I think the code all looks good, and I think the new `DataStage` namespace is much better.

Just one question about whether the service should silently swallow exceptions or not

tonyheadford (commented):
Doh! clicked resolved by mistake on the comments. It's a good question regarding whether we want to swallow the exceptions or raise them in Sentry.  In this case we're really only copying the data from data staging to the school so there shouldn't be any errors really but we could hit a RB that we couldn't find by name for example here.  In regards to sentry where do the reports go and who monitors them?  Is sentry being used like a catchall so we get some notification of an unhandled error?
@aldavidson @asmega @benilovj

app/services/school_update_service.rb
asmega: Let's skip for now but might be useful later otherwise this will silently fail and we'll miss out on updates

PR #588, opened 2020-10-02
[DO NOT MERGE] [786] Build self isolating designs

asmega:
### Context

- https://trello.com/c/IDucONG1/786-build-designs-for-15-self-isolating

- Copy changes around ordering devices
- Welcome wizard `order_your_own` page now only shows if at that moment in time the school/user is not able to order devices

#### RB cannot order
image

#### School cannot order
image

#### School wizard `order_your_own` page changes
image

### Guidance to review

- Login as RB that orders centrally where no orders can be placed
- Should see relevant copy changes
- Login as school that places orders where no order can be placed
- Should see relevant copy changes
- Go thru school wizard with a school user where orders cannot be placed
- Should see `order_your_own` page
- Go thru school wizard with a school user where orders can be placed
- Should not see `order_your_own` page


config/locales/en.yml
```diff
@@ -83,8 +88,12 @@
...
     school_home: Get devices for your school
     school_can_order_devices: Your school can order devices
-    school_cannot_order_devices: Your school cannot order devices yet
-    school_user_cannot_order_devices: You cannot order devices yet
+    school_cannot_order_devices:
+      title: Your school cannot your full allocation yet
+      can_order_when: 'Your school can order your full allocation when:'
+    school_user_cannot_order_devices:
+      title: You cannot order your full allocation yet
+      can_order_when: 'You can order your full allocation when:'
```
fofr, line 96: Should `can_order_when` be in the `page_titles.` block?

fofr (approved):

config/locales/en.yml
asmega: I don't think having a `page_titles` block particularly makes sense. We want to put more copy into this file and we should be separating by more logical chunks eg. by page to group things together. Otherwise page titles of the pages will be in this section and all the other content for those pages will be in some other section.

fofr: This was the way Apply built their titles, and we reused that pattern. I'm not especially attached to it.

asmega:
This work is no longer needed but we can leave this branch here for reference

PR #590, opened 2020-10-02
[732] If all devices ordered page

asmega:
### Context

- https://trello.com/c/tEnKg7FQ/732-build-all-devices-ordered-page-for-schools

- This is displayed instead of the cannot order page when the user can order but have reached their cap

### Screenshots

image

### Guidance to review

- Login as a school user
- The school can order devices but cap == allocation
- Attempt to user
- Should see the new page instead of the cannot order page

### DIFF
--- a/app/controllers/school/devices_controller.rb
+++ b/app/controllers/school/devices_controller.rb
@@ -1,15 +1,17 @@
 class School::DevicesController < School::BaseController
   def order
-    if @school.can_order? && @school.can_order_devices?
-      if @user.awaiting_techsource_account?
-        render :can_order_awaiting_techsource
-      elsif !@user.orders_devices?
-        render :school_can_order_user_cannot
-      else
+    if @school.can_order_devices_right_now?
+      if @user.has_an_active_techsource_account? && @school.can_order_for_specific_circumstances?
+        render :can_order_for_specific_circumstances
+      elsif @user.has_an_active_techsource_account? && @school.can_order?
         render :can_order
+      elsif @user.awaiting_techsource_account?
+        render :can_order_awaiting_techsource
+      else # user has no techsource account
+        render :school_can_order_user_cannot
       end
-    elsif @school.can_order_for_specific_circumstances? && @school.can_order_devices?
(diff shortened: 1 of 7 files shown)


app/controllers/school/devices_controller.rb
```diff
@@ -10,6 +10,8 @@ def order
       end
     elsif @school.can_order_for_specific_circumstances? && @school.can_order_devices?
       render :can_order_for_specific_circumstances
+    elsif @school.can_order?
```
benilovj, line 13: I'm not able to follow the logic in this `if ... else ...` block without looking at the template names.

`@school.can_order?` and `@school.can_order_devices?` are a particular source of confusion. Would you be able to rename them to something longer but hopefully more explicit?

asmega: Agreed theres quite a few permutations and getting more confusing. I've made it more explicit even though it's not actually needed

spec/controllers/school/devices_controller_spec.rb
```diff
@@ -43,7 +43,7 @@
 
         it 'renders can_order' do
```
benilovj, line 44: this needs to be updated as it no longer matches the expectation...

app/controllers/school/devices_controller.rb
benilovj: thanks. I still think that this logic needs an overhaul because:

1. there's a bunch of similar-sounding methods that do similar things and are quite hard to follow
2. there's too much logic in controllers that needs to be pushed down to the model or service level

however that probably shouldn't block this change. I will pick it up separately as I have some ideas.

benilovj (approved):
Fine for now, except the spec description that needs to be updated.

benilovj (changes requested):
I'm not sure this will display the right thing when the school has placed all their orders for specific circumstances...

spec/controllers/school/devices_controller_spec.rb
asmega: fixed up now

benilovj (approved):

PR #592, opened 2020-10-02
777 gias updates stage trust data

tonyheadford:
### Context
[Trello card](https://trello.com/c/8AvaD8wz/777-gias-updates-stage-trust-data)
Import GIAS trusts to the data stage

Add import and parsing of trust data from GIAS (original import left in place for now).
Add job to schedule daily update of the import staging of the trust data (not sure if this actually needs to be daily?)

### Guidance to review
Running `StageTrustDataJob.perform_now` will download and import the latest Trust data from GIAS to the data stage.
View the imported Trust data via the `DataStage::Trust` model.

__Note:__ this does not affect the `Trust` data in the service as yet, those updates will come in future PRs.

### DIFF
--- a/spec/models/trust_data_file_spec.rb
+++ b/spec/models/trust_data_file_spec.rb
@@ -8,18 +8,18 @@ RSpec.describe TrustDataFile, type: :model do
       let(:attrs) do
         [
           {
-            group_uid: '1001',
+            gias_group_uid: '1001',
             companies_house_number: '01234222',
             name: 'Little Hampton Academy',
             address_1: '12 High St',
-            address_2: 'Little Hampton Count',
+            address_2: 'Little Hampton Court',
             town: 'London',
             postcode: 'NW1 1AA',
             status: 'Open',
-            group_type: 'Multi-academy trust',
+            organisation_type: 'Multi-academy trust',
           },
           {
-            group_uid: '4001',
+            gias_group_uid: '4001',
             companies_house_number: '09933123',
             name: 'Callio Forsythe Academy',
             address_1: 'Big Academy',
@@ -27,7 +27,7 @@ RSpec.describe TrustDataFile, type: :model do
             town: 'Easttown',
             postcode: 'EW1 1AA',
             status: 'Closed',
-            group_type: 'Single-academy trust',
+            organisation_type: 'Single-academy trust',
           },
         ]
       end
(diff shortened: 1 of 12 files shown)


spec/models/trust_data_file_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+          address_2: 'Strange Lane',
+          town: 'Easttown',
+          postcode: 'EW1 1AA',
+          status: 'closed',
+          organisation_type: 'Single-academy trust',
+        )
+      end
+    end
+
+    context 'when a trust is has no companies house number or is not a single or multi academy trust' do
```
benilovj, line 70:
```suggestion
    context 'when a trust has no Companies House number or is not a single- or multi-academy trust' do
```

spec/support/csv_file_helper.rb
```diff
@@ -43,6 +43,20 @@ module CSVFileHelper
     link_type: 'LinkType',
   }.freeze
 
+  TRUST_ATTRS = {
```
benilovj, line 46: This might be over-engineering, but if this constant lived on `TrustDataFile`, then you could use it both in the model and the specs...

benilovj (approved):
Looks good.

Just one question... Wondering about the trade-offs of filtering on the import into the staged area, vs. taking a carbon copy of the source data and filtering in the code that copies from the staged table to the operational model.

spec/support/csv_file_helper.rb
tonyheadford: Nice idea

spec/models/trust_data_file_spec.rb
tonyheadford: Yes, I did think about this and wanted to do just that. I think in the case of the CHN it's used as the unique identifier when looking them up and there were a handful of Trusts that had blank numbers (all closed trusts) but we're now pulling in closed ones and I thought that might just complicate things to have blank ones. If they were ever populated in GIAS we'd pull them in so it seemed fine?  We could pull in the other entities from the file if it was useful though if that's what you meant (i.e. ignore the SAT/MAT filter)?

benilovj: > We could pull in the other entities from the file if it was useful though if that's what you meant (i.e. ignore the SAT/MAT filter)?

I think it's fine the way it is, I was just thinking out loud, it's easy enough to change later and re-fill if these assumptions aren't right...

PR #596, opened 2020-10-05
[789] Responsible body order in school page

asmega:
### Context

- https://trello.com/c/iD7D9zEl/789-build-ordering-from-the-school-page-in-the-responsible-bodies-journey

- Add `DeviceCountComponent` which handle panel to show devices available with optional button action
- Add concept of `viewer` in several areas so content is context to the person viewing it
- Additional I18n strings to handle contextual copy
- Add extra page for when an RB user views a school, they can order devices

#### When an RB user views a school
image

#### New page from school to order devices
image

### Guidance to review

- Login as an RB user
- View a school


app/models/school.rb
```diff
@@ -104,6 +104,21 @@ def preorder_status_or_default
     end
   end
 
+  def perceived_state
```
asmega, line 107: not sure about this one. The state in the database seems fixed and we kind of want a virtual one. Any better ideas would be welcomed

benilovj: i'm not convinced about the need to have a separate status. I think it'd be much simpler to extend `PreorderInformation#infer_status` with some of this logic and update `PreorderInformation#status` appropriately.

benilovj: We'd need to recalculate some existing statuses in the DB but that's OK...

asmega: I've added an extra commit that melds `perceived_state` into `PreorderInformation#infer_status`

app/components/school_preorder_status_tag_component.rb
```diff
@@ -17,15 +18,28 @@ def type
...
+  end
+
+  def text_key
+    hash = {
+      'trust' => :responsible_body,
+      'local_authority' => :responsible_body,
+    }
+
+    hash.fetch(viewer.class.to_s.underscore, :default)
   end
```
benilovj, line 43: I think this would be clearer if it were something like this:
```suggestion
  def text_key
    viewer.class.to_s.underscore.in?(%w[trust local_authority]) ? :responsible_body : :default
  end
```

app/models/preorder_information.rb
```diff
@@ -34,9 +36,17 @@ def infer_status
     elsif school_will_order_devices? && any_school_users? && !chromebook_information_complete?
       'school_contacted'
     elsif school_will_order_devices? && any_school_users? && chromebook_information_complete?
-      'school_ready'
+      if school.std_device_allocation&.has_devices_available_to_order?
```
benilovj, line 39:
```suggestion
      if school.can_order_devices_right_now?
```
I introduced this method on #616 so you'd need a rebase to pick it up.

app/models/preorder_information.rb
```diff
@@ -34,9 +36,17 @@ def infer_status
...
+        'school_can_order'
+      else
+        'school_ready'
+      end
     elsif chromebook_information_complete?
-      'ready'
+      if orders_managed_centrally? && school.std_device_allocation&.has_devices_available_to_order?
```
benilovj, line 45:
```suggestion
      if orders_managed_centrally? && school.can_order_devices_right_now?
```

app/views/responsible_body/devices/schools/show.html.erb
```diff
@@ -7,7 +7,13 @@
       <%= @school.name %>
     </h1>
 
-    <%- unless @school.has_std_device_allocation? %>
+    <%- if @school.has_std_device_allocation? %>
+      <%= render DeviceCountComponent.new(
+        max_count: @school.std_device_allocation.cap,
+        ordered_count: @school.std_device_allocation.devices_ordered,
+        show_action: @school.std_device_allocation.devices_ordered < @school.std_device_allocation.cap,
```
benilovj, line 14: Can we use the school helper methods here? (I think this should show the same thing?)
```suggestion
        show_action: @school.can_order_devices_right_now?,
```

benilovj (approved):

PR #597, opened 2020-10-05
[804] extract privacy page from wizard and make it a single unified standalone page

aldavidson:
### Context

As part of the larger ticket [649 - Support multiple schools per user](https://trello.com/c/lDAgx85e/649-handling-multiple-schools), we'll need to have one `SchoolWelcomeWizard` per user-and-school combination. This means we need to remove the privacy notice from the wizard workflow, and make it standalone (the privacy notice is effectively per user, not per school). A subsequent PR will actually add the school ID to the SchoolWelcomeWizard class and deal with the user journey implications.

In discussion with @fofr we agreed that the privacy notices for school users and responsible body users were the same and could be unified into one - which also means that this work is foundational for supporting multiple RBs per user in future too. 

* change the default step on `SchoolWelcomeWizard` from 'privacy' to 'allocation'
* add a standalone `PrivacyNoticeController` and move the responsible body privacy notice template to that scope 

There should be no noticeable difference in the user flow.

### Guidance to review

* probably a good idea to check actually clicking through with a new user of all three relevant types (RB, school, hybrid)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -75,7 +75,7 @@ private
         "couldn't figure out root_url_for user",
         logger: 'logger',
         extra: {
-          time_at: Time.now,
+          time_at: Time.zone.now,
           user_id: user.id,
         },
         tags: {
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -50,4 +50,27 @@ private
   def not_found
     render 'errors/not_found', status: :not_found and return
   end
+
+  def root_url_for(user)
+    if user.needs_to_see_privacy_notice? && !user.seen_privacy_notice?
+      privacy_notice_path
+    elsif user.is_mno_user?
+      mno_extra_mobile_data_requests_path
+    elsif user.is_responsible_body_user? && !user.hybrid?
+      responsible_body_home_path
+    elsif user.is_school_user?
+      if user.school.preorder_information&.school_will_order_devices? &&
+          user.school.preorder_information&.chromebook_info_still_needed?
+        school_before_you_can_order_path
+      else
+        school_home_path
+      end
+    elsif user.is_computacenter?
+      computacenter_home_path
+    elsif user.is_support?
+      support_internet_service_performance_path
+    else
+      '/'
+    end
+  end
 end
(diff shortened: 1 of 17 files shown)


app/controllers/application_controller.rb
```diff
@@ -50,4 +50,27 @@ def render_404_if_feature_flag_inactive(feature_flag)
...
+        school_before_you_can_order_path
+      else
+        school_home_path
+      end
+    elsif user.is_computacenter?
+      computacenter_home_path
+    elsif user.is_support?
+      support_internet_service_performance_path
+    else
+      '/'
```
fofr, line 73: Should we log when this happens, given it shouldn't?

app/controllers/users_controller.rb
```diff
@@ -10,14 +10,6 @@ def create
 
     if @user.save
       token = SessionService.send_magic_link_email!(@user.email_address)
-      # NOTE: this is purely to allow us to display the link in the footer
-      # REMOVE when we have emails actually being sent
-      identifier = @user.sign_in_identifier(token)
-      @debug_info = {
```
benilovj, line 16: Shoud we also remove all references to the `show_debug_info` feature flag?

db/schema.rb
```diff
@@ -282,12 +282,13 @@
     t.boolean "is_support", default: false, null: false
     t.boolean "is_computacenter", default: false, null: false
     t.datetime "privacy_notice_seen_at"
-    t.bigint "legacy_school_id"
     t.boolean "orders_devices"
+    t.bigint "legacy_school_id"
```
benilovj, line 286: Time to drop this too (maybe a separate PR)

benilovj (approved):
Seems fine. When testing this, would you be able to just confirm that school users with 'orders_devices: true` are still added to `Computacenter::UserChanges` even after this change? They should, but I'm not sure we have any feature-level specs that confirm this...

app/controllers/application_controller.rb
aldavidson: good idea - done

app/controllers/users_controller.rb
aldavidson: yes, good idea - but I'll do that as a separate PR

aldavidson:
> Seems fine. When testing this, would you be able to just confirm that school users with 'orders_devices: true`are still added to`Computacenter::UserChanges` even after this change? They should, but I'm not sure we have any feature-level specs that confirm this...

I've added an example to `spec/features/signing_in_as_different_types_of_user_spec.rb` that covers this (and the opposite case)

PR #603, opened 2020-10-05
Remove the 'public account creation' feature

benilovj:
### Context

We designed and built this feature when we thought the service would allow users to self-register and associate themselves with responsible bodies. The service design has since gone into a different direction and there isn't likely to be a strong user need for this feature going forward.

Remove the associated feature flag and code. This helps reduce the build time.


aldavidson (approved):
:100:

PR #611, opened 2020-10-06
Fix links in the header of the MNO requests table

benilovj:
### Context

We recently switched out the implementation of `govuk_link_to`, from
the one that was defined in the app, to a version that is defined in the
`govuk-components` gem.

This meant that the MNO requests table links weren't being generated quite correctly
and weren't getting the 'govuk-link' class applied to them.

This change hardcodes the path helper to the MNO request index path, so
I've changed the helper name to reflect that.

### DIFF
--- a/app/helpers/view_helper.rb
+++ b/app/helpers/view_helper.rb
@@ -43,7 +43,7 @@ module ViewHelper
     render GovukComponent::Breadcrumbs.new(breadcrumbs: breadcrumbs)
   end
 
-  def sortable_table_header(title, value = title, opts = params)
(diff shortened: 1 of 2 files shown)


fofr (approved):

fofr:
Is this another example of `govuk_link_to` not having the same usage as `link_to`?

benilovj:
> Is this another example of `govuk_link_to` not having the same usage as `link_to`?

Yep, raised an issue with `govuk-components`: https://github.com/DFE-Digital/govuk-components/issues/66

PR #612, opened 2020-10-06
775 gias updates changes to trusts

tonyheadford:
### Context
[Trello card](https://trello.com/c/xuCqURSA/775-gias-updates-changes-to-trusts)

Update Trusts from the staged GIAS data
Adds address to responsible bodies

### Guidance to review
Basic mechanism is the same as updates to schools, the last update is recorded and changes made to the staging area since the last update are applied to the Trusts.
Use `TrustUpdateService.new.update_trusts` to process updates


benilovj (approved):
Looks reasonable to me!

PR #616, opened 2020-10-06
[666] Order users are notified when a school order state allows ordering

asmega:
### Context

If a school cap is set, the order users need to know that they can go in and place their orders.

Out of scope for this PR:

- notifying an order user when their TechSource account has been confirmed (will be a follow-up PR)
- sending an automated email to CC (will be a follow-up PR)

- Inform the order users when the cap for their school has been set
- Feature flag the email, until we've let the feature run in production for a little bit and see from the Slack messages that it's triggering at the right point in time


benilovj (approved):
Approving technically but since I pushed most of the changes on this PR, @asmega should definitely have a look at the parts I've done.

PR #622, opened 2020-10-07
[792] data feed for performance analysis

tonyheadford:
### Context
[Trello card](https://trello.com/c/TTcTxpf6/792-data-feed-for-performance-analysis)
Provide an endpoint that lists each school with an allocation > 0 or cap > 0 as JSON

Adds base class `PerformanceDataController` that requires token for authentication
Adds schools controller that inherits from `PerformanceDataController` to provide the school data

### Guidance to review
Amount of schools returned will be large (19-20,000)
Implementation uses chunked response to attempt to minimise resource use.
You need to configure a performance data access token using an env var:
```
GHWT__SUPPORT__PERFORMANCE_DATA_ACCESS_TOKEN=my-token-here
```
I've just used `SecureRandom.hex` for testing but this could be anything
You can access the endpoint with `curl` as follows (with `jq` to pretty the output and an example token value):
```
curl -H 'Authorization: Bearer 03343bef3b21962c4f8764fdb83b696b' 'http://localhost:3000/support/performance-data/schools' | jq '.'
```
As long as the auth token in the request matches the env-var version all should be good.


app/controllers/support/performance_data/schools_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    response.headers['Content-Type'] = 'application/json'
+    response.stream.write '['
+
+    school_query.find_each.with_index do |school, index|
+      response.stream.write ",\n" unless index.zero?
+      response.stream.write attrs_for(school).to_json
+    end
+
+    response.stream.write ']'
+    response.stream.close
```
aldavidson, line 14: Let's get this on dev to make sure PowerBI is OK with reading a chunked response. If it does have any issues, I guess the fallback alternative would be to render it to a static file every X minutes.

aldavidson (approved):
This all looks good AFAICT. Let's deploy it onto dev / staging and let the analysts try hooking it up to PowerBI :+1:

PR #624, opened 2020-10-08
[811] Allow a user with multiple schools to choose one on login

aldavidson:
### Context

[Trello card 811](https://trello.com/c/qr8eoJww/811-allow-a-user-with-multiple-schools-rbs-to-select-one-on-login) as part of larger ticket [649](https://trello.com/c/lDAgx85e/649-handling-multiple-schools)

* If a user has multiple schools, on signing in they are now shown a list of "Your schools" and can choose one to continue with
* If a user only has one school, it redirects to that school's home as before
* Scope all school URLs with the school URN
* Add the school name as a caption to each title under /schools

Screenshots:

Screenshot from 2020-10-08 13-35-49

Screenshot from 2020-10-08 13-36-07


fofr:
@aldavidson If I wanted to test this locally, what do I need to do to give myself access to multiple schools?

aldavidson:
> @aldavidson If I wanted to test this locally, what do I need to do to give myself access to multiple schools?

At the moment, it's easiest to do it from the console - a user `has_many :schools`, so you can just do: `user.schools << another_school`
(assuming user is you, and another_school is a School object)

fofr (changes requested):
Reviewing this locally, looks really good :100:

- Can the school index page be updated to match the design (title renamed, breadcrumbs added): https://ghwt-design-history.herokuapp.com/multiple-schools/#school-page
- Anywhere there is a breadcrumb, for a multi-school user there should be an extra crumb at the start for "Your schools", and "Home" should be replaced with the school name
- I like the addition of the school name as a caption in the H1 during the school wizard, can we only include it though when the user is a multi-school user
- On the `/what-happens-next` page, the button should say "Finish and go to __school__ page" if it's multi-school

app/controllers/school/base_controller.rb
```diff
@@ -12,12 +12,12 @@ def require_school_user!
   end
 
   def set_school
-    @school = @user.school
+    @school = @user.schools.size == 1 ? @user.school : School.find_by_urn(params[:urn])
```
benilovj, line 15: There's a couple of issues here:

1. Because you're not scoping to the user's schools, once a user has multiple schools, they can access any school they like.
2. A much lesser problem, more like unexpected behaviour: if the user has only one school, any URN they provide yields the school they're attached to, so it looks like a mismatch between the URL entered and the data returned.

The way to fix both:

```suggestion
    @school = @user.schools.find_by!(urn: params[:urn])
```

benilovj: I'm guessing we're missing a controller spec for the case:

- given a user with multiple schools
- when they to access a school they don't have access to
- then they get a 404 when

benilovj (changes requested):

tonyheadford (commented):
LGTM Al not an easy job. Great catch btw Jake.

aldavidson:
> LGTM Al not an easy job. Great catch btw Jake.

Seconded (on both counts 😀 )

aldavidson:
>     * Can the school index page be updated to match the design (title renamed, breadcrumbs added): https://ghwt-design-history.herokuapp.com/multiple-schools/#school-page

Should this change be in all cases, or only when the user has multiple schools?

aldavidson:
> > ```
> > * Can the school index page be updated to match the design (title renamed, breadcrumbs added): https://ghwt-design-history.herokuapp.com/multiple-schools/#school-page
> > ```
> 
> Should this change be in all cases, or only when the user has multiple schools?

Reply given in Slack - for all cases

aldavidson:
> (...)    * I like the addition of the school name as a caption in the H1 during the school wizard, can we only include it though when the user is a multi-school user

@fofr  all done, ready for re-review

benilovj:
Looking good. Could we introduce something like a `schools_breadcrumb` helper that encapsulates the `if..else..` logic that's repeating in a quite large number of templates?

fofr (approved):
Design looks good now

benilovj (approved):

PR #634, opened 2020-10-12
816 delete wizard when user removed from school

aldavidson:
### Context

[Trello card 816](https://trello.com/c/ZC3zBON3/816-check-what-happens-to-a-schoolwelcomewizard-when-its-user-is-removed-from-the-school) - avoids the potential for 'dangling' school welcome wizards that point to a no-longer-existing combination of user and school. There's not currently any way in the interface to remove a user from a school, but there may be soon, so it's good to avoid potential issues in advance.

* Add callback & spec for correct behaviour

* Also whilst debugging this change, I found that spurious UserChange records with nil user_ids were being generated in specs. As far as I can tell this looks like a quirk of transactional handling, but it's worth tidying up - so i did.


benilovj (approved):
Code looks fine but the specs don't appear to cover the CC user change generation logic? Feels like something we should test for, given we've already had bugs around deletion in that area.

PR #637, opened 2020-10-12
Introduce a spec helper for stubbing the outgoing Computacenter API

benilovj:
`Settings.computacenter.outgoing_api.endpoint` is a global setting. Specs
should ensure they don't modify this setting during setup and execution,
otherwise it affects the outcome of other specs.

As a rule, it should be on in the specs (like it is in production). Any resultant
network calls should be stubbed out via the newly introduced spec helper. If it's
necessary to change this setting, the spec should put it back to its original value.

### DIFF
--- a/config/settings/test.yml
+++ b/config/settings/test.yml
@@ -1,6 +1,8 @@
 computacenter:
   outgoing_api:
     endpoint: http://computacenter.example.com/
+    username: user
+    password: pass
 
 govuk_notify:
   # API key for the GOV.UK Notify service, used for sending emails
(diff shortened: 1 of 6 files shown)


tonyheadford (approved):
Looks good although I think `stub_computacenter_outgoing_api_calls` would be a better name, 'allow' sounds like you're enabling calls to go to CC to me

benilovj:
> Looks good although I think `stub_computacenter_outgoing_api_calls` would be a better name, 'allow' sounds like you're enabling calls to go to CC to me

Sure - changed in e672962

PR #742, opened 2020-11-02
[930] Notify support if school can order but no one contacted

asmega:
### Context

- https://trello.com/c/jP4OIh88/930-send-an-automated-email-to-ghwt-support-when-a-school-is-closed-and-it-has-no-users-responsible-for-it

- Contact support if school can order but no one has been contacted as school or responsible body does not have any associated users

#### Email template
image

### Guidance to review

- Find a school that does not have users or a responsible body that does not have any users
- Ensure the school can order devices. ie has correct caps and allocations spaces then change to `can_order`
- Should send email to support regarding the fact the school can order but no one was contacted

### DIFF
--- a/app/mailers/can_order_devices_mailer.rb
+++ b/app/mailers/can_order_devices_mailer.rb
@@ -39,6 +39,14 @@ class CanOrderDevicesMailer < ApplicationMailer
                           personalisation: personalisation)
   end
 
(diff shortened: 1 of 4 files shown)


aldavidson (approved):
LGTM :+1:

benilovj (approved):

PR #748, opened 2020-11-02
[878] Soft user delete

asmega:
### Context

- https://trello.com/c/QPKfn09A/878-allow-support-to-remove-user-access-from-dfe-and-cc-services

- Support can soft deletes users
- This simply sets the `deleted_at` on a user with the current timestamp
- These soft deleted users cannot login
- They are also no longer visible to support
- Soft deleted are sent to CC as if the user had been removed
- All mailers should no longer send to deleted users

### Guidance to review

- Login as support user
- Find a user and delete them
- Support user should no longer see user
- Attempt to login as deleted user
- Should no longer be able to login
- If they were CC relevant the user changes should mark them as to be removed
- As support user attempt to mark the school as being able to order
- This user should not get the email notification

### DIFF
--- a/app/controllers/support/responsible_bodies/users_controller.rb
+++ b/app/controllers/support/responsible_bodies/users_controller.rb
@@ -38,7 +38,7 @@ class Support::ResponsibleBodies::UsersController < Support::BaseController
   def destroy
     @responsible_body = ResponsibleBody.find(params[:responsible_body_id])
     @user = @responsible_body.users.find(params[:id])
-    @user.update(deleted_at: Time.zone.now)
+    @user.update!(deleted_at: Time.zone.now)
 
     redirect_to support_responsible_body_path(@responsible_body)
   end
--- a/app/controllers/support/responsible_bodies/users_controller.rb
+++ b/app/controllers/support/responsible_bodies/users_controller.rb
@@ -40,6 +40,8 @@ class Support::ResponsibleBodies::UsersController < Support::BaseController
     @user = @responsible_body.users.find(params[:id])
     @user.update!(deleted_at: Time.zone.now)
 
+    flash[:success] = 'User has been deleted'
+
     redirect_to support_responsible_body_path(@responsible_body)
   end
 
--- a/app/controllers/support/responsible_bodies/users_controller.rb
+++ b/app/controllers/support/responsible_bodies/users_controller.rb
@@ -1,11 +1,11 @@
 class Support::ResponsibleBodies::UsersController < Support::BaseController
+  before_action :set_responsible_body
+
   def new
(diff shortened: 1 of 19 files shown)


app/mailers/application_mailer.rb
```diff
@@ -2,6 +2,14 @@ class ApplicationMailer < Mail::Notify::Mailer
   default from: 'from@example.com'
   layout 'mailer'
 
+  def template_mail(template_id, headers)
```
asmega, line 5: the alternative is for every mailer to make this check and all new mailers to do the same. These feels less intrusive from a developer point of view. Please let me know if there is a better way of doing this

app/mailers/application_mailer.rb
```diff
@@ -2,6 +2,14 @@ class ApplicationMailer < Mail::Notify::Mailer
...
+    headers[:to] = Array(headers[:to]) - User.deleted.pluck(:email_address)
```
jamesrose, line 6: Scoping the finder to just by the expected recipients would mean we wouldn't need all the deleted users for every mail:

```suggestion
    headers[:to] = Array(headers[:to]) - User.deleted.where(email_address: headers[:to]).pluck(:email_address)
```

This still works regardless of whether `headers[:to]` is a string, array, or nil.

app/models/user.rb
```diff
@@ -29,6 +29,8 @@ class User < ApplicationRecord
   scope :who_can_order_devices, -> { where(orders_devices: true) }
   scope :with_techsource_account_confirmed, -> { where.not(techsource_account_confirmed_at: nil) }
   scope :who_have_seen_privacy_notice, -> { where.not(privacy_notice_seen_at: nil) }
+  scope :deleted, -> { where('deleted_at IS NOT NULL') }
+  scope :not_deleted, -> { where('deleted_at IS NULL') }
```
jamesrose, line 33:
```suggestion
  scope :deleted, -> { where.not(deleted_at: nil) }
  scope :not_deleted, -> { where(deleted_at: nil) }
```

app/controllers/support/responsible_bodies/users_controller.rb
```diff
@@ -35,6 +35,16 @@ def update
     end
   end
 
+  def destroy
+    @responsible_body = ResponsibleBody.find(params[:responsible_body_id])
```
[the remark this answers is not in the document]
asmega: I've made the change. I prefer methods that lazily set the instance variable but that pattern doesn't appear in this code base. In general I'm not a huge fan of before setters like this as any more than one and with conditions and it starts getting really messy fast.

jamesrose (approved):

PR #754, opened 2020-11-03
[950] Fix school status

asmega:
### Context

- https://trello.com/c/OYp10nrK/950-school-status-no-reflected-correctly

- There is a bug in `infer_status` where if the school cannot order but has devices available it would show as `school can order` when it should say `school ready` or `ready`

### Guidance to review

- Sign in as support user
- View a school
- Make he school `order_state` as `cannot_order` but the school should have devices available in an allocation
- Refresh
- Should see status of `School ready` if school is ordering or `ready` if RB orders
- Change the school `order_state` to `can_order`
- Should now say `school can order` if school orders or `responsible body can order`


tonyheadford (approved):
LGTM

PR #773, opened 2020-11-06
[946] updating an RB or school can generate user changes

asmega:
### Context

- https://trello.com/c/dcWNdAET/946-computacenter-arent-getting-soldtos-for-users-when-the-computacenterreference-is-added-to-the-rb-after-the-user-is-last-updated
- This fixes an issue where updating an RB or school does not generate a user change despite the new row would be different

- Updating a RB or school can generate user changes

### Guidance to review

- Create/Find an RB that does not have `computacenter_reference`
- Add a user to the RB that is computacenter relevant
- Note the `Computacenter::UserChange.count`
- Update the RB `computacenter_reference` to some value
- Should increment `Computacenter::UserChange.count` by 1
- This change should have populated the `cc_sold_to_number` 
- Should behave the same for a school update except should update `cc_ship_to_number`

### DIFF
--- a/app/models/responsible_body.rb
+++ b/app/models/responsible_body.rb
@@ -18,6 +18,8 @@ class ResponsibleBody < ApplicationRecord
     closed: 'closed',
   }, _prefix: 'gias_status'
 
+  after_update :maybe_generate_user_changes
+
   def humanized_type
     type.demodulize.underscore.humanize.downcase
   end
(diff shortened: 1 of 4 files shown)


aldavidson (approved):
The code looks fine, just one non-blocking point - can you add '_updating an_ RB or school' to the title of the PR?

PR #780, opened 2020-11-06
[981] stop generating UserChanges for users with a school but no shipTo

aldavidson:
### Context

[Trello card 981](https://trello.com/c/IZa6S3hI/981-stop-userchange-being-created-when-a-school-user-has-a-school-urn-but-no-shipto) - This closes a loophole that means a user thinks they can order, and our system thinks they're all set up, but the user can't actually be setup on TechSource because they are not linked to a `shipTo` reference on Computacenter's side.

stop `Computacenter::UserChangeGenerator` thinking a change is needed in this case (where a user has at least one school, but no corresponding `shipTo` number.... and they're not a `single_academy_trust_user`)

### Guidance to review

From the console:

```
> s = School.where(computacenter_reference: nil).first
=> <School: ....>
> u = User.create!(full_name: '...', email_address: '...', orders_devices: true, privacy_notice_seen_at: Time.now.utc)
=> <User: .....>
> Computacenter::UserChange.latest_for_user(u)
=> nil
```

### DIFF
--- a/app/models/computacenter/user_change_generator.rb
+++ b/app/models/computacenter/user_change_generator.rb
@@ -37,7 +37,12 @@ private
 
   def change_needed?
     @user&.id.present? && \
-      (is_addition? || is_removal? || (is_change? && computacenter_fields_have_changed?))
+      (is_addition? || is_removal? || (is_change? && computacenter_fields_have_changed?)) && \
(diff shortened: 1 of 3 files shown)


tonyheadford (approved):
LGTM

PR #782, opened 2020-11-06
[966] Match cap with allocation if can order

asmega:
### Context

- https://trello.com/c/bSKwhZUM/966-updating-the-allocation-should-update-cap-if-a-school-can-order

- If a school order state is `can_order` and allocation is updated, the cap is also increased to match the allocation

### Guidance to review

- Login as support user
- Find a school that `can_order`
- Increase/decrease the allocation
- Cap value should match the allocation value entered
- Change school so they `can_order_for specific_circumstances` or `cannot_order`
- Update allocation value
- Cap value should not change and remain the same

### DIFF
--- a/spec/services/allocation_updater_spec.rb
+++ b/spec/services/allocation_updater_spec.rb
@@ -1,12 +1,22 @@
 require 'rails_helper'
 
 RSpec.describe AllocationUpdater do
+  let(:mock_request) { instance_double(Computacenter::OutgoingAPI::CapUpdateRequest, timestamp: Time.zone.now, payload_id: '123456789') }
+  let(:mock_update_service) { instance_double(SchoolOrderStateAndCapUpdateService) }
+
+  before do
+    allow(Computacenter::OutgoingAPI::CapUpdateRequest).to receive(:new).and_return(mock_request)
+    allow(mock_request).to receive(:post!)
+    allow(mock_update_service).to receive(:update!)
+  end
+
   subject(:service) do
     described_class.new(school: school, device_type: 'std_device', value: 100)
   end
 
   context 'allocation exists and school can order full allocation' do
     let(:school) { create(:school, :in_lockdown) }
+    let(:update_service) { instance_double('SchoolOrderStateAndCapUpdateService') }
 
     before do
       create(:school_device_allocation,
@@ -22,6 +32,15 @@ RSpec.describe AllocationUpdater do
       expect(allocation.allocation).to be(100)
       expect(allocation.cap).to be(100)
     end
+
+    it 'calls notifies computacenter with cap updates' do
+      allow(SchoolOrderStateAndCapUpdateService).to receive(:new).and_return(mock_update_service)
+
+      service.call
+
+      expect(SchoolOrderStateAndCapUpdateService).to have_received(:new).with(school: school, order_state: school.order_state, std_device_cap: 100, coms_device_cap: 0)
+      expect(mock_update_service).to have_received(:update!)
+    end
   end
 
   context 'allocation does not exist and school can order full allocation' do
@@ -33,6 +52,15 @@ RSpec.describe AllocationUpdater do
       expect(allocation.allocation).to be(100)
       expect(allocation.cap).to be(100)
       expect(allocation).to be_persisted
+    end
+
+    it 'calls notifies computacenter with cap updates' do
+      allow(SchoolOrderStateAndCapUpdateService).to receive(:new).and_return(mock_update_service)
+
+      service.call
+
+      expect(SchoolOrderStateAndCapUpdateService).to have_received(:new).with(school: school, order_state: school.order_state, std_device_cap: 100, coms_device_cap: 0)
+      expect(mock_update_service).to have_received(:update!)
     end
   end
 
--- a/spec/services/allocation_updater_spec.rb
+++ b/spec/services/allocation_updater_spec.rb
@@ -1,0 +1,57 @@
+require 'rails_helper'
+
+RSpec.describe AllocationUpdater do
+  subject(:service) do
+    described_class.new(school: school, device_type: 'std_device', value: 100)
+  end
+
+  context 'allocation exists and school can order full allocation' do
+    let(:school) { create(:school, :in_lockdown) }
+
+    before do
+      create(:school_device_allocation,
+             :with_std_allocation,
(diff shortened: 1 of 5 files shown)


asmega:
hmm actually hang back, this might not actually send the cap update request to CC?

.rubocop.yml
```diff
@@ -49,6 +49,9 @@ RSpec/MultipleExpectations:
 RSpec/ContextWording:
   Enabled: false
 
+RSpec/MessageSpies:
+  Enabled: false
```
asmega, line 53: I had to do this so I can do this:
```ruby
expect(SchoolOrderStateAndCapUpdateService).to receive(:new).with(school: school, order_state: school.order_state, std_device_cap: 100, coms_device_cap: 0).and_call_original
```
Unless someone has any better ideas?

spec/services/allocation_updater_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+      expect(allocation.allocation).to be(100)
+      expect(allocation.cap).to be(100)
+    end
+
+    it 'calls notifies computacenter with cap updates' do
+      expect(SchoolOrderStateAndCapUpdateService).to receive(:new).with(school: school, order_state: school.order_state, std_device_cap: 100, coms_device_cap: 0).and_call_original
+
+      service.call
+
+      expect(mock_request).to have_received(:post!).twice
```
aldavidson, line 37: I don't think this expectation belongs in this spec - as far as testing this service is concerned, it shouldn't need to know about the internals of what the `SchoolOrderStateAndCapUpdateService` does when you call `.update!`, just that an instance gets created with the right params and `.update!` gets called on it. That can be tested with a mock/double, which should mean you don't need to disable the RSpec MessageSpies rubocop option or mock the HTTP request.

spec/services/allocation_updater_spec.rb
```diff
@@ -0,0 +1,84 @@
+require 'rails_helper'
+
+RSpec.describe AllocationUpdater do
+  let(:mock_request) { instance_double(Computacenter::OutgoingAPI::CapUpdateRequest, timestamp: Time.zone.now, payload_id: '123456789') }
```
aldavidson, line 4:
```suggestion
  let(:mock_update_service) { instance_double(SchoolOrderStateAndCapUpdateService) }

  before do
    allow(mock_update_service).to receive(:update!)
  end
```

spec/services/allocation_updater_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+
+    before do
+      create(:school_device_allocation,
+             :with_std_allocation,
+             allocation: 10,
+             cap: 10,
+             school: school)
+
+      allow(Computacenter::OutgoingAPI::CapUpdateRequest).to receive(:new).and_return(mock_request)
+      allow(mock_request).to receive(:post!)
```
aldavidson, line 22:
```suggestion
```

spec/services/allocation_updater_spec.rb
aldavidson, line 37:
```suggestion
      expect(SchoolOrderStateAndCapUpdateService).to receive(:new).with(school: school, order_state: school.order_state, std_device_cap: 100, coms_device_cap: 0).and_call_original

      service.call

      expect(mock_request).to have_received(:post!).twice      allow(SchoolOrderStateAndCapUpdateService).to receive(:new).and_return(mock_update_service)
      service.call

      expect(SchoolOrderStateAndCapUpdateService).to have_received(:new).with(school: school, order_state: school.order_state, std_device_cap: 100, coms_device_cap: 0)
      expect(mock_update_service).to have_received(:update!)
```

aldavidson, line 37:
```suggestion
            allow(SchoolOrderStateAndCapUpdateService).to receive(:new).and_return(mock_update_service)
      service.call

      expect(SchoolOrderStateAndCapUpdateService).to have_received(:new).with(school: school, order_state: school.order_state, std_device_cap: 100, coms_device_cap: 0)
      expect(mock_update_service).to have_received(:update!)
```

spec/services/allocation_updater_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+  end
+
+  context 'allocation does not exist and school can order full allocation' do
+    let(:school) { create(:school, :in_lockdown) }
+
+    before do
+      allow(Computacenter::OutgoingAPI::CapUpdateRequest).to receive(:new).and_return(mock_request)
+      allow(mock_request).to receive(:post!)
+    end
+
```
aldavidson, line 48:
```suggestion
```

spec/services/allocation_updater_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+      expect(allocation.cap).to be(100)
+      expect(allocation).to be_persisted
+    end
+
+    it 'calls notifies computacenter with cap updates' do
+      expect(SchoolOrderStateAndCapUpdateService).to receive(:new).with(school: school, order_state: school.order_state, std_device_cap: 100, coms_device_cap: 0).and_call_original
+
+      service.call
+
+      expect(mock_request).to have_received(:post!).twice
```
aldavidson, line 62:
```suggestion
      allow(SchoolOrderStateAndCapUpdateService).to receive(:new).and_return(mock_update_service)
      service.call

      expect(SchoolOrderStateAndCapUpdateService).to have_received(:new).with(school: school, order_state: school.order_state, std_device_cap: 100, coms_device_cap: 0)
      expect(mock_update_service).to have_received(:update!)
```

.rubocop.yml
aldavidson: I'd prefer not to disable the Rubocops if possible, they do usually indicate that something could be better. 
In this case I think it's an expectation that doesn't need to be in the spec - see my [comment on the expectation](https://github.com/DFE-Digital/get-help-with-tech/pull/782/files#r519729286) and there's a batch of suggested changes that address it

aldavidson (changes requested):
Hi Phil, looks good, I've made a batch of suggestions to the spec that mean you don't need to disable the MessageSpies.

asmega:
@aldavidson The rubocop rule has been put back and your suggested changes should now be present

aldavidson (approved):
:+1: LGTM

PR #787, opened 2020-11-09
bump govuk-components to 0.5.0, and fix govuk_link_to calls with blocks

aldavidson:
..to allow the upgrade to work

### Context

There's been a [dependabot PR that upgrades `govuk-components` from 0.4.0 to 0.5.0](https://github.com/DFE-Digital/get-help-with-tech/pull/629) for a while now, but it has a few specs that are failing.

* fix the `govuk_link_to ... do` calls that are failing, by changing them to non-block syntax
* bump the gem version

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ gem 'faker'
 # Manage multiple processes i.e. web server and webpack
 gem 'foreman'
 gem 'govuk_design_system_formbuilder'
-gem 'govuk-components'
+gem 'govuk-components', '>=0.5.0'
 
 gem 'http'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -188,7 +188,7 @@ GEM
       net-http-pipeline
     globalid (0.4.2)
       activesupport (>= 4.2.0)
-    govuk-components (0.4.0)
+    govuk-components (0.5.0)
       rails (~> 6.0.3, >= 6.0.3)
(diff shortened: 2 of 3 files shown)


aliuk2012 (approved):

benilovj:
This feels like a regression – I've been meaning to provide feedback into the `govuk-components` gem, the block syntax should work really...

aldavidson:
> This feels like a regression – I've been meaning to provide feedback into the `govuk-components` gem, the block syntax should work really...

Agreed, but we can do that separately. In this case there's no particular need for those calls to be using the block syntax and it was a straightforward fix. I'm keen to not have broken PRs hanging around any longer than necessary

aldavidson:
> This feels like a regression – I've been meaning to provide feedback into the `govuk-components` gem, the block syntax should work really...

@benilovj I did some experimentation on this last night, and I got to the bottom of the problem. There's a [full description on Trello ](https://trello.com/c/kIL5UsgY/986-feedback-to-govuk-components-that-govuklinkto-with-block-syntax-is-broken-in-050), but the TL;DR version is:

* it's not broken, the (correct) block syntax works fine
* it's our calling code that was broken (incorrect arguments)
* we didn't notice it was broken until 0.5 because the syntax we called was accidentally correct for a non-block call, and in 0.4 _all_ calls are non-block calls

PR #747, opened 2020-11-02
Added closed schools page for CC

tonyheadford:
### Context
[Trello card](https://trello.com/c/fabYSEkV/927-a-table-of-all-closed-schools-for-cc)
Add a table of closed schools to Computacenter area

### Changes proposed in this pull request
Add closed schools table showing RB and device/coms caps
Can view only those schools that can order for specific circumstances, those that can order or both.
Add link to closed schools on computacenter home page.
Added pagination to the closed school table

### Guidance to review
image
=====
image
=====
image

PR #791, opened 2020-11-09
Tweaks to the 'closed schools' CC report

benilovj:
### Context

PR #747 introduced a 'closed schools' report for Computacenter. This PR includes some tweaks some of the functionality and language in this report.

- Temporarily turn off pagination within the report
  - It's my hunch that the CC ops team is going to use Ctrl+F to find school caps using the `URN`, `shipTo`, or `soldTo` numbers. Because of pagination, it's very unlikely that the school they're looking for is going to be on the first page and there's currently no other search functionality included.
  - I've not actually removed the pagination, I've just made the pages contain 10,000 items. If my hunch is incorrect, then it's easy enough to change this back, or if it's correct, then some of the code can be removed in a subsequent PR.
- Fix an N+1 query problem in the controller (which becomes noticeable/very acute when pagination is disabled)
- Re-align some of the report language with the rest of the service:
  - `School closures` => `Schools that can order` (because the report also includes schools that can order for specific circumstances and don't, strictly speaking, have to be closed)
  - `Partially closed` => `Ordering for specific circumstances` (same thinking as above)
  - `Fully closed` => `Ordering due to COVID-19 disruption` (again, schools don't have to be fully closed in order to have the `can_order` order state)

### Guidance to review

Is the new terminology correct?

### DIFF
--- a/app/controllers/computacenter/closed_schools_controller.rb
+++ b/app/controllers/computacenter/closed_schools_controller.rb
@@ -28,15 +28,15 @@ private
 
   def parse_view_mode
     mode = params[:view]
-    mode = 'all' unless mode.in? %w[partially-closed fully-closed]
+    mode = 'all' unless mode.in? %w[specific-circumstances closures-or-self-isolating]
     mode
   end
 
   def query_for_view_mode
     case view_mode
-    when 'partially-closed'
+    when 'specific-circumstances'
       :can_order_for_specific_circumstances
-    when 'fully-closed'
+    when 'closures-or-self-isolating'
       :can_order
     else
       :that_can_order_now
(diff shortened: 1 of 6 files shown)


benilovj:
image

tonyheadford (approved):
LGTM, yes the naming is difficult.

emmajhf (commented):
Hi @benilovj I think these changes are good. One thought: do we want to change 'Ordering due to COVID-19 disruption' to 'Ordering for closures or self-isolating pupils'. This is more specific, and it's arguable that 'disruption' covers everything.

fofr:
Looks good, I agree with @emmajhf.

It also looks like the tabs are overrunning the page container a little, the left of the All tab should line up with the S for schools.

PR #794, opened 2020-11-10
[972] Log Computacenter cap calls to DB

asmega:
### Context

- https://trello.com/c/GAZWcXlU/972-cap-updates-are-being-logged-our-side-but-cc-dont-have-records

- When making calls to CC we now log the request and response in the DB against the allocation
- This is so we can determine what actually happens when something does not work as expected with evidence 

### Guidance to review

- Add settings to connect to CC test environment
- Update a cap that would initiate a call to CC
- The request and response should be logged against the allocation record

### DIFF
--- a/app/models/cap_update_call.rb
+++ b/app/models/cap_update_call.rb
@@ -1,0 +1,3 @@
+class CapUpdateCall < ApplicationRecord
+  belongs_to :school_device_allocation
+end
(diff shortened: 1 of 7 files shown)


aldavidson (approved):
LGTM :+1:

benilovj (approved):

PR #800, opened 2020-11-11
Added SAT converter for schools

tonyheadford:
### Context
[Trello card](https://trello.com/c/yK1dGT8m/999-convert-school-to-sat)
Add ability to convert a non-maintained school into a SAT

Adds `SchoolToSatConverter` that handles the conversion
Add support doc
Handle Trust without `:companies_house_number` in exports to computacenter

### Guidance to review
Intended to be run from the rails console in support ops.

`SchoolToSatConverter.new(School.find_by(urn: 123456)).convert_to_sat`

This will create a new single academy trust, using the name of the school and move the given school into it.
Trust name and companies house number are optional params to `convert_to_sat`

### DIFF
--- a/app/models/computacenter/responsible_body_urns.rb
+++ b/app/models/computacenter/responsible_body_urns.rb
@@ -36,6 +36,8 @@ module Computacenter::ResponsibleBodyUrns
       when 'LocalAuthority'
         "LEA#{gias_id}"
       when 'Trust'
+        return '' if companies_house_number.blank?
+
         "t#{companies_house_number.to_i}"
       end
     end
--- a/app/services/school_to_sat_converter.rb
+++ b/app/services/school_to_sat_converter.rb
@@ -1,0 +1,40 @@
+class SchoolToSatConverter
+  attr_reader :school, :trust
+
+  def initialize(school)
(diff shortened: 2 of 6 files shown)


app/models/computacenter/responsible_body_urns.rb
```diff
@@ -36,6 +36,8 @@ def computacenter_identifier
       when 'LocalAuthority'
         "LEA#{gias_id}"
       when 'Trust'
+        return '' if companies_house_number.blank?
```
jamesrose, line 39: @tonyheadford Out of curiosity, why do we do this? Perhaps I've missed something in the doc.

tonyheadford: In these cases we're making up a responsible body to hang the school on - a school has to have an RB in the service.  We may not be able to find a suitable companies house number for the school (as it is a Trust attribute) so we have to handle the case where it could be blank.  The companies house number is used as an identifier for CC (I don't know the exact reason why), they have agreed that we could send them blank ones for these.  Does that help, not sure I'm answering the right question!

asmega (approved):

PR #802, opened 2020-11-11
Bugfix for crash on the 'request devices for specific circumstances' page

benilovj:
### Context

When a user is part of both a responsible body and a school, and tries to access the 'request devices for specific circumstances' page via the responsible body side of the service, an exception occurs:

https://sentry.io/organizations/dfe-get-help-with-tech/issues/2018379706/?environment=production&project=5320558&query=is%3Aunresolved

This is an edge-case and should be only affecting a small number of users.

Fix the crash

### Guidance to review

Is this edge case worth a spec? Not sure...

### DIFF
--- a/app/views/shared/devices/request_devices.html.erb
+++ b/app/views/shared/devices/request_devices.html.erb
@@ -1,4 +1,4 @@
-<% if @user.is_school_user? %>
+<% if @school %>
   <%- title = t('page_titles.school_request_devices') %>
   <%- content_for :before_content do %>
     <% if @school.mno_feature_flag %>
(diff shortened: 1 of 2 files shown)


jamesrose (approved):
Looks good to me. I do think it is worth adding a spec if this is a real situation, albeit a small set of users.

tonyheadford (approved):
I can't see the sentry logs so not sure what the error was, so a spec would help here, but otherwise all good.

benilovj:
Thanks both, I've added specs which cover:

- the case when the user is from the responsible body only
- the case when the user is a member of both the RB and some schools but views the page from the RB side

I'm pretty sure there are already specs covering the journey from the school side.

PR #805, opened 2020-11-11
Redirect /start and /sign-in if user is already signed in

fofr:
Send users directly to their appropriate home page

In research we've observed signed in users reaching the "Sign in" and "Start" pages again, and unnecessarily going back through the sign in flow.


benilovj (approved):
Great!

PR #806, opened 2020-11-11
Enable support users to edit a school's Chromebook details

benilovj:
### Context

In the case that:

1. A school or RB user orders some Chromebooks
2. Post-ordering, Computacenter determine that the school's Chromebook details are missing or wrong
3. They inform the school/RB
4. If the school user writes back into support with their Chromebook details, instead of entering the details into the service

then the support agent has to explain the user to go enter it into the service themselves. This means wasted effort and a delay for the user.

Allow support agents to update Chromebook details (yes/no on whether the school needs Chromebooks, the domain and recovery email), in case they get a ticket with those details.

### Guidance to review

The wording is a bit off ("when you order", "we", etc) as a result of reusing partials from the RB and school parts of the service, but it felt like a reasonable trade-off to having to duplicate the templates just for the support interface.

#### Screenshots

image

image

### DIFF
--- a/app/components/support/school_details_summary_list_component.rb
+++ b/app/components/support/school_details_summary_list_component.rb
@@ -57,7 +57,13 @@ private
   end
 
   def chromebook_rows_if_needed
-    super.map { |row| row.except(:change_path, :action, :action_path) }
+    super.map do |row|
+      row
+        .except(:change_path, :action, :action_path)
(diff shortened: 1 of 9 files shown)


fofr:
Nice 👍

jamesrose (approved):
Nice work!

PR #809, opened 2020-11-12
Authorise /support with pundit

benilovj:
### Context

We wish to introduce more fine-grained authorisation for various parts of the support interface. This will immediately allow us to add a read-only view for Computacenter users.

Introduce [Pundit](https://github.com/varvet/pundit), a library/DSL for encapsulating authorisation.
Hide any parts of the UI if a user is unable to access the area.

### Guidance to review

* I've disabled a couple of Rubocop cops in a couple of places, where it felt appropriate – would like to hear thoughts if others agree
* I'm not 100% sure about the right way to split testing between controller and policy specs.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -54,12 +54,14 @@ RSpec/EmptyLineAfterSubject:
   Exclude:
     - 'spec/factories.rb'
 
-# the Pundit RSpec DSL triggers this cop false positively
+# The Pundit RSpec DSL triggers this cop false positively
+# Although this may be resolved fairly soon, see https://github.com/rubocop-hq/rubocop-rspec/issues/333
 RSpec/RepeatedExample:
   Exclude:
     - 'spec/policies/**/*'
 
-# the Pundit RSpec DSL triggers this cop false positively
+# The Pundit RSpec DSL triggers this cop false positively
+# Although this may be resolved fairly soon, see https://github.com/rubocop-hq/rubocop-rspec/issues/333
 RSpec/RepeatedDescription:
   Exclude:
     - 'spec/policies/**/*'
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -54,6 +54,16 @@ RSpec/EmptyLineAfterSubject:
   Exclude:
     - 'spec/factories.rb'
 
+# the Pundit RSpec DSL triggers this cop false positively
+RSpec/RepeatedExample:
+  Exclude:
+    - 'spec/policies/**/*'
+
+# the Pundit RSpec DSL triggers this cop false positively
+RSpec/RepeatedDescription:
+  Exclude:
+    - 'spec/policies/**/*'
+
 Naming/MethodParameterName:
   AllowedNames:
     - e
--- a/app/controllers/support/base_controller.rb
+++ b/app/controllers/support/base_controller.rb
@@ -1,13 +1,23 @@
 class Support::BaseController < ApplicationController
+  include Pundit
+
   before_action :require_dfe_user!
+  after_action :verify_authorized
+
+  rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
 
 private
 
   def require_dfe_user!
-    if !SessionService.is_signed_in?(session)
+    unless SessionService.is_signed_in?(session)
       redirect_to_sign_in
-    elsif !@current_user.is_support?
-      render 'errors/forbidden', status: :forbidden
     end
   end
+
+  def user_not_authorized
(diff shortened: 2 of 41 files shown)


app/views/support/schools/show.html.erb
```diff
@@ -17,21 +17,25 @@
...
     <h2 class="govuk-heading-l">Users</h2>
 
     <% if @users.present? %>
       <% @users.each do |user| %>
         <div class="user">
           <h3 class="govuk-heading-m govuk-!-margin-bottom-1"><%= user.full_name %></h3>
-          <p class="govuk-body">
-            <%= govuk_link_to "Edit user<span class=\"govuk-visually-hidden\"> #{user.full_name}</span>".html_safe, edit_support_school_user_path(school_urn: @school.urn, id: user.id) %>
-          </p>
+          <% if policy(User).edit? %>
```
jamesrose, line 34:
```suggestion
          <% if policy(user).edit? %>
```

Is it worth passing in the user instance if we have it?

jamesrose (commented):
Looks good - I like it. I can see how this would make it straight forward to add another user type in support which you mentioned was likely to happen. 

Have you thought about how we might test controllers that allow equal access for both user types? Ideally we wouldn't need to duplicate everything for each.

app/views/support/schools/show.html.erb
benilovj: done

benilovj:
@jamesrose @aldavidson I think this PR is finally ready for review.

app/controllers/support/base_controller.rb
```diff
@@ -1,13 +1,24 @@
 class Support::BaseController < ApplicationController
+  include Pundit
+
   before_action :require_dfe_user!
+  after_action :verify_authorized
+  after_action :verify_policy_scoped, only: %i[index results] # # rubocop:disable Rails/LexicallyScopedActionFilter
```
aldavidson, line 6: `index` makes sense, but it seems awkward to be defining actions with `only` on a non-standard method (results) that's only in a couple of controllers - TBH I can see what Rubocop is getting at here.

In fact, having checked the code in question, arguably the right thing to do here is rename the `results` methods to `index`, but even with or without that, there's a Rubocop-compliant alternative:

```
after_action :verify_policy_scoped, if: :index_method?

...
private
  def index_method?
    %i[index results].include?(action_name)
  end
  
```

.rubocop.yml
```diff
@@ -54,6 +54,16 @@ RSpec/EmptyLineAfterSubject:
...
+# the Pundit RSpec DSL triggers this cop false positively
+RSpec/RepeatedExample:
+  Exclude:
+    - 'spec/policies/**/*'
+
+# the Pundit RSpec DSL triggers this cop false positively
+RSpec/RepeatedDescription:
+  Exclude:
+    - 'spec/policies/**/*'
+
```
aldavidson, line 66: I'm OK with disabling these (at least for now, as [the Rubocop team have been working on this issue](https://github.com/rubocop-hq/rubocop-rspec/issues/333)) :+1:

aldavidson (approved):
Overall, I think this is great! :1st_place_medal:  

I've made a suggestion about the `:before_action` in `Support::BaseController` that will remove the need to disable the LexicallyScopedActionFilter Rubocop, but it's not a blocker

I also think the split between controller / policy testing is about right - the controller tests should be as if you know nothing about _how_ the controller does what it does, and the policy tests are more like traditional unit tests

.rubocop.yml
benilovj: Thanks for digging into it, I've added a link to that PR here: a2f6ffd

app/controllers/support/base_controller.rb
benilovj: Thanks for the suggestion, I've adopted it. Renaming `results` to `index` didn't feel quite right since the `results` action is current a POST in both instances, not a GET, and making it into a GET with a query param would have meant putting the emails search strings into the URL...

benilovj:
I'll give @jamesrose a chance to review this tomorrow..

jamesrose (commented):
Echoing @aldavidson's comment: I think this is great! 🎉 

As discussed, I think CC need an additional ability to be able to edit Chromebook information. As far as I'm aware this the only non read-only functionality they need.

benilovj:
> As discussed, I think CC need an additional ability to be able to edit Chromebook information. As far as I'm aware this the only non read-only functionality they need.

Is it OK if I do this on a follow-up PR? Keen to get this one merged...

PR #811, opened 2020-11-12
[1006] Support can see mno request counts

asmega:
### Context

- https://trello.com/c/eNgLhIpl/1006-show-mno-requests-per-school-in-support-dont-show-personal-info

- In support when viewing a school show extra mobile data request counts and breakdown

### Screenshots

image

image

### Guidance to review

- Login as support user
- Find a school with no extra data requests and view the school
- Should see 0 requests
- Find a school with extra data requests - or may have to setup this data
- Should see extra data request counts and breakdown

### DIFF
--- a/app/components/support/school_details_summary_list_component.rb
+++ b/app/components/support/school_details_summary_list_component.rb
@@ -2,7 +2,7 @@ class Support::SchoolDetailsSummaryListComponent < ResponsibleBody::SchoolDetail
   def rows
     array = super
     array << headteacher_row if headteacher.present?
-    array.insert(array.find_index { |row| row[:key] == 'Can place orders?' }, mno_row)
+    array.insert(array.find_index { |row| row[:key] == 'Can place orders?' }, mno_row) if @school.mno_feature_flag?
     array
   end
 
--- a/app/components/support/school_details_summary_list_component.rb
+++ b/app/components/support/school_details_summary_list_component.rb
@@ -1,10 +1,31 @@
 class Support::SchoolDetailsSummaryListComponent < ResponsibleBody::SchoolDetailsSummaryListComponent
   def rows
-    super +
-      (headteacher.present? ? [headteacher_row] : [])
+    array = super
+    array << headteacher_row if headteacher.present?
+    array.insert(array.find_index { |row| row[:key] == 'Can place orders?' }, mno_row)
+    array
   end
 
 private
+
+  def mno_row
+    {
+      key: 'Extra mobile data',
+      value: mno_value,
+    }
+  end
+
+  def mno_value
+    [
+      "Total: #{@school.extra_mobile_data_requests.count}",
+      "Requested: #{@school.extra_mobile_data_requests.requested.count}",
+      "In progress: #{@school.extra_mobile_data_requests.in_progress.count}",
+      "Queried: #{@school.extra_mobile_data_requests.queried.count}",
+      "Complete: #{@school.extra_mobile_data_requests.complete.count}",
+      "Cancelled: #{@school.extra_mobile_data_requests.cancelled.count}",
+      "Unavailable: #{@school.extra_mobile_data_requests.unavailable.count}",
+    ].join('<br>').html_safe
+  end
 
   def who_will_order_row
     super.except(:change_path, :action, :action_path)
--- a/app/components/support/school_details_summary_list_component.rb
+++ b/app/components/support/school_details_summary_list_component.rb
@@ -16,8 +16,7 @@ private
   end
 
   def mno_value
-    [
-      "Total: #{@school.extra_mobile_data_requests.count}",
+    description = [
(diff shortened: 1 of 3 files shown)


app/components/support/school_details_summary_list_component.rb
```diff
@@ -1,11 +1,32 @@
 class Support::SchoolDetailsSummaryListComponent < ResponsibleBody::SchoolDetailsSummaryListComponent
   def rows
-    super +
-      (headteacher.present? ? [headteacher_row] : [])
+    array = super
+    array << headteacher_row if headteacher.present?
+    array.insert(array.find_index { |row| row[:key] == 'Can place orders?' }, mno_row)
```
benilovj, line 5: can we only show this row if the `mno_feature_flag` is enabled?

app/components/support/school_details_summary_list_component.rb
```diff
@@ -1,11 +1,32 @@
...
+    array
   end
 
 private
 
+  def mno_row
+    {
+      key: 'Extra mobile data',
```
benilovj, line 13:
```suggestion
      key: 'Extra mobile data requests',
```

benilovj:
@fofr I tried using a `govuk_summary_list` to get the stats to look a bit tidier, but it generates way too much spacing and the bold doesn't look right:

image

Any ideas?

Here's the code in `app/components/support/school_details_summary_list_component.rb` to generate that:

```ruby
  def mno_value
    govuk_summary_list(classes: 'govuk-summary-list--no-border') do |component|
      component.slot(:row, key: 'Total', value: @school.extra_mobile_data_requests.count)
      component.slot(:row, key: 'Requested', value: @school.extra_mobile_data_requests.requested.count)
      component.slot(:row, key: 'In progress', value: @school.extra_mobile_data_requests.in_progress.count)
      component.slot(:row, key: 'Queried', value: @school.extra_mobile_data_requests.queried.count)
      component.slot(:row, key: 'Complete', value: @school.extra_mobile_data_requests.complete.count)
      component.slot(:row, key: 'Cancelled', value: @school.extra_mobile_data_requests.cancelled.count)
      component.slot(:row, key: 'Unavailable', value: @school.extra_mobile_data_requests.unavailable.count)
    end
  end
```

app/components/support/school_details_summary_list_component.rb
asmega: done

app/components/support/school_details_summary_list_component.rb
asmega: done

fofr:
If it's zero, can we just say "0 requests" rather than the table?

fofr:
Maybe also something like this, I'm not sure the full details of requests will always be needed:

Screen Shot 2020-11-12 at 15 57 46

fofr:
We could try a table for the requests.

asmega:
I've swapped the list out for a details component. Screenshots have been updated to reflect this

app/components/support/school_details_summary_list_component.rb
```diff
@@ -1,11 +1,33 @@
...
+    description = [
+      "Requested: #{@school.extra_mobile_data_requests.requested.count}",
+      "In progress: #{@school.extra_mobile_data_requests.in_progress.count}",
+      "Queried: #{@school.extra_mobile_data_requests.queried.count}",
+      "Complete: #{@school.extra_mobile_data_requests.complete.count}",
+      "Cancelled: #{@school.extra_mobile_data_requests.cancelled.count}",
+      "Unavailable: #{@school.extra_mobile_data_requests.unavailable.count}",
+    ].join('<br>').html_safe
+
+    govuk_details(summary: "Total: #{@school.extra_mobile_data_requests.count}", description: description)
```
fofr, line 28: Can we say `10 requests` rather than `Total: 10`

asmega: Yep, this has been done

app/components/support/school_details_summary_list_component.rb
```diff
@@ -1,11 +1,33 @@
 class Support::SchoolDetailsSummaryListComponent < ResponsibleBody::SchoolDetailsSummaryListComponent
   def rows
-    super +
-      (headteacher.present? ? [headteacher_row] : [])
+    array = super
+    array << headteacher_row if headteacher.present?
+    array.insert(array.find_index { |row| row[:key] == 'Can place orders?' }, mno_row) if @school.mno_feature_flag?
```
jamesrose, line 5: Part of me wonders whether we should extract all these row keys into one place  like a constant or a translation in `en.yml`.

jamesrose (approved):

app/components/support/school_details_summary_list_component.rb
asmega: Yea, we have a few magic strings floating about

PR #828, opened 2020-11-16
[926] Log devices ordered updates against school

asmega:
### Context

- https://trello.com/c/xdUha5ih/926-fix-number-of-ordered-devices-including-over-orders

- Add a new table and log CC devices ordered updates to this table

### Guidance to review

- Emulate a incoming CC request to update devices ordered
- Should see this now being logged to new table

### DIFF
--- a/app/models/computacenter/devices_ordered_update.rb
+++ b/app/models/computacenter/devices_ordered_update.rb
@@ -1,0 +1,9 @@
+module Computacenter
+  class DevicesOrderedUpdate < ApplicationRecord
+    self.table_name = 'computacenter_devices_ordered_updates'
+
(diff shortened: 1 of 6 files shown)


app/models/computacenter/devices_ordered_update.rb
```diff
@@ -0,0 +1,8 @@
+module Computacenter
+  class DevicesOrderedUpdate < ApplicationRecord
+    self.table_name = 'computacenter_devices_ordered_updates'
+
+    belongs_to :school, primary_key: :computacenter_reference,
```
asmega, line 5: can this belong to an RB instead? is it kind of polymorphic?

jamesrose (approved):

PR #829, opened 2020-11-16
Initial virtual pool implementation

tonyheadford:
### Context
[Trello card](https://trello.com/c/Nj7Thku9/988-virtual-cap-calculate-virtual-cap)
Calculate virtual caps. This is a first iteration of a virtual pool for devices for us to kick about and poke holes in.

Adds `VirtualCapPool` model to hold `:cap` and `:devices_ordered` pooled amounts for specific a `:device_type`
Adds `SchoolVirtualCap` model as a join table between `SchoolDeviceAllocation` and `VirtualCapPool`
Implements accessors for `:cap` and `:devices_ordered` on `SchoolDeviceAllocation` that determine whether to return the pool values or the local object values
'Raw' accessor methods `:raw_cap` and `:raw_devices_ordered` added to access local model values so original values can be accessed when in a pool (e.g. for pool calculation)
Feature flagged using the virtual pool cap and devices ordered values from the `SchoolDeviceAllocation` so enable `:virtual_caps` to use the values directly, otherwise the pool amounts are just logged to the Rails logger when these are accessed and the local values returned.
Added `touch: true` to pass changes up the `SchoolDeviceAllocation` -> `SchoolVirtualCap` -> `VirtualCapPool` chain to trigger a recalculation of the pool amounts (which in turn, down the line, can push changes to CC).

### Guidance to review

There is a migration to run to add the new tables.
A school can be added to its responsible body's virtual cap pool by calling `responsible_body.add_school_to_virtual_cap_pool(school)`
This will iterate the `device_allocations` on the school and add them to the virtual pool for the correct device type ('std_device' or 'coms_device' currently).
The `:cap` and `:devices_ordered` values remain unchanged on the `SchoolDeviceAllocation` object.
The `SchoolDeviceAllocation` uses a new association `:school_virtual_cap` to determine whether it is in a pool or not.
The pool `:cap` and `:devices_ordered` are updated when adding a school.  They can also be updated by invoking `responsible_body.calculate_virtual_caps!` or for a specific pool by `responsible_body.std_device_pool.recalculate_caps!` or `responsible_body.coms_device_pool.recalculate_caps!`.  This methods sums up the `:cap` and `:devices_ordered` for all allocations in the pool.
Once in a pool, changes to a school's `SchoolDeviceAllocation` will trigger a recalculation of the pool's amounts.

### DIFF
--- a/app/models/responsible_body.rb
+++ b/app/models/responsible_body.rb
@@ -8,6 +8,10 @@ class ResponsibleBody < ApplicationRecord
   has_many :extra_mobile_data_requests
   has_many :schools
 
+  has_many :virtual_cap_pools
+  has_one :std_device_pool, -> { std_device }, class_name: 'VirtualCapPool'
+  has_one :coms_device_pool, -> { coms_device }, class_name: 'VirtualCapPool'
+
   scope :excluding_department_for_education, -> { where.not(type: 'DfE') }
 
   extend Computacenter::ResponsibleBodyUrns::ClassMethods
@@ -19,6 +23,17 @@ class ResponsibleBody < ApplicationRecord
   }, _prefix: 'gias_status'
 
   after_update :maybe_generate_user_changes
+
+  def calculate_virtual_caps!
+    virtual_cap_pools.each(&:recalculate_caps!)
+  end
+
+  def add_school_to_virtual_cap_pools(school)
+    school.device_allocations.each do |allocation|
+      pool = virtual_cap_pools.send(allocation.device_type).first_or_create!
+      pool.add_school(school)
+    end
+  end
 
   def humanized_type
     type.demodulize.underscore.humanize.downcase
--- a/app/models/responsible_body.rb
+++ b/app/models/responsible_body.rb
@@ -28,10 +28,10 @@ class ResponsibleBody < ApplicationRecord
     virtual_cap_pools.each(&:recalculate_caps!)
   end
 
-  def add_school_to_virtual_cap_pools(school)
+  def add_school_to_virtual_cap_pools!(school)
     school.device_allocations.each do |allocation|
       pool = virtual_cap_pools.send(allocation.device_type).first_or_create!
-      pool.add_school(school)
+      pool.add_school!(school)
     end
   end
 
--- a/app/models/responsible_body.rb
+++ b/app/models/responsible_body.rb
@@ -8,7 +8,7 @@ class ResponsibleBody < ApplicationRecord
   has_many :extra_mobile_data_requests
   has_many :schools
 
-  has_many :virtual_cap_pools
+  has_many :virtual_cap_pools, dependent: :destroy
   has_one :std_device_pool, -> { std_device }, class_name: 'VirtualCapPool'
   has_one :coms_device_pool, -> { coms_device }, class_name: 'VirtualCapPool'
 
--- a/app/models/virtual_cap_pool.rb
+++ b/app/models/virtual_cap_pool.rb
@@ -1,0 +1,41 @@
+class VirtualCapPool < ApplicationRecord
+  belongs_to :responsible_body
+  has_many :school_virtual_caps, dependent: :destroy
+  has_many :school_device_allocations, through: :school_virtual_caps
+  has_many :schools, through: :school_device_allocations
+
+  enum device_type: {
+    'coms_device': 'coms_device',
+    'std_device': 'std_device',
+  }
+
+  def recalculate_caps!
+    self.cap = school_device_allocations.sum(:cap)
+    self.devices_ordered = school_device_allocations.sum(:devices_ordered)
+    save!
+    # TODO: trigger events for CC here?
+  end
+
+  def add_school(school)
+    if school.responsible_body_id == responsible_body_id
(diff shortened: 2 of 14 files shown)


app/models/virtual_cap_pool.rb
```diff
@@ -0,0 +1,41 @@
...
+  enum device_type: {
+    'coms_device': 'coms_device',
+    'std_device': 'std_device',
+  }
+
+  def recalculate_caps!
+    self.cap = school_device_allocations.sum(:cap)
+    self.devices_ordered = school_device_allocations.sum(:devices_ordered)
+    save!
+    # TODO: trigger events for CC here?
```
jamesrose, line 16: I wonder whether _any_ change on this model should trigger update events to CC?

app/models/virtual_cap_pool.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def add_school(school)
```
jamesrose, line 19: Did you imagine the caller of this would check that this is an eligible school? Or perhaps it belongs here?

jamesrose (commented):
@tonyheadford This is awesome. Shall we catch up this morning to talk about next steps?

app/models/virtual_cap_pool.rb
tonyheadford: I've added some additional checks here to ensure the school is in a `can_order` or `can_order_for_specific_circumstances` state when adding to the pool, but this logic/checks might better be handled by whatever it is that puts the school in that state.

app/models/responsible_body.rb
```diff
@@ -20,6 +24,17 @@ class ResponsibleBody < ApplicationRecord
 
   after_update :maybe_generate_user_changes
 
+  def calculate_virtual_caps!
+    virtual_cap_pools.each(&:recalculate_caps!)
+  end
+
+  def add_school_to_virtual_cap_pools!(school)
```
asmega, line 31: is there a safeguard to accidentally adding a school twice to the pool or is this overkill?

app/models/virtual_cap_pool.rb
```diff
@@ -0,0 +1,49 @@
...
+    else
+      raise VirtualCapPoolError, "Cannot add school to virtual pool #{school.urn} #{school.name}"
+    end
+  end
+
+private
+
+  def school_can_be_added_to_pool?(school)
+    school.responsible_body_id == responsible_body_id &&
+      (school.can_order? || school.can_order_for_specific_circumstances?) &&
```
asmega, line 33: why can one not add a school to the pool that cannot order yet? will we not want to add schools in advance to create pool?

db/schema.rb
```diff
@@ -367,13 +376,25 @@
     t.index ["item_type", "item_id"], name: "index_versions_on_item_type_and_item_id"
   end
 
+  create_table "virtual_cap_pools", force: :cascade do |t|
+    t.string "device_type", null: false
+    t.bigint "responsible_body_id", null: false
```
asmega, line 381: does the column need an index?

spec/models/school_device_allocation_spec.rb
```diff
@@ -82,4 +82,43 @@
...
+    subject(:allocation) { described_class.create!(device_type: 'std_device', cap: 100, devices_ordered: 87, allocation: 100, school: school) }
+
+    ### moving this into an around block causes test fails elsewhere in the suite?!?
+    before do
+      allocation
+      responsible_body.add_school_to_virtual_cap_pools!(school)
+      responsible_body.std_device_pool.update!(cap: 256, devices_ordered: 145)
+      allocation.reload
+
+      @original_state = FeatureFlag.active?(:virtual_caps)
```
asmega, line 99: Al wrote a helper for this eg `context 'when authenticated', with_feature_flags: { mno_offer: 'active' }`

asmega (approved):

app/models/virtual_cap_pool.rb
tonyheadford: Currently the rule is schools can only be added to the pool when they are able to order, but once in they stay in.

spec/models/school_device_allocation_spec.rb
tonyheadford: Oh that sounds good. I was having issues with `around` blocks yesterday, for some reason it was causing unrelated tests in other parts of the suite to fail that worked fine when the around block wasn't used, there is also `FeatureFlag.temporarily_active` that takes a block but I took it all out when trying to fix the tests. But I'll look at that 👍

db/schema.rb
tonyheadford: Will add one

app/models/virtual_cap_pool.rb
aldavidson: I'm trying to figure out the one/many-to-many relationships here.... as far as i can see, a RB can have many VirtualCapPools because it would have one per device type, but no more than 2. Is that right?

If that's the case, then should this model have a uniqueness validation, along the lines of:
`validates :device_type, uniqueness: { scope: [:responsible_body_id] }`
to guard against an RB possibly having multiple pools per device_type ?

tonyheadford: Yes, good call

PR #830, opened 2020-11-17
[1020] prevent computacenter support users seeing users who have not seen the privacy notice

aldavidson:
### Context

[Trello card #1020](https://trello.com/c/x2xRoTKR/1020-hide-contact-information-from-cc-people-for-users-that-havent-yet-seen-the-privacy-notice) - there's a loophole by which the personally identifying contact details of users could be shown to third parties before they have seen the privacy notice, if those third parties are the Computacenter users who have been given access to the support UI.

* add `.safe_to_show_to(user)` scope to `User`
* add `privacy_notice_required: ` argument to existing scopes on `ResponsibleBody` which reference the users table
* update all relevant support controller methods to use those scopes and arguments as appropriate
* update relevant support feature specs to test the new restrictions 
* as the support/users_controller uses `@user` to refer to the user who is the target of an action (show / edit / etc), and we also need to refer to the currently signed-in user (which is set as `@user` in the application_controller.identify_user! method), make the `identify_user!` method also set `@current_user` and use that as an argument to the scopes. 

NOTE that this last conflict needs dealing with properly, as it can lead to a confusing user experience if you visit the support/user_controller - so a separate PR will be raised later to remove the conflict fully.


jamesrose (approved):
Looks good to me! 🚢

app/controllers/support/responsible_bodies_controller.rb
```diff
@@ -3,15 +3,15 @@ def index
     @responsible_bodies = ResponsibleBody
       .select('responsible_bodies.*')
       .excluding_department_for_education
-      .with_users_who_have_signed_in_at_least_once
-      .with_user_count
+      .with_users_who_have_signed_in_at_least_once(privacy_notice_required: @current_user.is_computacenter?)
```
benilovj, line 6: I'm not convinced we need to filter this count... I think CC should still see that the users are there, but they just shouldn't see who they are. Otherwise it looks misleading...

app/controllers/support/responsible_bodies_controller.rb
```diff
@@ -3,15 +3,15 @@ def index
...
+      .with_user_count(privacy_notice_required: @current_user.is_computacenter?)
```
benilovj, line 7: As above ☝️

benilovj (approved):
I think this achieves what it sets out to do, but I think could be made simpler if we don't filter the user counts on the RB list, and just show the same count to all users.

app/controllers/support/responsible_bodies_controller.rb
aldavidson: ditto :)

app/controllers/support/responsible_bodies_controller.rb
benilovj: > Should we also add notes to the user lists that say "You're only seeing users who have seen the privacy policy" or something similar?

@aldavidson yes that's a good idea and the right way to go. @fofr or @ralph-hawkins could we get a bit of micro-design for this message?

fofr: I'm with @aldavidson and think we should filter the count. I don't think CC need to know there are accounts they cannot see.

benilovj:
Fair enough, let's go with the built solution then.

aldavidson:
> Fair enough, let's go with the built solution then.

So... am I ok to merge this as-is?

benilovj:
:shipit:

PR #835, opened 2020-11-19
add facility for support users to move users between schools and RBs

aldavidson:
### Context

[Trello card 1021](https://trello.com/c/swcEUVuq/1021-allow-support-to-move-users-between-schools-and-or-rbs) - requests to change users' associated organisations are fairly regular now, and each time the support team member has to ask a dev to do it for them from the console.

* add UI to allow support users to make those changes themselves

### Guidance to review

This was very much a 'whatever works' UI, with more of an eye on speed of implementation rather than elegance. As such, it could really do with a checkover from @fofr or @ralph-hawkins 

Screenshots:

**Point of entry - new 'Change' link on (existing) user search results**
Screenshot from 2020-11-19 11-21-31

**New page - users' associated organisations**
Screenshot from 2020-11-19 11-22-34

**New page - finding a school to associate**
Screenshot from 2020-11-19 11-26-28

**New page - finding a responsible body to associate**
Screenshot from 2020-11-19 11-28-31

### DIFF
--- a/app/components/support/user_search_result_component.html.erb
+++ b/app/components/support/user_search_result_component.html.erb
@@ -4,5 +4,8 @@
 
 <%= govuk_summary_list(classes: 'govuk-summary-list--no-border') do |component|
   component.slot(:row, key: 'Email address', value: user.email_address)
-  component.slot(:row, key: 'Associated with', value: unordered_list_of_orgs, action: govuk_link_to('Change', associated_organisations_support_user_path(user)) )
+  component.slot(:row,
+                  key: 'Associated with',
+                  value: unordered_list_of_orgs,
+                  action: (UserPolicy.new(current_user, user).editable? ? govuk_link_to('Change', associated_organisations_support_user_path(user)) : nil) )
 end %>
--- a/app/components/support/user_search_result_component.html.erb
+++ b/app/components/support/user_search_result_component.html.erb
@@ -4,5 +4,5 @@
 
 <%= govuk_summary_list(classes: 'govuk-summary-list--no-border') do |component|
   component.slot(:row, key: 'Email address', value: user.email_address)
-  component.slot(:row, key: 'Associated with', value: unordered_list_of_orgs)
+  component.slot(:row, key: 'Associated with', value: unordered_list_of_orgs, action: govuk_link_to('Change', associated_organisations_support_user_path(user)) )
 end %>
--- a/app/components/support/user_search_result_component.html.erb
+++ b/app/components/support/user_search_result_component.html.erb
@@ -7,5 +7,5 @@
   component.slot(:row,
                   key: 'Associated with',
                   value: unordered_list_of_orgs,
-                  action: (UserPolicy.new(current_user, user).editable? ? govuk_link_to('Change', associated_organisations_support_user_path(user)) : nil) )
+                  action: (Pundit.policy(current_user, user).editable? ? govuk_link_to('Change', associated_organisations_support_user_path(user)) : nil) )
 end %>
--- a/app/controllers/support/users/schools_controller.rb
+++ b/app/controllers/support/users/schools_controller.rb
@@ -1,5 +1,6 @@
 class Support::Users::SchoolsController < Support::BaseController
   before_action :set_user, :set_school
+  before_action { authorize User }
 
   def index
     @form = Support::NewUserSchoolForm.new(user: @user, name_or_urn: user_school_params[:name_or_urn])
@@ -26,8 +27,7 @@ class Support::Users::SchoolsController < Support::BaseController
 private
 
   def set_user
-    # @user = User.safe_to_show_to(@current_user).find(params[:id])
-    @user = User.find(params[:id])
+    @user = policy_scope(User).find(params[:id])
   end
 
   def set_school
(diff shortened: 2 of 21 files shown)


app/controllers/support/users/schools_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    @user_school = @user.user_schools.find_by_school_id(@school.id)
+    @user_school&.destroy!
+    flash[:success] = "#{@user.full_name} is no longer associated with #{@school.name}"
+    redirect_to associated_organisations_support_user_path(@user.id)
+  end
+
+private
+
+  def set_user
+    # @user = User.safe_to_show_to(@current_user).find(params[:id])
```
asmega, line 29: can we delete this comment if its not needed?

app/form_objects/support/new_user_school_form.rb
```diff
@@ -0,0 +1,16 @@
...
+  attr_accessor :user, :name_or_urn
+
+  def initialize(user:, name_or_urn: nil)
+    @user = user
+    @name_or_urn = name_or_urn
+  end
+
+  def matching_schools
+    School.includes(:responsible_body)
+          .where('urn = ? OR LOWER(name) LIKE(?)', @name_or_urn.to_i, "%#{@name_or_urn.downcase}%")
```
asmega, line 13: would `ILIKE` work here?

tonyheadford (approved):
LGTM - agree with Phil's comments. One non-blocking question: should we clear any session info for a user if we're changing what they can see i.e. the user is logged in while this is happening?

app/form_objects/support/new_user_school_form.rb
aldavidson: oooh, good call!
 always try to use standard SQL by default rather than vendor-specific extensions, but I think in this case:

1. we gain clarity
2. it's easy to port it back to a standards-compliant way if needed
3. there'll be a record of this commit in the log if we ever _do_ need to change DBs
4. .....how likely is it that we'll ever actually port this to another DB anyway? (not very)

So yeah, done - thanks!

app/controllers/support/users/schools_controller.rb
aldavidson: I've just rebased from main, which means I'm now able to use the `safe_to_show_to` scope - so I've done the opposite (removed the un-restricted find and un-commented this line)

aldavidson:
> LGTM - agree with Phil's comments. One non-blocking question: should we clear any session info for a user if we're changing what they can see i.e. the user is logged in while this is happening?

I think we're OK without - the only thing we hold in session is `user_id` (and occasionally `return_url`), and TBH it's not straightforward to do except for the current_user.

app/views/support/users/associated_organisations.html.erb
```diff
@@ -0,0 +1,106 @@
...
+              <td class="govuk-table__cell">
+                <%= govuk_link_to "#{school.name} (#{school.urn})", support_school_path(urn: school.urn) %>
+                <br>
+                <%= school.type_label %>
+              </td>
+              <td class="govuk-table__cell">
+                <%= govuk_link_to school.responsible_body.name, support_responsible_body_path(id: school.responsible_body_id) %>
+              </td>
+              <td class="govuk-table__cell">
+                <%= govuk_button_to('Remove', support_user_school_path(urn: school.urn, id: @user.id), method: :delete, form_class: 'form-inline')  %>
```
jamesrose, line 46: Should they be able to remove an RB without swapping it for another?

app/views/support/users/associated_organisations.html.erb
```diff
@@ -0,0 +1,106 @@
...
+          </tr>
+        </tbody>
+      </table>
+
+    <%- else %>
+      <p class="govuk-body">(none)</p>
+    <%- end %>
+
+    <%= form_for @user_responsible_body_form, url: responsible_bodies_support_user_path(@user), method: :get do |f| %>
+      <%= f.govuk_fieldset legend: { text: 'Associate with a responsible body', size: 'm' } do %>
```
jamesrose, line 100:
```suggestion
      <%= f.govuk_fieldset legend: { text: 'Move to a different responsible body', size: 'm' } do %>
```

Would something like this be clearer as to what we're doing?

jamesrose (approved):
This is going to save support people a ton of time. Nice work!

Only minor things on whether we allow schools orphaned without RBs. I'm positive @fofr can do better than my copy suggestion, as well.

app/views/support/users/associated_organisations.html.erb
aldavidson: yes - it's entirely possible for users to only be associated with schools, and not with RBs

app/components/support/user_search_result_component.html.erb
```diff
@@ -4,5 +4,8 @@
 
 <%= govuk_summary_list(classes: 'govuk-summary-list--no-border') do |component|
   component.slot(:row, key: 'Email address', value: user.email_address)
-  component.slot(:row, key: 'Associated with', value: unordered_list_of_orgs)
+  component.slot(:row,
+                  key: 'Associated with',
+                  value: unordered_list_of_orgs,
+                  action: (UserPolicy.new(current_user, user).editable? ? govuk_link_to('Change', associated_organisations_support_user_path(user)) : nil) )
```
benilovj, line 10: I **think** this can also be
```suggestion
                  action: (Pundit.policy(current_user, user).editable? ? govuk_link_to('Change', associated_organisations_support_user_path(user)) : nil) )
```
which may abstract the implementation detail of pundit a little bit more...

PR #836, opened 2020-11-19
[926] When CC logs cancellation maybe notify user

asmega:
### Context

- https://trello.com/c/xdUha5ih/926-fix-number-of-ordered-devices-including-over-orders

- When CC makes call to update `devices_ordered`
- If is a cancellation (reduction of `devices_ordered`)
- We might contact user/school/rb they can order if they are in the correct state to be able to order

### Guidance to review

- Make CC call
- If an increase does not send email
- If decrease should send email assuming school can order

### DIFF
--- a/spec/models/computacenter/api/cap_usage_update_spec.rb
+++ b/spec/models/computacenter/api/cap_usage_update_spec.rb
@@ -34,11 +34,11 @@ RSpec.describe Computacenter::API::CapUsageUpdate do
 
   describe '#apply!' do
     let(:preorder) { create(:preorder_information, :rb_will_order, :does_not_need_chromebooks, status: 'ready') }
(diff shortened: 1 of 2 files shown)


spec/models/computacenter/api/cap_usage_update_spec.rb
```diff
@@ -84,5 +84,43 @@
...
+      let(:args) do
+        {
+          'capType' => 'DfE_RemainThresholdQty|Std_Device',
+          'shipTo' => '123456',
+          'capAmount' => 100,
+          'usedCap' => 1,
+        }
+      end
+
+      it 'might send notification to order' do
```
aldavidson, line 100: can the spec be explicit about the circumstances in which it might or might not send the notification? Just saying it 'might' sounds like it's non-deterministic...

aldavidson (approved):
Looks OK, just one (non-blocking) note about the name of an example

PR #837, opened 2020-11-19
[1040] Hide closed schools

asmega:
### Context

- https://trello.com/c/XuWAZaBb/1040-hide-closed-schools-on-service

- Hide closed schools from views
- hidden from RB users and support users
- schools can still be accessed directly in url via urn
- but they are hidden from index views and support search

### Guidance to review

- Login as RB user viewing list of schools
- Close one of the schools
- School should no longer be visible to RB user
- Login as support user
- Close a school
- School should no longer be return via urn search
- View the RB of the closed school
- Should no longer see school in list of school for RB

### DIFF
--- a/app/models/responsible_body.rb
+++ b/app/models/responsible_body.rb
@@ -137,19 +137,19 @@ class ResponsibleBody < ApplicationRecord
   end
 
   def is_ordering_for_schools?
-    schools.that_are_centrally_managed.any?
+    schools.gias_status_open.that_are_centrally_managed.any?
   end
 
   def has_centrally_managed_schools_that_can_order_now?
-    schools.that_are_centrally_managed.that_can_order_now.any?
+    schools.gias_status_open.that_are_centrally_managed.that_can_order_now.any?
   end
 
   def has_schools_that_can_order_devices_now?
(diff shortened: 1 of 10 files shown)


app/models/responsible_body.rb
```diff
@@ -137,19 +137,19 @@ def self.with_completed_preorder_info_count
   end
 
   def is_ordering_for_schools?
-    schools.that_are_centrally_managed.any?
+    schools.gias_status_open.that_are_centrally_managed.any?
```
jamesrose, line 140: Is it worth adding coverage for these?

jamesrose (approved):

benilovj:
I think there's an alternative to this approach, which is to define the `gias_status_open` on the `schools` association, or is that too magic?

asmega:
> I think there's an alternative to this approach, which is to define the gias_status_open on the schools association, or is that too magic?

yea, too magic for me i think i'd put that in the `default_scope` box. I'd rather be slightly more verbose and explicit

app/models/responsible_body.rb
asmega: I've now added some additional coverage

PR #839, opened 2020-11-20
Reduce production log level to :info, to reduce the noise

benilovj:


config/environments/production.rb
```diff
@@ -48,7 +48,7 @@
 
   # Use the lowest log level to ensure availability of diagnostic information
   # when problems arise.
-  config.log_level = :debug
+  config.log_level = :info
```
jamesrose, line 51: There are a few `debug` logs that I think it would be worth escalating to an `info` if we do this, in particular the ones that log Computacenter responses.

benilovj: Good call - done in 39b6dd3

jamesrose (approved):
Looks good to me!

PR #842, opened 2020-11-20
Computacenter Support access: fixes and tweaks

benilovj:
### Context

#809 gave support console access to Computacenter users. However this PR:

- didn't include any links to the support console from the CC area
- still linked to/mentioned some functions that CC users don't have access to

- Add navigation and section links from the CC area to the support area
- Hide some functions in the support area if the user doesn't have those authorisations

### Guidance to review

**New nav link:**

image

----

**New 'Support home' section in the Computacenter home screen:**

image

----

**Support home no longer mentions enabling orders for schools to CC users:**

image

image

### DIFF
--- a/app/views/computacenter/home/show.html.erb
+++ b/app/views/computacenter/home/show.html.erb
@@ -28,10 +28,12 @@
 
     <p class="govuk-body">View details of schools with that can place orders.</p>
 
-    <h2 class="govuk-heading-m">
-      <%= govuk_link_to 'Support home', support_home_path %>
-    </h2>
+    <% if policy(:support).readable? %>
+      <h2 class="govuk-heading-m">
+        <%= govuk_link_to 'Support home', support_home_path %>
+      </h2>
 
-    <p class="govuk-body">Use this section to browse responsible bodies, find users and find schools.</p>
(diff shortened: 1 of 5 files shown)


app/views/computacenter/home/show.html.erb
```diff
@@ -27,5 +27,11 @@
     </h2>
 
     <p class="govuk-body">View details of schools with that can place orders.</p>
+
+    <h2 class="govuk-heading-m">
+      <%= govuk_link_to 'Support home', support_home_path %>
+    </h2>
```
jamesrose, line 33: Is it worth adding logic to hide these links if a CC user did not have read access? I'd be happy to leave this to your judgement - it might be worth waiting until there's a need. 

```suggestion
    <% if policy(:support).readable? %>
      <h2 class="govuk-heading-m">
        <%= govuk_link_to 'Support home', support_home_path %>
      </h2>
    <% end %>
```

jamesrose (approved):

app/views/computacenter/home/show.html.erb
benilovj: Sure, done in a9a848f. I may do a follow-up PR to add authorisation/pundit into the CC area... Ideally we'd be able to push it down to the `ApplicationController` at some point.

PR #843, opened 2020-11-20
add allocation to virtual cap pool

tonyheadford:
### Context
Add `:allocation` to virtual cap pool. This could help smooth out any potential issues with validations and devices_ordered appearing as being greater than the allocation at school level.

Add allocation to `VirtualCapPool`
Provide `allocation` accessor for `SchoolDeviceAllocation` that determines which value to use (pool/local) and is FeatureFlag-ged with `:virtual_caps`
Added unique constraint to ensure there is only one virtual cap pool per `device_type` for a responsible body (as suggested by @aldavidson in the previous PR).
Added an additional check so that a school must also be centrally managed to be added to a pool

### DIFF
--- a/app/models/school_device_allocation.rb
+++ b/app/models/school_device_allocation.rb
@@ -67,6 +67,19 @@ class SchoolDeviceAllocation < ApplicationRecord
     self[:devices_ordered]
   end
 
+  def allocation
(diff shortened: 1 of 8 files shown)


aldavidson (approved):
LGTM :+1:

PR #847, opened 2020-11-23
Improvements to the process of adding missing schools

benilovj:
### Context

There is operational documentation on how to add missing schools from the data staging area into the main `schools` table.

Refactor some existing code to simplify/eliminate certain steps in the documentation, as a stepping stone to automating them with service objects.

### Guidance to review

Does it make sense for `DataStage::School` to be aware of `School`, even if it's just an association?

### DIFF
--- a/app/models/data_stage/responsible_body.rb
+++ b/app/models/data_stage/responsible_body.rb
@@ -1,0 +1,16 @@
+class DataStage::ResponsibleBody
+  RB_NAME_MAP = {
+    'Bristol, City of' => 'City of Bristol',
+    'Dorset' => 'Dorset Council',
+    'Herefordshire, County of' => 'Herefordshire',
+    'Kingston upon Hull, City of' => 'Kingston upon Hull',
+  }.freeze
+
+  def self.find_by_name(responsible_body_name)
+    ::ResponsibleBody.find_by(name: translated(responsible_body_name))
+  end
+
+  def self.translated(responsible_body_name)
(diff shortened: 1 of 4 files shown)


tonyheadford (approved):
I'm a bit ambivalent about this. The original intention was that `SchoolUpdateService` would handle the logic and glue to create and update schools automatically, when the complexities and manual bits were more understood it kind of got a bit disjointed e.g .the `next unless school` line was originally where the `create_school` call was (hence it being a private method), so probably made more sense then. It's not an easy thing to refactor cleanly, it feels a bit icky to have the `RB_NAME_MAP` in the `DataStage::School` model to me, but then it does make some parts flow better how you've broken it up.

benilovj:
> I'm a bit ambivalent about this. The original intention was that `SchoolUpdateService` would handle the logic and glue to create and update schools automatically, when the complexities and manual bits were more understood it kind of got a bit disjointed e.g .the `next unless school` line was originally where the `create_school` call was (hence it being a private method), so probably made more sense then. It's not an easy thing to refactor cleanly,

> it feels a bit icky to have the `RB_NAME_MAP` in the `DataStage::School` model to me, but then it does make some parts flow better how you've broken it up.

@tonyheadford I see your point. I've now extracted a `DataStage::ResponsibleBody` class where I've put that logic into, which hopefully encapsulates it a bit better. I'm always trying to have thinner service classes and fuller models, hence why I think it's worth pulling out of `SchoolUpdateService`.

PR #868, opened 2020-11-27
[1094] re-do schools performance data endpoint in raw SQL, for huge speedup

aldavidson:
### Context

[Trello card 1094](https://trello.com/c/IxhghLAA/1094-performance-data-api-now-takes-7min-to-render-and-download) - The `/support/support/performance-data/schools.json` endpoint has become unsustainably slow since the release of the [virtual cap pool work](https://github.com/DFE-Digital/get-help-with-tech/pull/843) and [expanding the endpoint's dataset to also include comms devices](https://github.com/DFE-Digital/get-help-with-tech/pull/860). It's now taking almost 7 minutes to render and stream the response on production. 

* re-implement the endpoint in raw SQL, to optimise query performance
* use `connection.select_all` to return the records. `select_all` returns results from the database as an array of Hashes, thus removing the need to instantiate ActiveRecord objects.
 
### Guidance to review

Results from staging, where the number of `SchoolDeviceAllocation` records is comparable to prod:

```
> SchoolDeviceAllocation.where('cap > 0 OR allocation > 0').group('device_type').count
=> {"std_device"=>20008, "coms_device"=>21965}
```

#### Before:
```
time curl -s -H 'Authorization: Bearer (...)' https://get-help-with-tech-staging.london.cloudapps.digital/support/performance-data/schools.json
real	1m47.018s
```

#### After:

```
time curl -s -H 'Authorization: Bearer (...)' https://get-help-with-tech-staging.london.cloudapps.digital/support/performance-data/schools.json
real	0m4.204s
```

### DIFF
--- a/app/controllers/support/performance_data/schools_controller.rb
+++ b/app/controllers/support/performance_data/schools_controller.rb
@@ -5,9 +5,9 @@ class Support::PerformanceData::SchoolsController < Support::PerformanceDataCont
     response.headers['Content-Type'] = 'application/json'
     response.stream.write '['
 
-    school_query.find_each.with_index do |school, index|
+    School.connection.select_all(school_query).each_with_index do |row, index|
       response.stream.write ",\n" unless index.zero?
-      response.stream.write attrs_for(school).to_json
+      response.stream.write row.to_json
     end
 
     response.stream.write ']'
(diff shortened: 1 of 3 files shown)


tonyheadford (approved):
Great job!

PR #873, opened 2020-11-30
assets should be cached

aliuk2012:
### Context
This commit adds cache headers to our assets so that a users browser does
not have to download the assets on every page refresh when the assets
have not changed.
### Guidance to review

Opening the preview app and looking at the network tab all our assets should have a cache-control response header
image


tonyheadford (approved):
LGTM

PR #874, opened 2020-11-30
[1080] Add ability to flag specific RBs to have virtual cap enabled

jamesrose:
### Context

Like the MNO offer, we would like to roll out virtual caps in small batches of RBs so that we limit the risk of the rollout. In addition to the global virtual cap feature flag we're adding a per-RB flag.

1. By default, all trusts are excluded from the virtual cap rollout, even if the global feature flag is enabled.
2. Only use the `SchoolVirtualCap` and `VirtualCapPool` is the global feature is on and the individual RB flag is enabled.

### Guidance to review

1. Find school
2. Toggle global virtual cap feature on: no change to cap/allocation/device_ordered
3. Toggle `vcap_feature_flag` enabled on its RB: virtual cap enabled for all the schools within that RB that are eligible

### DIFF
--- a/app/models/responsible_body.rb
+++ b/app/models/responsible_body.rb
@@ -75,6 +75,10 @@ class ResponsibleBody < ApplicationRecord
     end
   end
 
+  def has_virtual_cap_feature_flags?
+    FeatureFlag.active?(:virtual_caps) && vcap_feature_flag?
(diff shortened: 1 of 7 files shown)


tonyheadford (approved):
Looks great!

PR #876, opened 2020-12-01
1037 build an interface for cc to ship-to and sold-to numbers

tonyheadford:
### Context
[Trello card](https://trello.com/c/ds5m99jv/1037-build-an-interface-for-cc-to-shipto-soldto-numbers)
Add an interface to expose changes to schools and responsible bodies to Computacenter and enable them to adjust the `Sold To` and `Ship To` references for responsible bodies and schools as appropriate.  This will then remove the need to capture this information and send CSV files to CC when changes are made.

Add `computacenter_change` column to `School` and `ResponsibleBody` and enum to models reflect change state.
Add "Changes to schools" page that lists changes and form to verify `Ship To` references for each school in the list.
Add CSV download of school changes
Add "Changes to responsible bodies" page that lists changes and form to verify `Sold To` references for each responsible body in the list.
Add CSV download of responsible body changes
Add links to pages on home page of Computacenter area.

### Guidance to review
The new `computacenter_change` enum is used to determine which schools or responsible bodies are included in the lists for Computacenter.  When a change is made to a school that needs to be propagated to Computacenter the `computacenter_change` should be updated accordingly.

Currently we expose new and amended schools and responsible bodies to CC via a CSV. Creating a new school or responsible body will set the `computacenter_change` to `new` so these will be automatically included in the list.

If a change is made to a school or responsible body that CC must be made aware of then the `computacenter_change` should be set to `amended`.  Being an enum there are built in helpers for these i.e. `school.computacenter_change_amended!`.

We currently only notify CC of changes to the organisation name or address, or if a school moves to a different trust (and does not change URN)

Once a school or responsible body has been verified by CC and any necessary changes made, the `computacenter_update` is set to `none` and will no longer appear in the list.

image
image
image
image
image
image
image
image

Responsible Bodies

image
image
image
image
image
image

### DIFF
--- a/app/components/computacenter/school_changes_details_component.rb
+++ b/app/components/computacenter/school_changes_details_component.rb
@@ -30,7 +30,7 @@ private
       value: school.urn,
     },
      {
-       key: 'School Name',
+       key: 'School name',
        value: school.name,
      },
      {
--- a/app/components/computacenter/school_changes_details_component.rb
+++ b/app/components/computacenter/school_changes_details_component.rb
@@ -1,0 +1,48 @@
+class Computacenter::SchoolChangesDetailsComponent < ViewComponent::Base
+  attr_accessor :school
+
+  def initialize(school:)
+    @school = school
+  end
+
+  def rows
+    array = rb_rows.concat(school_rows)
+    array.concat(ship_to_row) if school.computacenter_reference.present?
+    array
+  end
+
+private
+
+  def rb_rows
+    [{
+      key: 'RB URN',
+      value: school.responsible_body.computacenter_identifier,
+    },
+     {
+       key: 'Sold To',
+       value: school.responsible_body.computacenter_reference,
+     }]
+  end
+
+  def school_rows
+    [{
+      key: 'School URN',
+      value: school.urn,
+    },
+     {
+       key: 'School Name',
+       value: school.name,
+     },
+     {
+       key: 'Address',
+       value: school.address,
+     }]
+  end
+
+  def ship_to_row
+    [{
+      key: 'Ship To',
+      value: school.computacenter_reference,
+    }]
+  end
+end
--- a/app/controllers/computacenter/responsible_body_changes_controller.rb
+++ b/app/controllers/computacenter/responsible_body_changes_controller.rb
@@ -8,7 +8,7 @@ class Computacenter::ResponsibleBodyChangesController < Computacenter::BaseContr
     respond_to do |format|
       format.html do
         @responsible_bodies = fetch_responsible_bodies
-        @show_download_link = ResponsibleBody.with_changes_relevant_to_computacenter.any?
+        @show_download_link = ResponsibleBody.requiring_a_new_computacenter_reference.any?
       end
       format.csv { send_data csv_generator, filename: "rb-changes-#{Time.zone.now.strftime('%Y%m%d')}.csv" }
     end
@@ -65,11 +65,11 @@ private
   def query_for_view_mode
     case view_mode
     when 'new'
-      new_responsible_bodies
+      ResponsibleBody.requiring_a_new_computacenter_reference.where.not(computacenter_change: :amended)
     when 'amended'
-      amended_responsible_bodies
+      ResponsibleBody.requiring_a_new_computacenter_reference.where(computacenter_change: :amended)
     when 'all'
-      new_responsible_bodies.or(amended_responsible_bodies)
+      ResponsibleBody.requiring_a_new_computacenter_reference
     end
   end
 
--- a/app/controllers/computacenter/responsible_body_changes_controller.rb
+++ b/app/controllers/computacenter/responsible_body_changes_controller.rb
@@ -1,7 +1,10 @@
 class Computacenter::ResponsibleBodyChangesController < Computacenter::BaseController
   before_action :set_responsible_body, except: :index
+  after_action :verify_authorized
+  after_action :verify_policy_scoped, only: :index
 
   def index
+    authorize ResponsibleBody
     respond_to do |format|
       format.html do
         @responsible_bodies = fetch_responsible_bodies
@@ -12,11 +15,13 @@ class Computacenter::ResponsibleBodyChangesController < Computacenter::BaseContr
   end
 
   def edit
+    authorize @responsible_body, :update_computacenter_reference?
     @form = Computacenter::SoldToForm.new(responsible_body: @responsible_body,
                                           sold_to: @responsible_body.computacenter_reference)
   end
 
   def update
+    authorize @responsible_body, :update_computacenter_reference?
     @form = Computacenter::SoldToForm.new(sold_to_params.merge(responsible_body: @responsible_body))
 
     if @form.valid?
@@ -33,6 +38,7 @@ private
 
   def set_responsible_body
     @responsible_body = ResponsibleBody.gias_status_open.find(params[:id])
+    authorize @responsible_body, :show?
   end
 
   def csv_generator
@@ -40,7 +46,8 @@ private
   end
 
   def fetch_responsible_bodies
-    ResponsibleBody.gias_status_open
+    policy_scope(ResponsibleBody)
+      .gias_status_open
       .merge(query_for_view_mode)
       .order(ResponsibleBody.arel_table[:type].asc, ResponsibleBody.arel_table[:name].asc)
   end
@@ -61,7 +68,7 @@ private
       new_responsible_bodies
     when 'amended'
       amended_responsible_bodies
-    else
+    when 'all'
       new_responsible_bodies.or(amended_responsible_bodies)
     end
   end
--- a/app/controllers/computacenter/responsible_body_changes_controller.rb
+++ b/app/controllers/computacenter/responsible_body_changes_controller.rb
@@ -1,0 +1,80 @@
+class Computacenter::ResponsibleBodyChangesController < Computacenter::BaseController
+  before_action :set_responsible_body, except: :index
+
+  def index
+    respond_to do |format|
+      format.html { @responsible_bodies = fetch_responsible_bodies }
+      format.csv { send_data csv_generator, filename: "rb-changes-#{Time.zone.now.strftime('%Y%m%d')}.csv" }
+    end
+  end
+
+  def edit
+    @form = Computacenter::SoldToForm.new(responsible_body: @responsible_body,
+                                          sold_to: @responsible_body.computacenter_reference)
+  end
+
+  def update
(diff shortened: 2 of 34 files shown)


benilovj:
@tonyheadford is it possible to add some screenshots to the PR?

tonyheadford:
> @tonyheadford is it possible to add some screenshots to the PR?

Yes I'm doing that now :)

app/components/computacenter/school_changes_details_component.rb
```diff
@@ -0,0 +1,48 @@
...
+     }]
+  end
+
+  def school_rows
+    [{
+      key: 'School URN',
+      value: school.urn,
+    },
+     {
+       key: 'School Name',
```
benilovj, line 33:
```suggestion
       key: 'School name',
```

app/controllers/computacenter/responsible_body_changes_controller.rb
```diff
@@ -0,0 +1,88 @@
+class Computacenter::ResponsibleBodyChangesController < Computacenter::BaseController
```
benilovj, line 1: It'd be nice to add Pundit authorisation to this controller. It's not strictly necessary because `Computacenter::BaseController` already does authorisation, but adding it now makes the job of moving to Pundit eventually easier.

app/controllers/computacenter/school_changes_controller.rb
```diff
@@ -0,0 +1,89 @@
+class Computacenter::SchoolChangesController < Computacenter::BaseController
```
benilovj, line 1: It'd be nice to add Pundit authorisation to this controller. It's not strictly necessary because `Computacenter::BaseController` already does authorisation, but adding it now makes the job of moving to Pundit eventually easier.

app/controllers/computacenter/responsible_body_changes_controller.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  def sold_to_params
+    params.require(:computacenter_sold_to_form).permit(:sold_to, :change_sold_to)
+  end
+
+  def update_sold_to
+    @responsible_body.update!(computacenter_reference: @form.sold_to, computacenter_change: 'none')
+  end
+
+  def show_download_link?
```
benilovj, line 85: It feels to me like this could better sit as a scope on `ResponsibleBody` (maybe in `Computacenter::ResponsibleBodyUrns::ClassMethods`) (I'm struggling with the most appropriate name so take it as a strawperson):

```ruby
scope :with_changes_relevant_to_computacenter, { gias_status_open.where(computacenter_change: %w[new amended]) } 
```

and then you can:

- do away with this method
- update the controller action to:

    ```
    @show_download_link = ResponsibleBody.with_changes_relevant_to_computacenter.any?
    ```

app/controllers/computacenter/school_changes_controller.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  def ship_to_params
+    params.require(:computacenter_ship_to_form).permit(:ship_to, :change_ship_to)
+  end
+
+  def update_ship_to
+    @school.update!(computacenter_reference: @form.ship_to, computacenter_change: 'none')
+  end
+
+  def show_download_link?
```
benilovj, line 86: Similar thinking to the comment I left on `ResponsibleBody#show_download_link?`

app/form_objects/computacenter/sold_to_form.rb
```diff
@@ -0,0 +1,12 @@
+class Computacenter::SoldToForm
+  include ActiveModel::Model
+
+  attr_accessor :change_sold_to, :sold_to, :responsible_body
+
+  validates :sold_to, numericality: { only_integer: true, message: 'Sold To must be a number' }
```
benilovj, line 6: Would be good to push the validation message into `en.yml` for consistency.

app/form_objects/computacenter/sold_to_form.rb
```diff
@@ -0,0 +1,12 @@
...
+  validates :change_sold_to, inclusion: { in: %w[yes no], message: 'Tell us whether the Sold To number needs to change' }
```
benilovj, line 7: Would be good to push the validation message into `en.yml` for consistency.

app/services/responsible_body_exporter.rb
```diff
@@ -6,30 +6,40 @@ class ResponsibleBodyExporter
...
-        csv << [
-          rb.computacenter_identifier,
-          *build_name_fields_for(rb),
-          rb.address_1,
-          rb.address_2,
-          rb.address_3,
-          rb.town,
-          rb.postcode,
-        ]
+    if filename
```
benilovj, line 14: Will we still need the filename generation after this change?

app/services/school_data_exporter.rb
```diff
@@ -6,30 +6,41 @@ class SchoolDataExporter
...
-        csv << [
-          school.responsible_body.computacenter_identifier,
-          *build_name_fields_for(school),
-          school.address_1,
-          school.address_2,
-          school.address_3,
-          school.town,
-          school.postcode,
-        ]
+    if filename
```
benilovj, line 14: Do we still need the filename generation after this change?

app/views/computacenter/responsible_body_changes/index.html.erb
```diff
@@ -0,0 +1,74 @@
...
+          path: computacenter_responsible_body_changes_path(view: 'amended'),
+          selected: @view_mode == 'amended')
+        %>
+      </ul>
+      <div class="govuk-tabs__panel">
+        <table class="govuk-table" id="responsible-body-changes">
+          <thead class="govuk-table__head">
+            <tr class="govuk-table__row">
+              <th scope="col" class="govuk-table__header">RB URN</th>
+              <th scope="col" class="govuk-table__header">RB Name</th>
```
benilovj, line 48:
```suggestion
              <th scope="col" class="govuk-table__header">RB name</th>
```

app/controllers/computacenter/school_changes_controller.rb
```diff
@@ -0,0 +1,89 @@
...
+    SchoolDataExporter.new.export_schools(fetch_schools)
+  end
+
+  def fetch_schools
+    School.joins(:responsible_body)
+      .includes(:responsible_body)
+      .gias_status_open
+      .merge(query_for_view_mode)
+      .order(ResponsibleBody.arel_table[:type].asc, ResponsibleBody.arel_table[:name].asc, School.arel_table[:name].asc)
+  end
```
benilovj, line 47: Is there a way to push this method (and maybe `query_for_view_mode`)  onto the `School` model? It feels too domain-logicy to be on the controller...

benilovj (approved):
Looks good! A few non-blocking, minor things.

benilovj:
Another 2 things, probably a question for @fofr 

1. Do we think we definitely need the tabs? I think it'll simplify the code to avoid them, and if CC check this section daily, as I'd expect them to, there shouldn't be too many schools/RB changed that it needs tabbing/splitting, and could just be a straight list instead.

2. Do we need to think about how to flag if any changes have happened (with a badge or counter, say), on the section screen, so CC users know when they need to look?

benilovj:
We will be able to greatly simplify the instructions in `doc/` after this change goes in. It may be worth adding that simplification as part of this PR.

app/services/school_data_exporter.rb
tonyheadford: I guess it depends on how quickly CC transition to use list this and whether we need a fall back?

app/services/responsible_body_exporter.rb
tonyheadford: I guess it depends on how quickly CC transition to use list this and whether we need a fall back?

app/controllers/computacenter/responsible_body_changes_controller.rb
tonyheadford: Nice idea

app/services/school_data_exporter.rb
benilovj: yep fair enough, it's worth scheduling a separate card for cleaning this up shortly after CC transition to the new process.

app/services/responsible_body_exporter.rb
benilovj: yep fair enough, maybe worth scheduling a separate card for cleaning this up shortly after CC transition to the new process.

app/controllers/computacenter/responsible_body_changes_controller.rb
benilovj: Done in 47ca5957

app/controllers/computacenter/school_changes_controller.rb
benilovj: Done in ce7e664f

benilovj:
> Another 2 things, probably for @fofr:

I Slacked with @fofr about this last week...

> 
> 1. Do we think we definitely need the tabs? I think it'll simplify the code to avoid them, and if CC check this section daily, as I'd expect them to, there shouldn't be too many schools/RB changed that it needs tabbing/splitting, and could just be a straight list instead.

We'll leave the tabs as they are right now.

> 2. Do we need to think about how to flag if any changes have happened (with a badge or counter, say), on the section screen, so CC users know when they need to look?

Paul's suggestion was to include the count on the CC homepage:

e.g. 

- Changes to schools (4)
- Changes to responsible bodies (3)

I've now added this design change to the PR (see 065ac7b04498646111e306c694dc52a1ca75a256 and 8bd8a603cbbeed63f29dc0d368ff6450ca8aef97).

benilovj:
I've now done as much as I was planning to do on this. This includes:

* actioning the suggestions I made as part of my PR review (e.g. adding Pundit authorisation, pushing code out of controllers into the models, adding counts to the user interface)
* unifying the `School` and `ResponsibleBody` scopes, in order to eliminate subtle bugs between the HTML and CSV functionality (2eaf39489640d7d1068f85f4ab388c8403d4b0f4)

In order to accomplish these, I made some changes that were arguably stand-alone but are included as part of this PR:

* pushing Pundit integration into `ApplicationController` so Pundit can be used beyond just the support interface (1b7a9f1b86a26a657f6180c4ba21bd5238e91433)
* fixing an issue with the `:school` Factory Bot factory which was causing duplicate schools to be created when associations were specified (bc326f19e77f546677636bdeb7d5755141a35e9f)

@tonyheadford @jamesrose I appreciate you've got other things going on, but I've made large enough updates that someone should re-review this before merging.

tonyheadford (commented):
Looks great!

PR #877, opened 2020-12-01
[1083] replace devices ordered panel with 'Order devices now' link when VCaps are enabled

aldavidson:
### Context

[Trello card 1083](https://trello.com/c/2mIHtMV8/1083-virtual-cap-update-school-page) - as part of the [Virtual Cap Pools epic](https://trello.com/c/Blj9SZ65/1007-virtual-cap), we need to update the Responsible Body school page according to the [updated design](https://ghwt-design-history.herokuapp.com/rbs-ordering-for-groups-of-schools/#simplified-school-page).

* When the `virtual_caps` feature flag is enabled AND the current responsible body has the `vcap_feature_flag` attribute set to true, remove the panel which would normally show how many devices have been ordered from their cap.

This is because we cannot accurately know the numbers per school if the school is part of a pool, so we need to defer to a central place for ordering.  

### Guidance to review

* set the `:virtual_caps` feature flag to `'active'`
* choose a Responsible Body with some schools, and set `vcap_feature_flag: true` on it
* log in as a user from that Responsible Body, go to 'Get laptops and tables' > 'List of schools' and click on a school

### DIFF
--- a/app/views/responsible_body/devices/schools/show.html.erb
+++ b/app/views/responsible_body/devices/schools/show.html.erb
@@ -7,23 +7,29 @@
       <%= @school.name %>
     </h1>
 
-    <%- if @school.has_std_device_allocation? %>
-      <%= render DeviceCountComponent.new(
-        school: @school,
-        show_action: @school.std_device_allocation.devices_ordered < @school.std_device_allocation.cap,
-        action: { 'Order devices' => order_devices_responsible_body_devices_school_path(urn: @school.urn) }) %>
-    <% else %>
-      <%= render partial: 'shared/school_without_allocation' %>
-    <%- end %>
+    <%- if FeatureFlag.active?(:virtual_caps) && @responsible_body.vcap_feature_flag %>
+      <%- if @school.can_order? || @school.can_order_for_specific_circumstances? %>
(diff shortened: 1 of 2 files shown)


app/views/responsible_body/devices/schools/show.html.erb
```diff
@@ -7,23 +7,29 @@
...
 
-    <%- if @school.has_std_device_allocation? %>
-      <%= render DeviceCountComponent.new(
-        school: @school,
-        show_action: @school.std_device_allocation.devices_ordered < @school.std_device_allocation.cap,
-        action: { 'Order devices' => order_devices_responsible_body_devices_school_path(urn: @school.urn) }) %>
-    <% else %>
-      <%= render partial: 'shared/school_without_allocation' %>
-    <%- end %>
+    <%- if FeatureFlag.active?(:virtual_caps) && @responsible_body.vcap_feature_flag %>
```
jamesrose, line 10: I added `ResponsibleBody#has_virtual_cap_feature_flags?` that does this late yesterday. Sorry, should have let everyone know.

jamesrose (approved):

app/views/responsible_body/devices/schools/show.html.erb
aldavidson: done :+1:

PR #882, opened 2020-12-02
Create preparing_ipads.md

gemmaemoore:


aldavidson:
Combined with #883 into #885

PR #883, opened 2020-12-02
Added iPad guidance

gemmaemoore:


fofr:
Is this missing the iPad guidance – I only see the title – or is that already included elsewhere?

gemmaemoore:
Hi Paul, PR #882 has the new content. Previously I added the descriptions
in here so they display on the /devices page and I created the markdown
files for each of the guides. Jake was probably doing other magic under the
hood though!

aldavidson:
Combined with #882 into #885

PR #884, opened 2020-12-02
Send different invite email when a RB ordering centrally is using virtual caps

jamesrose:
### Context

Trello: https://trello.com/c/HyeLDn5U/1102-virtual-caps-update-automatic-email-which-is-sent-when-a-trust-ordering-centrally-is-invited-to-order-when-a-school-reports-disr

> We currently send an automated email out when we raise the cap for a school having reported disruption. This sends a generic message inviting them to order which is the same for both schools ordering directly and trusts ordering on behalf of schools. 

> When we implement virtual caps we will no longer know if that school has received their allocation of devices only how many are remaining within the trusts 'virtual cap' as a result we need to adapt our communications for trusts ordering centrally to reflect this. 

Use a different email template when the RB of a school ready to order is ordering centrally _and_ in the virtual cap feature.

### Guidance to review

Is a preorder status of `rb_can_order` enough to qualify that it is 1) centrally managed, and 2) ready to order - even after placing an initial order?

1. Enable global virtual caps feature flag 
2. Find RB that is ordering centrally and toggle the feature flag on
3. Open school for ordering
4. Email should be sent

### DIFF
--- a/app/models/school_can_order_devices_notifications.rb
+++ b/app/models/school_can_order_devices_notifications.rb
@@ -48,6 +48,8 @@ private
       :user_can_order_but_action_needed
     elsif status?('rb_can_order', 'school_can_order', school: school) && user.orders_devices? && !user.seen_privacy_notice?
       :nudge_user_to_read_privacy_policy
+    elsif status?('rb_can_order', school: school) && school.responsible_body.has_virtual_cap_feature_flags? && user.in?(school.order_users_with_active_techsource_accounts)
+      :user_can_order_in_virtual_cap
     elsif status?('ready', 'school_ready', 'rb_can_order', 'school_can_order', school: school) && user.in?(school.order_users_with_active_techsource_accounts)
       :user_can_order
     end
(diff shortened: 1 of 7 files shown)


app/models/school_can_order_devices_notifications.rb
```diff
@@ -48,6 +48,8 @@ def what_message_to_send(school, user)
       :user_can_order_but_action_needed
     elsif status?('rb_can_order', 'school_can_order', school: school) && user.orders_devices? && !user.seen_privacy_notice?
       :nudge_user_to_read_privacy_policy
+    elsif status?('rb_can_order', school: school) && school.responsible_body.has_virtual_cap_feature_flags? && user.in?(school.order_users_with_active_techsource_accounts)
```
jamesrose, line 51: Is a preorder status of rb_can_order enough to qualify that it is 1) centrally managed, and 2) ready to order - even after placing an initial order?

asmega: Yes i think that is sufficient. It is derived from `PreorderInformation#infer_status`. 
- It checks the `PreorderInformation` on who is ordering
- Even after an order if they can still order it looks like it will stay as `rb_can_order`.

asmega (approved):
LGTM

PR #887, opened 2020-12-02
[1064] Add validation to school_or_rb_domain

asmega:
### Context

https://trello.com/c/Ea9OavT1/1064-improve-google-schoolorrbdomain-data-collection

- add validation to `PreorderInformation#school_or_rb_domain`
- This should only run on create otherwise existing invalid records cannot be be saved/updated
- in the future if we ever get all records passing validation we can remove this from `on: :create` 

### Guidance to review

- Sign up and follow process
- on prompting for chromebook details do this will assorted domains that are not valid
- Another test, create record with invalid `school_or_rb_domain`
- Update another attribute on this record, should be saved as validation should not run

### DIFF
--- a/app/models/preorder_information.rb
+++ b/app/models/preorder_information.rb
@@ -7,6 +7,12 @@ class PreorderInformation < ApplicationRecord
   belongs_to :school_contact, optional: true
 
   validates :status, presence: true
+  validates :school_or_rb_domain,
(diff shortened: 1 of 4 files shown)


asmega:
So we already have validation by asking google if domains are valid so looks like this work is not needed

PR #890, opened 2020-12-03
[1122] Fix intermittent duplicate-key spec failures

aldavidson:
### Context

While investigating [intermittent ActiveRecord::RecordNotUnique errors in our test suite](https://trello.com/c/KIkSSDVf/1122-fix-intermittent-duplicate-key-spec-failures), I isolated the problem down to test data leaking, specifically two examples which created their data in `around` hooks. 

This is outside of the per-test transaction scope, so DatabaseCleaner's `:transaction` strategy is unable to roll it back, and the record persists into all subsequent examples.

This can occasionally cause a unique constraint violation even on values created with Faker::(type).unique, as Faker does not
use the database for existence checking.

* restructure the two `around` hooks as `before`/`after` pairs

### DIFF
--- a/spec/services/responsible_body_exporter_spec.rb
+++ b/spec/services/responsible_body_exporter_spec.rb
@@ -33,7 +33,7 @@ RSpec.describe ResponsibleBodyExporter, type: :model do
       sat
       exporter.export_responsible_bodies
     end
-    
+
     after do
       remove_file(filename)
     end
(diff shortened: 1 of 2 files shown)


asmega (approved):

asmega:
I'm slightly surprised and annoyed by this. One would think earlier `around(:each)` blocks would be on the wrap later ones

aldavidson:
> I'm slightly surprised and annoyed by this. One would think earlier `around(:each)` blocks would wrap later ones

I think it's unclear either way - uou could also make an argument that each subsequent `around` block wraps around any previous ones....

PR #891, opened 2020-12-03
[1107] Notify MNOs if there are new requests

asmega:
### Context

- https://trello.com/c/TprR7Aph/1107-build-automatic-notification-for-mnos-when-they-receive-a-request

- Notify MNO users if they have new requests
- Emails are sent every weekday at 8am

### Screenshots

image

### Guidance to review

- Ensure there is an MobileNetwork with a user
- Create a request for this mno
- Trigger job
- User should see email

### DIFF
--- a/app/jobs/notify_mnos_job.rb
+++ b/app/jobs/notify_mnos_job.rb
@@ -2,14 +2,10 @@ class NotifyMnosJob < ApplicationJob
   queue_as :default
 
   def perform
-    MobileNetwork
-      .joins(:extra_mobile_data_requests)
-      .where(extra_mobile_data_requests: { status: %w[requested] })
-      .where('extra_mobile_data_requests.created_at > ?', since_last_email)
-      .having('COUNT(mobile_networks.id) > 0')
-      .group('mobile_networks.id').each do |mno|
-      number_of_new_requests = mno.extra_mobile_data_requests.requested.where('created_at > ?', since_last_email).count
-      mno.users.each do |user|
+    users.each do |user|
+      number_of_new_requests = user.mobile_network.extra_mobile_data_requests.requested.where('created_at > ?', since_last_email_for_user(user)).count
+
+      if number_of_new_requests.positive?
         MnoMailer.notify_new_requests(user: user, number_of_new_requests: number_of_new_requests).deliver_now
       end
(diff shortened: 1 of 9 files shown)


fofr:
If an MNO has a new request at 10am on 2 Dec, will they get an email about it at 8am on 3 Dec?

If they then do not process that request, will they get another email about the same request at 8am on 4 Dec?

asmega:
I've now updated this so we only email if there have been new requests since the last time we emailed them. It's a bit hard coded but it should do the job and I don't expect this to change drastically.

app/jobs/notify_mnos_job.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def since_last_email
+    now = Time.zone.now
+
+    if now.monday?
+      now - 72.hours
+    else
+      now - 24.hours
+    end
+  end
```
aldavidson, line 28: What about using the `EmailAudit` record to record when the last mail of this type was sent to that user, and then querying that log here? It should be more robust in the case of jobs not running at the expected time for whatever reason, or schedule changes etc

tonyheadford (approved):
Agree with @aldavidson comment that the `since_last_email` could be a bit more robust but don't want to block this.

app/jobs/notify_mnos_job.rb
asmega: - I've added this in the last commit.
- I felt like i have rushed this a little so could do with a fine comb.
- It won't be as efficient as the previous implementation due to extra queries it'll have to perform.
- There are some drawbacks to doing it this way:
  - different users from same network may get different emails which could be misleading
  - new users will never have had an email sent this is hard coded to 72 hours ago if this is the case

aldavidson: This code looks fine to me :+1: 
I wouldn't worry about the extra queries too much - the job runs once a day in the background at 5am,  so it won't impact much.
It's a fair point about different users getting different emails, but I think that's better than the previous situation. It's likely that the number of requests will have changed in between the email being sent and the user clicking on the interface anyway. 
A default timespan of 72hrs seems fine to me.

aldavidson (approved):

PR #893, opened 2020-12-04
1131 add debug build output to diagnose intermittent build error on assetpackpath

aldavidson:
### Context

[Trello card #1131](https://trello.com/c/zTfOX6I9/1131-fix-intermittent-build-error-on-assetpackpath) - We're still getting intermittent build errors, of the form:

```
 Failure/Error: <%= tag :meta, property: 'og:image', content: asset_pack_path('media/images/govuk-opengraph-image.png') %>

     ActionView::Template::Error:
       784: unexpected token at ''
     # ./app/views/layouts/application.html.erb:13:in `_app_views_layouts_application_html_erb___776826877070931037_107660'
(...)
     # ------------------
     # --- Caused by: ---
     # JSON::ParserError:
     #   784: unexpected token at ''
     #   ./app/views/layouts/application.html.erb:13:in `_app_views_layouts_application_html_erb___776826877070931037_107660'
```

I've been unable to replicate this issue locally, it only seems to occur on Github. So I've added a couple of bits of debug output to try and get more information,  and hopefully diagnose this issue.

1. We're using parallel tests, and the error only seems to occur partway through a run - so I've added a line to `rails_helper.rb` that will help us identify the exact per-process run order when a problem occurs (only if `DEBUG_SPEC_RUN_ORDER` is set, which it will be on Github)
2. As `asset_pack_path` ultimately looks up asset paths in `manifest.json`, that seems like a good candidate for the source of the JSON::ParserError. When an error occurs, it will be hugely helpful to know exactly what's in that file - so I've added a build step to `cat` the contents of the manifest to STDOUT.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -90,6 +90,8 @@ jobs:
 
       - name: Compile assets
         run: bundle exec rails webpacker:compile
+      - name: Dump out manifest contents in case of JSON::ParserError
+        run: cat public/packs/manifest.json
       - name: Run rubocop
         run: bundle exec rubocop app config db lib spec Gemfile --format clang --parallel
       - name: Run rspec
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -98,6 +98,7 @@ jobs:
         run: bundle exec rake parallel:spec
         env:
           RAILS_ENV: test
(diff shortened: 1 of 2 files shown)


jamesrose (approved):
Nice!

PR #895, opened 2020-12-04
Hide 'Who will order?' change link to RBs using virtual caps

jamesrose:
### Context

Trello: https://trello.com/c/x4uZNNZ0/1042-switch-changing-who-can-order-devices-at-the-school-level-to-a-support-task

It is currently a developer task to back out of using virtual caps. We need to at least hide the link in the UI that allows RBs to change who will order when it is enabled.

Hide 'Change' link for RBs using virtual caps

### Guidance to review

1. Find RB using virtual caps and managing centrally
2. Navigate to school page
3. You should not see the 'Change' link for 'Who can order?'

---

1. Find RB with devolved ordering
2. Navigate to school page
3. You should see the 'Change' link for 'Who can order?'

### DIFF
--- a/app/components/responsible_body/school_details_summary_list_component.rb
+++ b/app/components/responsible_body/school_details_summary_list_component.rb
@@ -39,19 +39,25 @@ private
   end
 
   def who_will_order_row
-    who = (@school.preorder_information || @school.responsible_body).who_will_order_devices_label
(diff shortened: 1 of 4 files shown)


aldavidson (approved):
LGTM :+1:

PR #896, opened 2020-12-04
1108 virtual cap update your schools page

tonyheadford:
### Context
[Trello card](https://trello.com/c/Gvsyoype/1108-virtual-cap-update-your-schools-page)
Update list of schools page for responsible body to support virtual caps.

Updated school list page to new content, 3 separate lists to split schools by order state.
Only show statuses for schools that are not ready to order or additionally, that have ordered in the case of devolved schools
Available devices pooled summary card only visible when responsible body orders centrally and feature flags for virtual caps enabled.

### Guidance to review
As a RB user, where the RB manages centrally, with a mix of schools that can order/can_order_for_specific_circumstances and cannot order. From the Responsible Body home page visit the Get laptops and tablets -> Your Schools links. You should then see a list of schools divided into 3 tables.  There will be no available devices summary card at the top of the page.
image

Enable the feature flags for virtual caps (`FEATURES_virtual_caps=active bundle exec rails s`) and `responsible_body.udpate!(vcap_feature_flag: true)`
Repeat the journey, the summary card should now be visible at the top of the list.
image

In both cases, the school allocation figures displayed use the 'raw' original value from the `SchoolDeviceAllocation` not the value from the virtual pool.

### DIFF
--- a/app/components/conditional_school_preorder_status_tag_component.html.erb
+++ b/app/components/conditional_school_preorder_status_tag_component.html.erb
@@ -1,0 +1,1 @@
+<%= render GovukComponent::Tag.new(text: text, colour: type.to_s) if should_show_status? %>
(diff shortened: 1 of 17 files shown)


jamesrose (approved):
LGTM - nice work!

PR #899, opened 2020-12-07
Updates delivery message

ralph-hawkins:
### Context
https://trello.com/c/oBmUnDoy/1101-update-delivery-content-to-reflect-new-approach-to-large-orders

Computacenter is starting to process large orders and have difficulty doing this within the 48 hours SLA and have experienced issues with schools accepting such large deliveries at short notice. As a result, they are adopting a new approach to fulfilling large orders. 

From the 7th December, where an order is placed on TechSource for 150 or more devices Computacenter will:
- Telephone the delivery contact to arrange delivery via a specialist courier
- Confirm via a ZenDesk email the date of the delivery, we will also explain that the courier link in the dispatch email will not work.

Everywhere delivery timeline is mentioned it's updated from '5 working days'  to '2 working days'. On the overview guidance and on the order page (where the user has more than 149 devices available to them) we tell the user 'For large orders (more than 150 devices), we’ll call the main delivery contact to arrange delivery'. 

Updates to the RB ordering page will be in a separate PR.

### DIFF
--- a/app/views/school/devices/can_order.html.erb
+++ b/app/views/school/devices/can_order.html.erb
@@ -14,7 +14,7 @@
 
     <p class="govuk-body">Order your devices now. Do not delay.</p>
     <p class="govuk-body">Most devices will be delivered within 2 working days.</p>
-    <% if @school.std_device_allocation.available_devices_count > 149 %>
+    <% if @school&.std_device_allocation&.available_devices_count > 149 %>
       <p class="govuk-body">For large orders (more than 150 devices), we’ll call your delivery contact to arrange delivery.</p>
     <% end %>
     <p class="govuk-body">Ordering will close when your school reopens.</p>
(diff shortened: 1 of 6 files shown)


app/views/school/devices/can_order.html.erb
```diff
@@ -12,9 +12,12 @@
 
     <%= render DeviceCountComponent.new(school: @school) %>
 
-    <p class="govuk-body">Only order devices for children who cannot attend school and do not have access to a laptop or tablet.</p>
-    <p class="govuk-body">We’re working hard to get devices delivered within 5 working days.</p>
-    <p class="govuk-body">Ordering will close when your school reopens.</p>
+    <p class='govuk-body'>Order your devices now. Do not delay.</p>
+    <p class='govuk-body'>Most devices will be delivered within 2 working days.</p>
+    <% if @school.std_device_allocation.available_devices_count > 149 %>
```
fofr, line 17: Do we need a test for this?

We might also want to use the safe navigation operator in case the school doesn't have an allocation:
```suggestion
    <% if @school&.std_device_allocation&.available_devices_count > 149 %>
```

fofr (approved):

PR #904, opened 2020-12-08
[1077] Extend support user search to include users with no associated schools or RBs

aldavidson:
### Context

[Trello card 1077](https://trello.com/c/Bs8B6yIR/1077-support-find-rb-or-school-users-doesnt-return-users-with-no-associations) - At the moment it's possible for support users to remove all associated schools or responsible bodies from a user, but then be unable to find them again in order to re-associate them. 

Remove the restriction on the user search which limited it to only users already associated with a school or responsible body.
 
### Guidance to review

Support home > Find responsible body or school users 
Fill in a search term - there is no minimum length, so you can submit something like '@'. 

The search should:
* return the first 100 users matching the term, whether or not they have a responsible body / school
* respond quickly, even when searching the whole database (I've tested this locally against 33k users)
* indicate when users have no associated organisations (see screenshot)

Screenshot from 2020-12-08 14-01-00

### DIFF
--- a/app/components/support/user_search_result_component.html.erb
+++ b/app/components/support/user_search_result_component.html.erb
@@ -6,6 +6,6 @@
   component.slot(:row, key: 'Email address', value: user.email_address)
   component.slot(:row,
                   key: 'Associated with',
-                  value: unordered_list_of_orgs,
+                  value: organisations,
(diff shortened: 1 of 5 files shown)


jamesrose (approved):
Awesome!

PR #907, opened 2020-12-08
Remove reduced allocation features and flag

fofr:
We no longer need this messaging.


benilovj:
Do you need to remove these also?:

- https://github.com/DFE-Digital/get-help-with-tech/blob/main/config/routes.rb#L69
- https://github.com/DFE-Digital/get-help-with-tech/blob/main/app/views/school/devices/changed_allocation.html.erb
- https://github.com/DFE-Digital/get-help-with-tech/blob/main/app/views/responsible_body/devices/home/changed_allocations.html.erb

fofr:
I've removed the extra files now.

jamesrose (approved):
🚢

PR #927, opened 2020-12-10
1097 virtual cap force handle multiple chromebook domains

tonyheadford:
### Context
[Trello card](https://trello.com/c/JFczhXmY/1097-virtual-cap-force-the-orders-of-trusts-that-qualify-for-virtual-cap-and-have-multiple-chromebook-domains-into-the-techsource-man)
Handling of ordering of Chromebooks where multiple G Suite domains exist for responsible bodies managing centrally in virtual pools.  In these circumstances. when a responsible body places an order at TechSource, the order needs to be forced into a manual process in order to resolve which domain should be used for Chromebooks.

Provide an interface for CC to view (and download as CSV) which responsible bodies that manage centrally and have multiple chromebook domains.
Add additional copy to the "Order devices" page to inform the users that TechSource will be in touch.
During sending cap updates to CC, if the responsible body manages centrally and multiple Chromebook domains exist then the cap amounts are adjusted to the `devices_ordered` amount (i.e. nothing can be ordered) in the API call payload as a means to force any orders into the TechSource manual queue.

### Guidance to review
__Set up__:
* Run the app or console with virtual cap feature flag
```
FEATURES_virtual_caps=active bundle exec rails s
```

* Make one or more responsible bodies manage centrally and be enable for virtual caps.

```ruby
responsible_body.update!(who_will_order_devices: 'responsible_body', vcap_feature_flag: true)
```

* Configure some of the RBs schools to want chromebooks and have different chromebook domains. Do this from the console so as to avoid validation of the domain. Also set the recovery address email.

```ruby
school.preorder_information.update!(who_will_order_devices: 'responsible_body', will_need_chromebooks: 'yes', school_or_rb_domain: 'mydomain.com', recovery_email_address: 'user@email.com')
```

* Ensure the schools have `SchoolDeviceAllocation`s with some allocation and cap values etc.

* Add the schools to the RB's virtual pool.

```ruby
responsible_body.add_school_to_virtual_cap_pools!(school)
```

__Method__
Log into the service as a user for the responsible body. Navigate to Home -> Get laptops and tables -> Order devices and the extra __Chromebook__ section should be visible.
image

----

Log in as a Computacenter user
The Home page should have an extra option __Responsible bodies managing multiple Chromebook domains__
image

----

Follow the __Responsible bodies managing multiple Chromebook domains__ link and you should see a table of responsible bodies with multiple Chromebook domains and be able to see the one setup earlier in the list.

image

----

__Checking API Payload__
You may need to set the `GHWT__COMPUTACENTER__OUTGOING_API__ENDPOINT` env var to point to a dummy service to see the cap updates being generated.  Copy the following into a file and run it from the command line and set the env var to point to `http://localhost:8000`:

```ruby
require 'webrick'
​
server = WEBrick::HTTPServer.new Port: 8000
​
trap 'INT' do server.shutdown end
​
server.mount_proc '/' do |req, res|
  res.body = '<CapAdjustmentResponse dateTime="2020-09-14T21:55:37Z" payloadID="11111111-1111-1111-1111-111111111111"><HeaderResult piMessageID="11111111111111111111111111111111" status="Success"/><FailedRecords/></CapAdjustmentResponse>'
end
​
server.start
```

Make changes to the `:cap` or `:devices_ordered` in the `SchoolDeviceAllocation` for one of the schools set up earlier. You should see output either in the rails console or from the WEBrick server showign the payload taht would be sent to Computacenter. Verify that the `capAmount` in the XML matches the `raw_devices_ordered` value for the `SchoolDeviceAllocation`.

### DIFF
--- a/app/controllers/computacenter/multi_domain_chromebooks_controller.rb
+++ b/app/controllers/computacenter/multi_domain_chromebooks_controller.rb
@@ -1,7 +1,6 @@
 require 'csv'
 
 class Computacenter::MultiDomainChromebooksController < Computacenter::BaseController
-
   def index
     @responsible_bodies = ResponsibleBody.managing_multiple_chromebook_domains.order(type: :asc, name: :asc)
     respond_to do |format|
@@ -21,10 +20,10 @@ private
       csv << ['RB Type', 'RB Name', 'RB URN', 'Sold To']
       @responsible_bodies.each do |responsible_body|
         csv << [
-            responsible_body.humanized_type,
-            responsible_body.computacenter_name,
-            responsible_body.computacenter_identifier,
-            responsible_body.computacenter_reference,
+          responsible_body.humanized_type,
+          responsible_body.computacenter_name,
+          responsible_body.computacenter_identifier,
+          responsible_body.computacenter_reference,
         ]
       end
     end
--- a/app/controllers/computacenter/multi_domain_chromebooks_controller.rb
+++ b/app/controllers/computacenter/multi_domain_chromebooks_controller.rb
@@ -1,0 +1,5 @@
+class Computacenter::MultiDomainChromebooksController < Computacenter::BaseController
+  def index
+    @responsible_bodies = ResponsibleBody.managing_multiple_chromebook_domains
+  end
+end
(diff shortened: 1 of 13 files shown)


aldavidson (approved):
This took quite a bit of thinking through to review, never mind code - great stuff :100:

PR #932, opened 2020-12-11
Hide 'Increased allocations' banner at RB level when it has no schools that were increased

jamesrose:
### Context

An extension of https://trello.com/c/EGYSK7OO/1141-add-notification-banner-for-schools-and-rbs-telling-them-that-allocations-have-increased-also-show-no-xmas-ordering-banner-%F0%9F%8E%84-%F0%9F%8E%85

We previously decided to show this generic messaging to RBs regardless of whether they had their allocations increased. This ensures that they have at least one school where allocations were increased.

### Guidance to review

RB level messaging should only show when 1) the global feature flag is enabled, and 2) It has at least one school where we've increased allocations and toggled its flag.

### DIFF
--- a/app/models/responsible_body.rb
+++ b/app/models/responsible_body.rb
@@ -83,6 +83,10 @@ class ResponsibleBody < ApplicationRecord
     end
   end
 
+  def has_increased_allocation_feature_flags?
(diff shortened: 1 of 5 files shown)


tonyheadford (approved):
LGTM

PR #935, opened 2020-12-11
Allow soldTo number to be added to DfE by adding Pundit policy

benilovj:
### Context

#876 introduced an interface for adding sold_to and ship_to numbers to responsible bodies and schools that don't have them. That PR had a bug whereby trying to add a sold_to number to `DfE`, a special responsible body (used for test devices) failed.

Authorise Computacenter users to add a ship_to number to `DfE`.


tonyheadford (approved):
Good catch!

PR #936, opened 2020-12-11
Don't try getting 'Sold To' numbers for RBs without schools

benilovj:
### Context

#876 introduced the functionality to allow CC users to assign sold_to numbers to responsible bodies. We don't want this to be done for RBs which don't manage any schools, and there's quite of few of these in production.

Filter out RBs without schools when identifying 'changed RBs' for CC's benefit.
Also refactor the responsible body FactoryBot factory, to make the local authority and trust factories work much more consistently.


tonyheadford (approved):
LGTM

PR #937, opened 2020-12-12
Responsible body auto-complete in the support console

benilovj:
### Context

It's currently possible to change a user's responsible body within the support console:

- the support agent starts typing part of the responsible body's name into a text field
- they submit the form
- they pick a responsible body from a set that's fuzzy-matched on the support agent's input, on the next screen

This interaction could be done without the full-page refresh and without having an extra page in the codebase.

Introduce an auto-complete for the responsible body. This is a JS-progressive enhancement over a dropdown which contains all the responsible bodies. Much of the code (and the approach) is cribbed from the [Apply for Teacher Training codebase](https://github.com/DFE-Digital/apply-for-teacher-training).

### Guidance to review

responsible-bodies-autocomplete

There's 3000+ responsible bodies, which makes for quite a long dropdown, but this is going to be used by only a small number of support users who are running on modern hardware.


aldavidson (commented):
Love the idea, but I'm getting some issues when trying to run it:

Screenshot from 2020-12-14 08-42-31

I've done the usual `yarn install` / `webpacker:compile` commands, is there something else needed?

aldavidson (approved):
I think this makes sense, I've tested it successfully locally after deleting everything in `public/packs`. That also gave me a chance to test the fallback solution for when the JS doesn't load (the bare select box) which works fine too.

There are ¬3100 RBs in the `gias_status_open` scope, which is pushing it a bit for loading into a select box if you're on a slow connection - but overall it's probably fine given that this solution cuts out a separate page load entirely.

benilovj:
@aliuk2012 mentioned an autocomplete component that he's introduced on `register-trainee-teachers`:

https://github.com/DFE-Digital/register-trainee-teachers/tree/master/app/components/form_components/autocomplete

I will look to bring this over on subsequent PRs, but for the moment this is the only autocomplete so it's not necessary going forward.

PR #941, opened 2020-12-14
Create CapUpdateCall records even when the cap update requests fails

jamesrose:
### Context

Previously we were only creating a `CapUpdateCall` record once a cap update was successfully received. This PR adds a new `failure` column that will toggle on when we receive an error from CC.


tonyheadford (approved):
Looks good!

PR #945, opened 2020-12-14
Allow schools in any order state to be added to virtual pools

tonyheadford:
### Context
[Trello card](https://trello.com/c/CX6BQRF4/1153-allow-schools-that-cannot-order-in-virtual-cap)
Currently, one of the requirements for a school to be added to a virtual cap pool is that they are in an ordering state (`can_order` or `can_order_for_specific_circumstances`). This PR removes this requirement so taht schools in any order state can be added to their responsible body's virtual cap pool.

Remove order state from eligibility criteria

### Guidance to review
Given a responsible body that centrally manages and a centrally managed school that is not in an ordering state, you should be able to add the school to the responsible body's virtual cap pool with `responsible_body.add_school_to_virtual_cap_pool!(school)`

### DIFF
--- a/app/models/virtual_cap_pool.rb
+++ b/app/models/virtual_cap_pool.rb
@@ -45,7 +45,6 @@ private
   def school_can_be_added_to_pool?(school)
     school.responsible_body_id == responsible_body_id &&
       school&.preorder_information&.responsible_body_will_order_devices? &&
-      (school.can_order? || school.can_order_for_specific_circumstances?) &&
       school.device_allocations.exists?(device_type: device_type) &&
       !school_virtual_caps.exists?(school_device_allocation: school.device_allocations.find_by(device_type: device_type))
   end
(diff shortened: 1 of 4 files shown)


app/models/virtual_cap_pool.rb
```diff
@@ -45,7 +45,6 @@ def has_school?(school)
   def school_can_be_added_to_pool?(school)
     school.responsible_body_id == responsible_body_id &&
       school&.preorder_information&.responsible_body_will_order_devices? &&
-      (school.can_order? || school.can_order_for_specific_circumstances?) &&
```
jamesrose, line 48: @tonyheadford I think you'll want to relax the eligibility in `SchoolOrderStateAndCapUpdateService` as well.

tonyheadford: Good spot, thanks!

jamesrose (approved):
Looks good to me! 🚢  Can you look at adding all the open schools on production?

PR #946, opened 2020-12-14
Use raw allocation numbers when implying cap from order state

jamesrose:
### Context

1. When a school within a virtual pool has its state changed, we shouldn't imply the cap on the school allocation using the v-cap devices_ordered or allocation number.
2. Show raw allocation number on allocation edit screen

---

The tests are a duplicate of the previous cap_implied_by_order_state ones except the school is within a virtual cap with 2x the allocation.


tonyheadford (approved):
Looks good!

PR #952, opened 2020-12-15
Don't return deleted users in the support console user search

benilovj:
### Context

The user search in the support console returns deleted users.

Hide deleted users from support console search results.


fofr:
For CC users does the search also exclude anyone who hasn't seen the privacy policy?

benilovj:
> For CC users does the search also exclude anyone who hasn't seen the privacy policy?

Yes, this is already handled in the Pundit authorisation scope:

https://github.com/DFE-Digital/get-help-with-tech/blob/main/app/policies/user_policy.rb#L3-L11

aldavidson (approved):

PR #953, opened 2020-12-15
[1168] spike an easier way of batch processing

aldavidson:
### Context

See [Trello card](https://trello.com/c/QuI2b5kR/1168-spike-an-easier-way-of-batch-processing) for full context.

When running arbitrary ad-hoc code from the console (for instance, to raise allocations for a selection of schools) we often end up re-writing basic code to handle errors, report success & failures, etc. This is wasteful and error-prone, and means that our actual processing code is more difficult to debug.

* Add `BatchJob` and `BatchJobLogEntry` classes.

This is a first iteration to handle the basic error handling and status reporting for running an arbitrary block against a series of records.  

### Guidance to review

Use it as follows:

```
# records must be iterable, but can be anything - ActiveRecord resultsets, Arrays of Hashes, etc.
job = BatchJob.new(job_name: 'some helpful name to identify this job', records: [ ... ]
job.process! do |record|
  # The BatchJob will yield records one at a time
  # Do whatever you like with this record
  # Let errors bubble out to be caught & recorded by the BatchJob
end
```

Each new instance of `BatchJob` gets a unique `run_id` on creation (or if you call '#new_run!' manually).
This is used to tie together the generated `BatchJobLogEntry` records for each run, which are then used to report progress / completion stats.

It will log as it goes, both to STDOUT and by writing `BatchJobLogEntry` record for each processed record. On completion, it will finish up by returning a Hash of job stats, like this -

```
{:failure=>2044,
 :success=>31,
 :max_created_at=>"2020-12-15T16:54:56.031Z",
 :min_created_at=>"2020-12-15T16:54:21.028Z",
 :number_of_records=>2075,
 :duration=>35.002967,
 :records_per_second=>59.28068897702301,
 :time_per_record=>0.016868899759036143}
```
You can generate this Hash at any time after _or during_ processing by calling:
```
BatchJob.stats(run_id: '...')
```
You can also get the latest run_id for a given job_name by calling:
```
BatchJobLogEntry.latest_for_job(job_name: 'job name').run_id
```

All generated `BatchJobLogEntry` records are both written to the database, AND collected in the jobs successes / failures arrays:

```
> job.failures.size
2044
> job.failures.first
#<BatchJobLogEntry:0x0000559707ec38f0
 id: 74295,
 record_id: "{:urn=>136577, :allocation=>53}",
 record_class: "Hash",
 job_name: "raise to 100%",
 run_id: "567af869-7f9f-4080-85ed-38299a8e251b",
 status: "failure",
 message: nil,
 error: "undefined method `update!' for nil:NilClass",
 created_at: Tue, 15 Dec 2020 16:54:56 GMT +00:00,
 updated_at: Tue, 15 Dec 2020 16:54:56 GMT +00:00>

```

There are no explicit transactions (yet) - so that `BatchJobLogEntry` records are visible to other processes as they are written.
This will be useful for situations where several people are waiting on the outcome of a job, or a developer has multiple sessions open, with the job running in one terminal while working on other things and occasionally seeing progress in another, etc.

Future iterations could include:
* a simple UI in the admin interface for reporting progress of jobs, clearing old log entries and so on.
* a batch_size parameter for chunking up records into convenient groups / transactions
* the ability to write pending `BatchJobLogEntry` records in advance for the records you intend to process - so that progress  & status of records processed so far doesn't get lost if the ssh session is dropped
* probably much more elegant code :)

### DIFF
--- a/app/models/batch_job.rb
+++ b/app/models/batch_job.rb
@@ -3,10 +3,10 @@ class BatchJob
   attr_accessor :records
   attr_accessor :run_id, :job_name, :logger
 
-  def initialize(records:, job_name:, logger: nil)
+  def initialize(records:, job_name:, logger: Rails.logger)
     @records = records
     @job_name = job_name
-    @logger = logger || Rails.logger
+    @logger = logger
     new_run!
   end
 
--- a/app/models/batch_job.rb
+++ b/app/models/batch_job.rb
@@ -1,0 +1,55 @@
+class BatchJob
+  attr_accessor :successes, :failures
+  attr_accessor :records
+  attr_accessor :run_id, :job_name, :logger
+
+  def initialize(records:, job_name:, logger: nil)
+    @records = records
+    @job_name = job_name
+    @logger = logger || Rails.logger
+  end
+
+  def process!(&block)
+    @run_id = SecureRandom.uuid
+    @successes = []
+    @failures = []
+    @records.each_with_index do |record, index|
+      log index: index, message: "processing #{record.class.name}: #{record.id}"
+      yield record
+      success!(record)
+    rescue Exception => e
+      failure!(record: record, error: e)
+    end
+    report_stats
+  end
+
+private
+
+  def success!(record)
+    entry = BatchJobLogEntry.create!(
+      job_name:     @job_name,
+      run_id:       @run_id,
+      record_id:    record.id,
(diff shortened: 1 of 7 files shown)


jamesrose (approved):
@aldavidson Awesome. This is great.

PR #958, opened 2020-12-16
Safely handle cap update failures triggered by a devices ordered update

jamesrose:
### Context

We are occasionally returning errors to CC when they send us consumption updates. Here's what is happening:

1. They send us a consumed update
2. We save it against the allocation which triggers cap updates back to CC for the entire pool
3. At least one of these cap updates fails because the cap is below the devices ordered number
4. We return a 500 error back to them

1. Don't return a failure if a cascading cap update to CC fails

N.B. We already store failures in `CapUpdateCall` so I am not concerned that we are suppressing visibility of cap update failures.


jamesrose:
A different way might be to set something on the allocation which means we don't send updates at all when devices_ordered is updated. Particularly open to suggestions on this one.

tonyheadford (approved):

PR #960, opened 2020-12-16
Add "Cannot order over Christmas" page

fofr:
Ordering will be closed over Christmas. During this period a different order page is shown.

This PR assumes that all schools and RBs will be placed into the cannot order state during the festive period.

(Best reviewed by ignoring whitespace)

### Context

localhost_3000_schools_124500_order-devices
localhost_3000_responsible-body_devices_order-devices


jamesrose (approved):
@fofr Awesome - thanks for doing this :)

PR #971, opened 2020-12-17
Create replace_a_faulty_device

gemmaemoore:


aldavidson:
Combined with #972 nto a new PR #984

PR #972, opened 2020-12-17
Added link to Replace a faulty device

gemmaemoore:


aldavidson:
Combined with #971 into a new PR #984

PR #985, opened 2020-12-17
Update support_and_maintenance.md

gemmaemoore:


jamesrose (approved):

Lloyd-K (commented):
Line 23 - 'Jan' should be 'January'
Line 48 - 'as this will result in both the router and SIM to stop working.' should be 'as this will cause both the router and the SIM to stop working.'

PR #989, opened 2020-12-17
1151 rb can order status not updating on virtual caps consistently

tonyheadford:
### Context
[Trello card](https://trello.com/c/wgRSgQdM/1151-rb-can-order-status-not-updating-on-virtual-caps-consistently)
`PreorderInformation` uses various sources of data to determine a status. When one of the values changes, the `PreorderInformation` status is not automatically updated and could then not reflect the current state of the school. This PR is to make the `PreorderInformation` able to update its status when a dependency changes.
 
Add callbacks to ensure changes to related data notify the `PreorderInformation` so that the `status` can be recalculated.
Enabling the `PreorderInformation` able to dynamically adjust its status value resulted in many broken specs, where objects had been partially created and/or specs relied on just setting the `PreorderInformation.status` in the spec setup.  I've added a helper for creating schools with a given preorder status to simplify this setup.

### Guidance to review
Relevant changes to `School`, `SchoolDeviceAllocation`, adding or removing a `User`/ `SchoolContact` and changes to `PreorderInformation` should trigger the `status` to be updated to reflect the current state.


spec/models/school_can_order_devices_notifications_spec.rb
```diff
@@ -1,26 +1,17 @@
 require 'rails_helper'
+require 'shared/school_creator'
```
asmega, line 2: `spec/rails_helper.rb` has the line 
```ruby
Dir[Rails.root.join('spec/support/**/*.rb')].sort.each { |f| require f }
```
If you we move this file to there is should get loaded before the suite runs and you won't need to explicitly require this file every time it's needed

spec/models/school_can_order_devices_notifications_spec.rb
```diff
@@ -172,15 +166,16 @@
     end
 
     context 'when school which is not ready changes from cannot_order to can lockdown order' do
-      let!(:user) { create(:school_user, school: school) }
+      # let!(:user) { create(:school_user, school: school) }
```
asmega, line 169: delete me

spec/shared/school_creator.rb
```diff
@@ -0,0 +1,76 @@
+def create_schools_at_status(preorder_status:, count: 1, responsible_body: nil)
```
asmega, line 1: this works so we should leave for now, but we should somehow move this into the factories in the future so there is one single api to create these things

app/models/preorder_information.rb
```diff
@@ -8,6 +8,9 @@ class PreorderInformation < ApplicationRecord
 
   validates :status, presence: true
 
+  before_save :check_and_update_status_if_necessary
+  after_touch :refresh_status!
```
jamesrose, line 12: Could we do `before_touch :set_status` or similar so that we needn't save twice?

spec/models/school_can_order_devices_notifications_spec.rb
tonyheadford: Good idea.

spec/models/school_can_order_devices_notifications_spec.rb
tonyheadford: Good spot

app/models/preorder_information.rb
tonyheadford: Unfortunately there is no `before_touch`. I don't think we'd be doing 2 saves here though as the `before_save` callback won't trigger the `after_touch`. The `after_touch` callback is only triggered by `touch`es and seems to be protected in a sensible way from getting into infinite loops.

asmega (approved):

PR #990, opened 2020-12-17
Confirm when deleting a user in the support console

benilovj:
### Context

Currently, deletion of a user in the support console happens immediately, as soon as a support agent clicks the button. Since this interaction is fairly rare and can only be undone by a developer, it's desirable to add a bit of friction to make sure it's really intended.

- Introduce an interstitial confirmation page prior to deletion.
- Add a feature spec covering the deletion scenario

### Guidance to review

The designs are analogous to [course deletion in Publish Teacher Training Courses](https://bat-design-history.netlify.app/publish-teacher-training-courses/deleting-and-withdrawing/).

image

image

### DIFF
--- a/app/views/support/users/confirm_destroy.html.erb
+++ b/app/views/support/users/confirm_destroy.html.erb
@@ -1,0 +1,23 @@
+<%- title = t('page_titles.confirm_destroy_user') %>
+<%- content_for :before_content, govuk_link_to('Back', support_user_path(@user), class: 'govuk-back-link') %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">
+      <span class="govuk-caption-xl"><%= @user.full_name %></span>
+      <%= title %>
+    </h1>
+
+    <p class="govuk-body">
+      Deleting this user will revoke their access to the service. If you need to reinstate them after deletion, contact the GHWT technical support team.
+    </p>
+
+    <%= form_for @user, url: support_user_path(@user), method: :delete do |f| %>
+      <%= f.govuk_submit 'Yes, delete user', warning: true %>
+    <%- end %>
+
+    <p class="govuk-body">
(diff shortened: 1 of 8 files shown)


app/views/support/users/confirm_destroy.html.erb
```diff
@@ -0,0 +1,23 @@
+<%- title = t('page_titles.confirm_destroy_user') %>
```
fofr, line 1: Page title should be:

"Are you sure you want to delete this user?"

Then the button text:

"Yes, delete this user"

app/views/support/users/confirm_destroy.html.erb
```diff
@@ -0,0 +1,23 @@
...
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">
+      <span class="govuk-caption-xl"><%= @user.full_name %></span>
+      <%= title %>
+    </h1>
+
+    <p class="govuk-body">
+      Deleting this user will revoke their access to the service. If you need to reinstate them after deletion, contact the GHWT technical support team.
```
fofr, line 12: Will it also remove them from the support portal (service now) and techsource?

app/views/support/users/confirm_destroy.html.erb
benilovj: fixed and force-pushed

image

app/views/support/users/confirm_destroy.html.erb
benilovj: > Will it also remove them from the support portal (service now)

Yes, this was already done as part of https://github.com/DFE-Digital/get-help-with-tech/pull/748

benilovj: Minor bugfix on the back of your comment: https://github.com/DFE-Digital/get-help-with-tech/pull/994

fofr (approved):

PR #992, opened 2020-12-18
[1173] allow support users to bulk-set allocations by URN

aldavidson:
### Context

[Trello card 1173](https://trello.com/c/7GfEMnUx/1173-add-a-facility-in-the-admin-ui-for-setting-allocations-on-subsets-of-schools) - The dev team have spent quite a lot of time planning / writing / running arbitrary code to bulk-update allocations for given sets of URNs in response to policy changes, and the rest of the service team have spent lots of time waiting for the dev team to do these. It would be much better to have an admin interface to allow the service team to do this themselves. 

* add a basic UI to allow users to enter URNs and allocations, and have them processed with failures reported

### Guidance to review

I've tested this with a copy-and-paste from two adjacent rows of a Google sheet (works fine)
This is a first iteration which performs the job synchronously from the controller - further work could include:

* allowing it to be run asynchronously from a Sidekiq job, although we'd need another iteration on the `BatchJob` class to get progress reporting to work in this scenario.
* allowing the list of failed URNs and errors to be downloaded
* allowing caps to be specified at the same time
* allowing the user to choose a new order_state for all given schools
etc etc

Screenshot from 2020-12-18 11-07-25
Screenshot from 2020-12-18 11-10-53


app/form_objects/support/urns_and_allocations_form.rb
```diff
@@ -0,0 +1,39 @@
...
+    @urns_and_allocations = args[:urns_and_allocations]
+  end
+
+  def parse_urns_and_allocations!
+    return [] if urns_and_allocations.empty?
+
+    @parsed_list = urns_and_allocations.split("\r\n")
+                                        .map(&:strip)
+                                        .reject(&:blank?)
+                                        .map { |line| line.split(/[\s+,]/) }
```
asmega, line 18: The following might be a bit more robust?
```
[37] pry(main)> "1      3".split(/\s*,\s*|\s*/)
=> ["1", "3"]
[38] pry(main)> "1  ,    3".split(/\s*,\s*|\s*/)
=> ["1", "3"]
```

aldavidson: good call - I've gone with the more concise but equally effective adjustment: 
`/[\s+,]/` -> `/[\s,]+/`
and added a feature spec

asmega (approved):

PR #994, opened 2020-12-18
Don't allow support operations on deleted users

benilovj:
### Context

Deleted users are filtered out in support search and various user lists, but it's still possible to get to the edit pages if you know the user's id.

Change the user management functionality to return 404 when they're applied to deleted users.


aldavidson (approved):
:+1:

PR #999, opened 2020-12-22
Add validations to get support forms

aliuk2012:
- Add email address format validation
- Add School URN validation to ensure its a 6 digit number
- Add College UKPRN validation to ensure its a 8 digit number

### DIFF
--- a/app/form_objects/support_ticket/contact_details_form.rb
+++ b/app/form_objects/support_ticket/contact_details_form.rb
@@ -5,4 +5,5 @@ class SupportTicket::ContactDetailsForm
 
   validates :full_name, presence: { message: 'Enter your full name' }
   validates :email_address, presence: { message: 'Enter your email address' }
+  validates :email_address, email_address: true
 end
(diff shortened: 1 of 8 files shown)


app/form_objects/support_ticket/contact_details_form.rb
```diff
@@ -5,4 +5,6 @@ class SupportTicket::ContactDetailsForm
 
   validates :full_name, presence: { message: 'Enter your full name' }
   validates :email_address, presence: { message: 'Enter your email address' }
+  validates :email_address, format: { with: URI::MailTo::EMAIL_REGEXP, message: "You must enter an email address in the correct format, like name@example.com"}
```
benilovj, line 8:
```suggestion
  validates :email_address, email_address: true
```
which would use the existing `EmailAddressValidator`.

spec/form_objects/support_ticket/college_details_form_spec.rb
```diff
@@ -9,9 +9,59 @@
       expect(form.errors.details[:college_name].first[:error]).to equal(:blank)
     end
 
-    it 'requires a college UKPRN' do
-      form.valid?
-      expect(form.errors.details[:college_ukprn].first[:error]).to equal(:blank)
+    describe 'UKPRN' do
```
benilovj, line 12: We already use the [shoulda-matchers gem](https://github.com/thoughtbot/shoulda-matchers) which can help avoid these longhand specs for validation – here's some prior art:

https://github.com/DFE-Digital/get-help-with-tech/blob/main/spec/models/school_contact_spec.rb#L12-L20

aliuk2012: Ah, much better thanks for letting me know that I was using the wrong pattern I found in the code base

benilovj (approved):

PR #1000, opened 2020-12-22
Support console: user management redesign (part 2)

benilovj:
~Depends on #998 - merge that first.~ ✅ 
~Depends on #1002 - merge that first.~ ✅ 

### Context

User management is very commonly used by support agents, and as such, should follow similar best practice design and code patterns.

### Changes proposed in this pull request

Tidy up the design of various user management functions in the support console:

- editing a user's responsible body
- granting access to, and removing access to schools

#### Setting a responsible body when there wasn't one before

**Before:**

image

**After:**

image

#### Removing a responsible body and moving to a new responsible body

**Before:**

image

**After:**

image

image

#### Removing school access

**Before:**

image

**After:**

image

#### Suggesting a new school for access

**Before:**

image

**After:**

image

PR #1002, opened 2020-12-24
School search: backend improvements

benilovj:
### Context

The support console allows support agents to grant school access to users. When the agent types in part of a school name (or a school URN), the system returns schools that match the provided search term.

There's an ambition to extend this functionality to a more generalised schools auto-complete, to use in other places in the support console and in the user support form. This technical change is a step towards realising this ambition.

1. Introduce an upper limit to how many results the school suggest returns. A user can hit this limit if they enter a very short or common search term (e.g. `St`). Returning more than 50 suggestions isn't a useful user interaction – in that case, the user should refine their search.

2. Improve the quality of matched results by switching to the [`pg_search` gem and PostgreSQL's full text search](https://github.com/Casecommons/pg_search#tsearch-full-text-search). For instance, after this change, searching for `st joseph's` now matches schools called `St. Joseph's` (with a full stop) as well as `St Joseph's` (without a full stop).

### Guidance to review

-

### DIFF
--- a/app/form_objects/support/new_user_school_form.rb
+++ b/app/form_objects/support/new_user_school_form.rb
@@ -9,8 +9,9 @@ class Support::NewUserSchoolForm
   end
 
   def matching_schools
-    School.includes(:responsible_body)
-          .where('urn = ? OR name ILIKE(?)', @name_or_urn.to_i, "%#{@name_or_urn}%")
-          .order(:name)
+    School
+      .matching_name_or_urn(@name_or_urn)
+      .includes(:responsible_body)
+      .order(:name)
   end
 end
--- a/app/form_objects/support/new_user_school_form.rb
+++ b/app/form_objects/support/new_user_school_form.rb
@@ -1,17 +1,15 @@
 class Support::NewUserSchoolForm
   include ActiveModel::Model
 
-  attr_accessor :user, :name_or_urn
+  MAX_NUMBER_OF_SUGGESTED_SCHOOLS = 50
 
-  def initialize(user:, name_or_urn: nil)
-    @user = user
-    @name_or_urn = name_or_urn
-  end
+  attr_accessor :user, :name_or_urn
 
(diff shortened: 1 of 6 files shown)


aliuk2012:
Thinking about the issue with limits and Support agents accidentally querying the list of schools with 1 or 2 character school names, perhaps we should add a min length of characters for the query eg. min length of 3.

benilovj:
> Thinking about the issue with limits and Support agents accidentally querying the list of schools with 1 or 2 character school names, perhaps we should add a min length of characters for the query eg. min length of 3.

This is difficult to do on this (backend-only) change, so I've done it here on #1000:

https://github.com/DFE-Digital/get-help-with-tech/pull/1000#issuecomment-752128652

app/form_objects/support/new_user_school_form.rb
```diff
@@ -9,8 +11,10 @@ def initialize(user:, name_or_urn: nil)
...
 
   def matching_schools
-    School.includes(:responsible_body)
-          .where('urn = ? OR name ILIKE(?)', @name_or_urn.to_i, "%#{@name_or_urn}%")
-          .order(:name)
+    School
+      .matching_name_or_urn(@name_or_urn)
+      .includes(:responsible_body)
+      .order(:name)
+      .limit(MAX_NUMBER_OF_SUGGESTED_SCHOOLS)
```
aliuk2012, line 18: I noticed that we dont have any unit tests for this form_object.r5

aliuk2012 (commented):
Only a minor issue relating to missing specs but otherwise looks good to me.

app/form_objects/support/new_user_school_form.rb
benilovj: I've added coverage to f4c220a and force-pushed.

aliuk2012 (approved):

PR #1011, opened 2020-12-30
Refactor select all/none js functionality & remove jQuery

aliuk2012:

This was in desperate need of being a refactor for the following reasons:

- jQuery: This relied on jQuery and seemed a bit of overkill to import and include it on every page. It also seemed to be included twice.
  jQuery is no longer a dependency for this code to work.

- would only work for a single instance of this component. It now supports
  multiple components

- The js was not scoped to just the component in question. For example,
  when the page has loaded, any element that had a class of `js-only` or `non-js-only` would be
  hidden/shown regardless of what our intentions were for those elements
  when the page loaded.

- There was a lot of additional markup added just to add the feature.
  The markup has now been cleaned up.

- No unit tests

### Guidance to review

I analyzed the bundles locally before & after the changes:

<table>
<tr>
    <th> 
    <th>Before
    <th>After
    <th>% Reduction
<tr>
	<td>gzipped
	<td>130KB
	<td>35KB
	<td>73%
<tr>
	<td>parsed
	<td>408KB
	<td>117KB
	<td>71%
</table>

#### Before:
image

#### After:
image

### DIFF
--- a/app/views/mno/extra_mobile_data_requests/_extra_mobile_data_request.html.erb
+++ b/app/views/mno/extra_mobile_data_requests/_extra_mobile_data_request.html.erb
@@ -1,14 +1,10 @@
 <tr class="govuk-table__row" id="request-<%= extra_mobile_data_request.id %>">
   <td class="govuk-table__cell">
-    <div class="govuk-form-group govuk-!-margin-bottom-0">
-      <%= form.govuk_collection_check_boxes :extra_mobile_data_request_ids,
-      [OpenStruct.new(id: extra_mobile_data_request.id, label: '')],
-      :id,
-      :label,
-      small: false,
-      classes: 'all-none-item',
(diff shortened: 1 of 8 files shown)


benilovj (approved):
Looks good – nice cleanup. Now that the module is generic, how about renaming it from `extra-mobile-data-requests.js` to something more generic?

aliuk2012:
There will be a follow-up PR, which will include some tidy ups on how webpack is generating our assets.

aliuk2012:
[`e0b4af8` (#1012)](https://github.com/DFE-Digital/get-help-with-tech/pull/1012/commits/e0b4af8e976c0135af4d93939a0562b3278dc7e4) renames `extra-mobile-data-requests.js`

benilovj:
:shipit:

PR #1014, opened 2020-12-31
School autocomplete progressive enhancement

benilovj:
### Context

Support agents need to assign users to new schools.

- Add a JSON endpoint which returns schools based on a partial name or URN match
- Progressively enhance the school search box with an accessible autocomplete which uses the newly introduced JSON endpoint.

### Guidance to review

https://user-images.githubusercontent.com/23801/103418327-10884900-4b86-11eb-97ba-e992fa8c18c7.mp4

### DIFF
--- a/app/controllers/support/schools_controller.rb
+++ b/app/controllers/support/schools_controller.rb
@@ -16,8 +16,13 @@ class Support::SchoolsController < Support::BaseController
         end
       end
     elsif request.get?
-      @schools = Support::NewUserSchoolForm.new(name_or_urn: params[:query]).matching_schools
-      render json: @schools.as_json(only: %i[id name urn postcode town])
+      @form = Support::NewUserSchoolForm.new(name_or_urn: params[:query])
+      if @form.valid?
+        @schools = @form.matching_schools
+        render json: @schools.as_json(only: %i[id name urn postcode town])
+      else
+        render json: { errors: @form.errors.full_messages }, status: :unprocessable_entity
+      end
     end
   end
 
--- a/app/controllers/support/schools_controller.rb
+++ b/app/controllers/support/schools_controller.rb
@@ -6,13 +6,18 @@ class Support::SchoolsController < Support::BaseController
   end
 
   def results
-    @search_form = SchoolSearchForm.new(search_params)
-    @schools = policy_scope(@search_form.schools).includes(:preorder_information, :responsible_body)
-    respond_to do |format|
-      format.html {}
-      format.csv do
-        send_data AllocationsExporter.new.export(@schools), filename: @search_form.csv_filename
+    if request.post?
+      @search_form = SchoolSearchForm.new(search_params)
+      @schools = policy_scope(@search_form.schools).includes(:preorder_information, :responsible_body)
+      respond_to do |format|
+        format.html {}
+        format.csv do
+          send_data AllocationsExporter.new.export(@schools), filename: @search_form.csv_filename
+        end
       end
+    elsif request.get?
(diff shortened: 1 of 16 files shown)


app/controllers/support/schools_controller.rb
```diff
@@ -6,8 +6,18 @@ def search
   end
 
   def results
-    @search_form = BulkUrnSearchForm.new(search_params)
-    @schools = policy_scope(@search_form.schools).includes(:preorder_information, :responsible_body)
+    if request.post?
+      @search_form = BulkUrnSearchForm.new(search_params)
+      @schools = policy_scope(@search_form.schools).includes(:preorder_information, :responsible_body)
+    elsif request.get?
+      @form = Support::NewUserSchoolForm.new(name_or_urn: params[:query])
```
aldavidson, line 13: Not a blocker, but this feels a bit fragile - and also limits potential for using the endpoint from other places

benilovj: Agreed – the name of the form object isn't really reflective of its responsibilities so I've renamed it in f9e00d1

aldavidson: :+1:

benilovj:
@aldavidson I've rebased this PR so it no longer conflicts with `main`. Would you mind taking another look?

aldavidson:
> @aldavidson I've rebased this PR so it no longer conflicts with `main`. Would you mind taking another look?

I've had a look, and tried it locally. 
The autocomplete works, and the original workflow is still there if you disable javascript - all good.

But I did find it a bit counter-intuitive that when you select a school from the autocomplete suggestion, you still then have to click 'Continue' to search for it, and then click 'Associate' to associate the school - I would have expected the act of clicking in the autocomplete list to replace at least the first of those clicks. I found myself repeatedly thinking I'd associated a school when I actually hadn't yet, and it feels like it adds a step to the workflow rather than removes one. Does that make sense?

benilovj:
> > @aldavidson I've rebased this PR so it no longer conflicts with `main`. Would you mind taking another look?
> 
> I've had a look, and tried it locally.
> The autocomplete works, and the original workflow is still there if you disable javascript - all good.
> 
> But I did find it a bit counter-intuitive that when you select a school from the autocomplete suggestion, you still then have to click 'Continue' to search for it, and then click 'Associate' to associate the school - I would have expected the act of clicking in the autocomplete list to replace at least the first of those clicks. I found myself repeatedly thinking I'd associated a school when I actually hadn't yet, and it feels like it adds a step to the workflow rather than removes one. Does that make sense?

Yes this makes sense, but I don't want to make this PR bigger than it already is, and I have a follow-up in flight already that would change the interaction design (https://github.com/DFE-Digital/get-help-with-tech/pull/1000). Can I do the work you've suggested on the other PR?

aldavidson:
> > > @aldavidson I've rebased this PR so it no longer conflicts with `main`. Would you mind taking another look?
> > 
> > 
> > I've had a look, and tried it locally.
> > The autocomplete works, and the original workflow is still there if you disable javascript - all good.
> > But I did find it a bit counter-intuitive that when you select a school from the autocomplete suggestion, you still then have to click 'Continue' to search for it, and then click 'Associate' to associate the school - I would have expected the act of clicking in the autocomplete list to replace at least the first of those clicks. I found myself repeatedly thinking I'd associated a school when I actually hadn't yet, and it feels like it adds a step to the workflow rather than removes one. Does that make sense?
> 
> Yes this makes sense, but I don't want to make this PR bigger than it already is, and I have a follow-up in flight already that would change the interaction design (#1000). Can I do the work you've suggested on the other PR?

Sure :+1:

aldavidson (approved):

PR #1020, opened 2021-01-04
[286] Add form objects needed for get support flow

aliuk2012:
### Context

The breakup of #889 into smaller PR, these new form objects are not being used in the service at the moment as the views/controllers, etc will be part of another PR. If you want to view the user journey you can see it by visiting https://dfe-ghwt-pr-889.herokuapp.com/get-support

The form objects will be used by the form builder. We do not want to
store the user's query in the service which is why it's not using 
traditional ActiveModel implementation to store it in a database table.

### DIFF
--- a/app/form_objects/support_ticket/describe_yourself_form.rb
+++ b/app/form_objects/support_ticket/describe_yourself_form.rb
@@ -1,0 +1,29 @@
+class SupportTicket::DescribeYourselfForm
+  include ActiveModel::Model
+
+  attr_accessor :user_type
+
+  validates :user_type, presence: { message: 'Tell us which of these best describes you' }
+
+  OPTIONS = {
+    school_or_single_academy_trust: 'I work for a school or single-academy trust',
+    multi_academy_trust: 'I work for a multi-academy trust',
+    local_authority: 'I work for a local authority',
+    college: 'I work for a college',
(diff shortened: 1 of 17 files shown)


app/form_objects/support_ticket/describe_yourself_form.rb
```diff
@@ -0,0 +1,32 @@
...
+  def describe_yourself_options
+    options = []
+
+    OPTIONS.each do |option_value, option_label|
+      options << OpenStruct.new(
+        value: option_value,
+        label: option_label,
+      )
+    end
+    options
```
asmega, line 26:
```suggestion
    OPTIONS.map do |option_value, option_label|
      OpenStruct.new(
        value: option_value,
        label: option_label,
      )
    end
```

app/form_objects/support_ticket/support_needs_form.rb
```diff
@@ -0,0 +1,39 @@
...
+  def support_needs_options
+    options = []
+
+    OPTIONS.each do |option_value, option_label|
+      options << OpenStruct.new(
+        value: option_value,
+        label: option_label,
+      )
+    end
+    options
```
asmega, line 25:
```suggestion
    OPTIONS.map do |option_value, option_label|
      OpenStruct.new(
        value: option_value,
        label: option_label,
      )
    end
```

PR #1022, opened 2021-01-04
[286] Routes/controllers/views for support wizard

aliuk2012:
### Context
The breakup of #889 into smaller PR, this can be merged as-is, and the user should not be able to find it because we have not published any URL in the service or comms. If you want to view the user journey you can see it by visiting https://dfe-ghwt-pr-889.herokuapp.com/get-support

- Adds routes/controllers to the get support form pages
- Adds views for the get support form pages
- Only has basic validation for required fields
- Does not create any zendesk tickets
### Guidance to review

Visit https://dfe-ghwt-pr-1022.herokuapp.com/get-support

### DIFF
--- a/app/views/support_tickets/thank_you.html.erb
+++ b/app/views/support_tickets/thank_you.html.erb
@@ -1,0 +1,24 @@
+<% content_for :title,  'Thank you for completing this form' %>
+
+<%- content_for :before_content do %>
+  <% breadcrumbs([{ 'Home' => guidance_page_path },
+                  {t('landing_pages.get_support.start') => support_ticket_path},
+                  'Thank you for completing this form',
+                 ]) %>
+<% end %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+
+    <h1 class="govuk-heading-l">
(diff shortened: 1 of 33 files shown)


app/views/support_tickets/thank_you.html.erb
```diff
@@ -0,0 +1,24 @@
...
+<% end %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+
+    <h1 class="govuk-heading-l">
+      Thank you for completing this form
+    </h1>
+    <p class="govuk-body">Our support team has received your request. If you do not receive an email from us within that time, please check your spam or junk folder.</p>
+
+    <p class="govuk-body">We aim to respond within 3 working days.</p>
```
aldavidson, line 17: Seems like the sentences here are slightly out-of-order - shouldn't "If you do not receive an email from us within that time"  come _after_ "We aim to respond within 3 working days" ?

aldavidson (approved):
Just one comment about wording on the thank you page,  but it's not a blocker - otherwise looks fine :+1:

PR #1026, opened 2021-01-04
Add further constraints around lowering caps with RBs with multiple Chromebook domains

jamesrose:
### Context

This is an extension of https://trello.com/c/JFczhXmY/1097-virtual-cap-force-the-orders-of-trusts-that-qualify-for-virtual-cap-and-have-multiple-chromebook-domains-into-the-techsource-man and https://github.com/DFE-Digital/get-help-with-tech/pull/927

There's a possibility that a school who is not centrally managed is within an RB that largely manages its schools. In that case and where there were multiple Chromebooks domains we are erroneously forcing devolved schools into the CC manual queue.

We need a check to determine whether the RB will order devices as well as checking if the RB has multiple Chromebook domains in managed schools before reducing the cap we send to CC. 

Don't force schools that are devolved in a centrally managed RB into the manual queue if the RB has multiple Chromebook domains

### Guidance to review

A devolved school in a multiple Chromebook domain and centrally managed RB should still use its unadjusted cap.

### DIFF
--- a/app/models/computacenter/outgoing_api/cap_update_request.rb
+++ b/app/models/computacenter/outgoing_api/cap_update_request.rb
@@ -55,7 +55,7 @@ class Computacenter::OutgoingAPI::CapUpdateRequest
 private
 
   def get_allocation_data
-    records = SchoolDeviceAllocation.includes(:school).where(id: @allocation_ids)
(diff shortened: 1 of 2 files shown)


tonyheadford (approved):
Looks great, nice catch

PR #1027, opened 2021-01-04
Support console: streamline the design of the user's 'update schools' page

benilovj:
### Context

The support console allows removing and adding new schools for users of the system.

Streamline the design of the user's 'update schools' page by:

- not surfacing information that's unnecessary to removing a school, like the responsible body and the type of school (since support requests almost always mention the name of the school from which the user needs to be removed)
- allowing bulk removal (by unticking multiple schools), instead of having to do it one button at a time

### Guidance to review

**Before:**

image

**After:**

image


aldavidson (approved):
LGTM :+1:

PR #1028, opened 2021-01-04
Support console: streamline the editing a user's responsible body

benilovj:
### Context

The support console allows removing and adding new schools for users of the system.

Apply the same interaction patterns for:

- adding
- updating
- removing a user's responsible body

as are present in the school-user-facing parts of the app.

#### Setting a responsible body when there wasn't one before

**Before:**

image

**After:**

image

#### Removing a responsible body and moving to a new responsible body

**Before:**

image

**After:**

image

image

### DIFF
--- a/app/controllers/support/users/responsible_body_controller.rb
+++ b/app/controllers/support/users/responsible_body_controller.rb
@@ -10,7 +10,8 @@ class Support::Users::ResponsibleBodyController < Support::BaseController
   end
 
   def update
-    @user.update!(responsible_body_id: responsible_body_params[:responsible_body])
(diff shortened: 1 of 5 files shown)


aldavidson (approved):
Looks good and works well. 

It's a _little_ bit jarring that when you go to add an RB to a user who doesn't already have one, that DfE is pre-selected and the only option in the list until you clear out the box (see screenshot) but not a blocker and not worth investing time to fix unless it's a quick one

Screenshot from 2021-01-05 09-56-46

benilovj:
@aldavidson good call: I think I've fixed it in 3e43102:

image

PR #1034, opened 2021-01-05
[1193] Show feature flags in support area

asmega:


app/views/support/home/feature_flags.html.erb
```diff
@@ -0,0 +1,37 @@
...
+        <tr class="govuk-table__row">
+          <th scope="col" class="govuk-table__header">Feature</th>
+          <th scope="col" class="govuk-table__header">Status</th>
+        </tr>
+      </thead>
+
+      <tbody class="govuk-table__body">
+        <% FeatureFlag::FEATURES.sort.each do |feature| %>
+          <tr class="govuk-table__row">
+            <th scope="row" class="govuk-table__header"><%= feature %></th>
```
aliuk2012, line 26:
```suggestion
            <th scope="row" class="govuk-table__header"><%= feature.humanize %></th>
```

aliuk2012: So that it's a bit easier for non-devs to read ;)

jamesrose (approved):
Thank you!

app/views/support/home/feature_flags.html.erb
asmega: I'd think i'd rather leave this as is for now. As I don't want it get lost in translation as it's the actual value and we don't have an associated description or look up value. If it gets confusing or misleading I think we can add something then

PR #1038, opened 2021-01-05
Creates partial with delivery timeframe and adds it to pages

ralph-hawkins:
### Context

We need to update the text that says deliveries will be made in 2 days to 'Most devices will be delivered to schools within 5 working days of an order being confirmed, subject to stock availability.'

This is the second time I've updated how many days delivery will take so I've created a partial and swapped it in to the places where we mention how long deliveries take. Not sure I've done it in the most elegant way though.

### Guidance to review

Look at all the pages with delivery mentioned, you'll need v-caps enabled for one of the pages.


Lloyd-K (approved):

app/views/devices_guidance/how_to_order.html.erb
```diff
@@ -325,7 +325,7 @@
 
           <div class="app-step-nav__panel" id="step-panel-receive-and-distribute-laptops-and-tablets-6">
             <p class="app-step-nav__paragraph">
-              Most devices will be delivered to schools within 2 working days of an order being confirmed, subject to stock availability. For large orders (more than 150 devices), we’ll call the main delivery contact to arrange delivery.
+              <%= render partial: 'shared/delivery_message' %> For large orders (more than 150 devices), we’ll call the main delivery contact to arrange delivery.
```
jamesrose, line 328: If this is a singular string, this might be better placed in the locales file (`config/locales/en.yml`) and referenced with `<%= t('key.name') %>`

aliuk2012 (approved):

PR #1045, opened 2021-01-06
Update content for schools who cannot order during lockdown

fofr:
### Context

Schools:

Screen Shot 2021-01-06 at 16 16 16

Responsible bodies:

Screen Shot 2021-01-06 at 15 56 15

Draft until I've checked the content

### DIFF
--- a/app/services/feature_flag.rb
+++ b/app/services/feature_flag.rb
@@ -8,10 +8,10 @@ class FeatureFlag
     mno_offer
     virtual_caps
     christmas_banner
-    ordering_closed_for_christmas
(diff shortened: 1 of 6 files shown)


emmajhf:
Morning @fofr. 

I wonder whether it's better not to lead with 'we've invited all secondaries to open' because this isn't much comfort to primaries that will be waiting until 22 Jan.

How about:

'You cannot order devices yet

We're opening orders for primary schools gradually, and will invite all to order by 22 January.

We'll email you as soon as you can place an order.'

emmajhf (approved):

PR #1051, opened 2021-01-07
Explain why schools that can order for specific circumstances can't order their full allocation yet (during lockdown)

fofr:
It's because we haven't opened up ordering, rather than a closure not being reported.

Screen Shot 2021-01-07 at 11 36 59


fofr:
There's an assumption in this that no secondary school is set to "ordering for specific circumstances"

PR #1061, opened 2021-01-07
tell Cloudfront to cache all static assets for 1hr

aldavidson:
### Context

see [Trello card #1242](https://trello.com/c/Q0lJyeVV/1242-improve-our-use-of-caching-and-cloudfront) - Cloudfront is not actually caching our static assets, due to 

a) what looks like a mis-matched ENV var reference that should have been replaced with a `Settings` var
b) the corresponding `Settings` var not being set anyway

Replace the ENV var STATIC_FILE_CACHE_TTL with `Settings.static_file_cache_ttl`
Set the value on prod to 1 hour

### Guidance to review

This can only really be tested on prod, as that's the only Cloudfront-ed environment - but we should do so carefully.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -9,8 +9,8 @@ Rails.application.configure do
   # Eager load code on boot. This eager loads most of Rails and
   # your application in memory, allowing both threaded web servers
   # and those relying on copy on write to perform better.
-  # Rake tasks automatically ignore this option for performance..
-  .
+  # Rake tasks automatically ignore this option for performance.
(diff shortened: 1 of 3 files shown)


aldavidson:
I'm closing this PR, as after further investigation this fixes a problem we don't actually have. The site _is_ serving the correct headers to Cloudfront, the issue is actually that Cloudfront is setup to forward all headers to our app, and therefore [not actually cache anything](https://docs.cloud.service.gov.uk/deploying_services/use_a_custom_domain/#caching)

PR #1062, opened 2021-01-07
Support console: redesign the 'add school to a user' flow

benilovj:
### Context

Support agents can grant access to schools to existing users via the support console.

- Avoid showing the interstitial 'choose a school' page if the support agent picks a single school from the auto-complete or enters an exact name or URN match
- Strip down the 'choose a school' interstitial page and make searching again easier
- Add validation if the support agent inadvertently enters a school name that's too short (to avoid returning thousands of results)

#### Picking a school in the auto-complete

https://user-images.githubusercontent.com/23801/103926416-cddbe900-5110-11eb-9b2c-5ce7e219d334.mp4

#### Choosing from multiple suggestions

https://user-images.githubusercontent.com/23801/103926397-c61c4480-5110-11eb-8af3-1c91f8d80654.mp4

#### Entering a school name that's too short

image

#### Not finding any matches

image

#### Finding too many results

image

### DIFF
--- a/app/form_objects/support/school_suggestion_form.rb
+++ b/app/form_objects/support/school_suggestion_form.rb
@@ -1,7 +1,7 @@
 class Support::SchoolSuggestionForm
   include ActiveModel::Model
 
-  MAX_NUMBER_OF_SUGGESTED_SCHOOLS = 50
+  MAX_NUMBER_OF_SUGGESTED_SCHOOLS = 20
 
   attr_accessor :name_or_urn, :school_urn, :except
 
--- a/app/form_objects/support/school_suggestion_form.rb
+++ b/app/form_objects/support/school_suggestion_form.rb
@@ -5,7 +5,7 @@ class Support::SchoolSuggestionForm
 
   attr_accessor :name_or_urn, :school_urn, :except
 
-  validates :name_or_urn, length: { minimum: 3 }, allow_blank: true
+  validates :name_or_urn, length: { minimum: 3 }, unless: ->(form) { form.school_urn.present? }
 
   def matching_schools
     schools = school_by_urn.presence || schools_by_name_or_urn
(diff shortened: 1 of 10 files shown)


tonyheadford (approved):
Looks great!  Couple of non-blocking questions - is 50 too many results here and is the sort order of the results name? Looks a bit off in the screen shot.

app/form_objects/support/school_suggestion_form.rb
```diff
@@ -3,19 +3,44 @@ class Support::SchoolSuggestionForm
 
   MAX_NUMBER_OF_SUGGESTED_SCHOOLS = 50
 
-  attr_accessor :name_or_urn, :school_urn
+  attr_accessor :name_or_urn, :school_urn, :except
 
-  validates :name_or_urn, length: { minimum: 3 }
+  validates :name_or_urn, length: { minimum: 3 }, allow_blank: true
```
aliuk2012, line 8: Why do we need the `allow_blank` set to true?

benilovj: It was intended to allow the case where `school_urn` is populated (by the JS-powered auto-complete) and `name_or_urn` is not. I've updated it in dbe5495 to make this clearer.

aliuk2012 (approved):

benilovj:
> Looks great! Couple of non-blocking questions - is 50 too many results here 

Yes, you're probably right, I've dropped it down to 20 in ada0801

> and is the sort order of the results name? Looks a bit off in the screen shot.

I've investigated, the sort order is determined by the `pg_search` gem using some complicated in-built logic that I haven't quite figured out yet. It can be over-ridden, but since the disambiguation screen is a bit of a fallback anyway (because for the most part, users will have chosen an exactly matching school via the autocomplete), I think we can punt this until it's proven to be more of a problem.

PR #1086, opened 2021-01-11
Add Vodafone to participating mnos but remove any user access

tonyheadford:
### Context
[Trello card](https://trello.com/c/C3OiZob9/1265-list-vodafone-in-mno-service-but-no-access-for-voda-until-agreement-is-signed)
Add Vodafone to the list of participating MNOs in the bulk request spreadsheet

Update bulk requests spreadsheet include Vodafone in the options list.

### Guidance to review
As a school user requesting extra mobile data:
* Bulk option - download and open the bulk template and observe that Vodafone is an option in the list of mobile networks to choose from


db/migrate/20210111113140_add_vodafone_to_participating_mno_list.rb
```diff
@@ -0,0 +1,12 @@
+class AddVodafoneToParticipatingMnoList < ActiveRecord::Migration[6.0]
+  def change
+    reversible do |dir|
+      dir.up do
+        mno = MobileNetwork.find_by(brand: 'Vodafone')
+        mno.users.delete_all
```
jamesrose, line 6: I'm not sure we'll want to remove them, but just `update(mobile_network_id: nil)`

jamesrose: Does this need to be a migration?

tonyheadford: That should only remove the association but we could do this manually if you're happier with that

tonyheadford: Removed.

jamesrose: The rest were done manually 🤷‍♂️

jamesrose (approved):

PR #1089, opened 2021-01-11
[1075] Introduce Further Education bodies

asmega:
### Context

- https://trello.com/c/R2235yoT/1075-spike-adding-fe-institutions-to-the-service
- Handle Further Education bodies

### Out of scope

- FE schools will only have one delivery address like any other school we currently have in the service

- Introduce Single Table Inheritance to `School`
- `School` will be the parent class and abstract class
- All current schools will become `CompulsorySchool`
- Further Education schools will be added as `FurtherEducationSchool`
- Add `FurtherEducationCollection` as a child of `ResponsibleBody`

All Further Education schools will be imported to have a responsible body despite the fact that technically they do not have a responsible body. This is because the there is a lot of code that assumes the fact that a school simply has an RB

- Add `UKPRN` to `schools` table
- NOTE: an FE body must have a UKPRN and may also have a URN too

- As we have use `URN` to identify schools in URLs the look up will now search for for schools via `UKPRN` then URN as schools can now potentially have 2 identifiers

#### CC integration

- As `FurtherEducationSchool` will have an RB (FurtherEducationCollege) the generation of RB and Schools in `/computacenter` remain similar to how current RBs and Schools worl
- `UserChanges` behave the same as any current `School` only not us the RB "urn" and School "urn" identifiers will be the same 

#### Next steps

- Either add copy/content directly to schools controllers namespace OR add another controller namespace called `/further_education_schools` to give custom journey for these users
- Once list of FE bodies are finalised, these can be imported into the service

### Questions

- Where do we attach users to? As we have a placeholder RB and the school they could be on either. My thoughts are they should be added to the RB. My understanding is that RB users according CC have more permissions eg request BIOS resets?

### Guidance to review

n/a

### DIFF
--- a/app/form_objects/school_search_form.rb
+++ b/app/form_objects/school_search_form.rb
@@ -1,14 +1,14 @@
 class SchoolSearchForm
   include ActiveModel::Model
 
-  attr_accessor :urns, :responsible_body_id, :order_state
+  attr_accessor :identifiers, :responsible_body_id, :order_state
 
-  def array_of_urns
-    @array_of_urns ||= urns.to_s.split("\r\n").map(&:strip).reject(&:blank?)
+  def array_of_identifiers
+    @array_of_identifiers ||= identifiers.to_s.split("\r\n").map(&:strip).reject(&:blank?)
   end
 
   def request_count
-    array_of_urns.size
+    array_of_identifiers.size
   end
 
   def missing_count
@@ -21,15 +21,15 @@ class SchoolSearchForm
 
   def schools
     school_records = School.gias_status_open.includes(:responsible_body, :std_device_allocation)
-    school_records = school_records.where(urn: array_of_urns) if array_of_urns.present?
+    school_records = school_records.where('urn IN (?) OR ukprn in (?)', array_of_identifiers, array_of_identifiers) if array_of_identifiers.present?
     school_records = school_records.where(responsible_body_id: responsible_body_id) if responsible_body_id.present?
     school_records = school_records.where(order_state: order_state) if order_state.present?
 
     @schools ||= school_records
   end
 
-  def missing_urns
-    array_of_urns - schools.pluck(:urn).map(&:to_s)
+  def missing_identifiers
+    array_of_identifiers - schools.pluck(:urn, :ukprn).flatten.compact.map(&:to_s)
   end
 
   def select_responsible_body_options
@@ -44,7 +44,7 @@ class SchoolSearchForm
 
   def csv_filename
     tokens = %w[allocations]
-    tokens << "#{array_of_urns.count}-URNs" if urns.present?
+    tokens << "#{array_of_identifiers.count}-URNs" if identifiers.present?
     tokens << "RB-#{responsible_body_id}" if responsible_body_id.present?
     tokens << order_state if order_state.present?
     tokens << Time.zone.now.utc.iso8601
--- a/app/models/computacenter/responsible_body_urns.rb
+++ b/app/models/computacenter/responsible_body_urns.rb
@@ -48,6 +48,9 @@ module Computacenter::ResponsibleBodyUrns
         else
           "t#{companies_house_number.to_i}"
         end
+      when 'FurtherEducationCollege'
+        # potential n+1
+        schools.first&.ukprn
       when 'DfE'
         'DfE'
       end
@@ -57,7 +60,7 @@ module Computacenter::ResponsibleBodyUrns
       case type
       when 'LocalAuthority'
         local_authority_official_name
-      when 'Trust'
+      when 'Trust', 'FurtherEducationCollege'
         name
       when 'DfE'
         'Department for Education'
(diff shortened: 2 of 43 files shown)


app/form_objects/school_search_form.rb
```diff
@@ -44,7 +44,7 @@ def select_order_state_options
 
   def csv_filename
     tokens = %w[allocations]
-    tokens << "#{array_of_urns.count}-URNs" if urns.present?
+    tokens << array_of_identifiers.count.to_s if identifiers.present?
```
jamesrose, line 47: Are there any issues changing this token format?

jamesrose, line 47: Are there any issues changing this token format?

app/models/computacenter/responsible_body_urns.rb
```diff
@@ -48,6 +48,9 @@ def computacenter_identifier
         else
           "t#{companies_house_number.to_i}"
         end
+      when 'FurtherEducationCollege'
+        # potential n+1
+        schools.first&.ukprn
```
jamesrose, line 53: Is it worth looking at whether we're eager loading schools where we use this method?

app/models/preorder_information.rb
```diff
@@ -41,15 +41,15 @@ def infer_status
     elsif school_will_order_devices? && any_school_users? && chromebook_information_complete?
       if school.can_order_devices_right_now?
         'school_can_order'
-      elsif (school.std_device_allocation&.devices_ordered || 0).positive?
+      elsif (school.std_device_allocation&.devices_ordered || 0).positive? # bug should do all device types
```
jamesrose, line 44: Good spot! Please could you add a Trello card if you don't want to fix it in this PR?

app/models/user.rb
```diff
@@ -146,6 +146,14 @@ def effective_responsible_bodies
     user_schools.map { |us| us.school&.responsible_body }.prepend(responsible_body).compact.uniq
   end
 
+  def computacenter_account_orgs
+    effective_responsible_bodies + further_education_schools
+  end
+
+  def further_education_schools
+    user_schools.map(&:school).filter { |s| s.is_a?(FurtherEducationSchool) }
```
jamesrose, line 154: Can we do this with AR?
```suggestion
    schools.where(type: 'FurtherEducationSchool')
```

app/services/importers/fe_csv.rb
```diff
@@ -0,0 +1,79 @@
...
+          email_address: row['Contact email'],
+          full_name: row['Contact name'],
+          role: 'contact',
+          title: nil,
+          phone_number: row['Contact phone'],
+        )
+
+        next if school.preorder_information.present?
+
+        school.create_preorder_information!(
```
jamesrose, line 66: Do we need to prevent FE colleges from being able to modify this in the UI?

jamesrose (commented):
@asmega Great work! I've had a first pass over this with some questions.

app/models/preorder_information.rb
asmega: I've raised a ticket for this to keep track of it

app/form_objects/school_search_form.rb
asmega: I've changed this back, although not technically correct I think it's the safest thing to do

app/models/computacenter/responsible_body_urns.rb
asmega: I had a look for uses of `computacenter_identifier` and most seem to be referring to one RB or if it is a list it's a short list. The only exception is the "closed schools" in the `/computacenter` area. Given this and there are not many FE bodies for the moment I think this is safe for now

app/services/importers/fe_csv.rb
asmega: None of the frontend changes have been done yet. But yes if we want to keep everything on the RB we can remove it from the interface. If we want to do this we may also want to remove some other features such as being able to add users on the school.

Maybe `/schools/123456` will link to the current controller and `/schools/12345678` can link of to another controller and set of views

app/models/computacenter/responsible_body_urns.rb
asmega: the "closed schools" area has been removed now so don't think this should cause an issue at least for now

app/models/user.rb
asmega: Even better, this turned out to be dead code probably when we switched to having the placeholder RB so I will delete

app/models/school.rb
```diff
@@ -2,7 +2,8 @@ class School < ApplicationRecord
   has_paper_trail
   include PgSearch::Model
 
-  belongs_to :responsible_body
+  belongs_to :responsible_body, optional: true
```
jamesrose, line 5: I don't think we need to make this optional anymore

jamesrose (approved):
Looks great. Thanks for your work on this Phil 💯 🚀

app/models/school.rb
asmega: good spot! I've removed this now

jamesrose (approved):

PR #1092, opened 2021-01-11
Use raw cap and allocation in validation

tonyheadford:
### Context
[Trello card](https://trello.com/c/MUrYMTgO/1272-fix-device-allocation-validations)
Observed that _somehow_ in 30 instances `cap` is greater than `allocation` on the `SchoolDeviceAllocation` models. Attempting to fix this at the console proved difficult because the validators were preventing the updates to the local model while validating the broken virtual pool figures.  We don't need to do this. If we ensure that the device allocations are correct then the pool will also be correct.

Use the `raw_` cap and allocation values in the validators.

### Guidance to review
Attempting to make a cap > allocation update in a `SchoolDeviceModel` should fail with a validation error.

### DIFF
--- a/app/models/school_device_allocation.rb
+++ b/app/models/school_device_allocation.rb
@@ -116,10 +116,4 @@ private
   def has_virtual_cap_feature_flags?
     school&.responsible_body&.has_virtual_cap_feature_flags? || false
   end
-
-  def cap_lte_allocation
-    if cap > allocation
-      errors.add(:cap, :lte_allocation, allocation: allocation)
-    end
-  end
 end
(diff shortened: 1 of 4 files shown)


app/models/school_device_allocation.rb
```diff
@@ -116,10 +116,4 @@ def cap_implied_by_order_state(order_state:, given_cap: nil)
   def has_virtual_cap_feature_flags?
     school&.responsible_body&.has_virtual_cap_feature_flags? || false
   end
-
-  def cap_lte_allocation
```
jamesrose, line 120: I saw this as well - we're not using it at all and it's a duplicate of the validator.

jamesrose (approved):
Looks good and makes sense to me. I think it'd be worth getting a second opinion before merging.

PR #1094, opened 2021-01-11
[1251] Click to sign in now an interstitial page

asmega:
### Context

- https://trello.com/c/ThIADjmi/1251-increase-allocation-for-schools-with-sixth-form-students

- There is a click to sign in page after clicking the email link to login
- We want to use this page to show an interstitial page instead with context specific content

### Screenshot

image

### Guidance to review

- Login as any user, experience should not be different
- Login as school user, same experience, should see extra text with list
- Login as school with sixth form that can order devices, should now see repurposed interstitial page
- Feature should only work when feature flag enabled

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -202,6 +202,25 @@ class User < ApplicationRecord
     deleted_at.present?
   end
 
+  def associated_schools
+    scope = School.gias_status_open.distinct
+
+    if school && responsible_body
+      scope = scope.joins(:responsible_body, user_schools: [:user])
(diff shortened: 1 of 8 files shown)


fofr:
Nice. I can foresee a few uses for this. 👍

app/models/user.rb
```diff
@@ -202,6 +202,25 @@ def soft_deleted?
     deleted_at.present?
   end
 
+  def associated_schools
+    scope = School.distinct
```
tonyheadford, line 206: This this would be better
```suggestion
    scope = School.gias_status_open.distinct
```

tonyheadford (approved):
Looks good, just one suggestion

PR #1096, opened 2021-01-12
Reduced closed schools pagination

asmega:
### Context

- This page is taking way too long to generate

- As a quick fix pagination is being set to 100 items per page

### Guidance to review

- Login as CC user
- Navigate to `/computacenter/schools-that-can-order`
- Should be able to paginate thru closed schools


jamesrose (approved):

benilovj:
I think pagination without search will make this page not very useful.

asmega:
> I think pagination without search will make this page not very useful.

Agreed. But the page taking 60+ seconds to load and possibly timing out is also not very useful. I think we should get this in to look at what page actually generates then decide if we want to keep it or look to see if we can can optimize it

asmega:
This feature is no longer needed and the page will be deleted instead

PR #1098, opened 2021-01-12
Eager load virtual cap for closed schools

asmega:
### Context

- This page is taking way too long to generate

- Eager load virtual caps

### Guidance to review

- Login as CC user
- Navigate to /computacenter/schools-that-can-order
- Page should load quicker


asmega:
This feature is no longer needed and the page will be deleted instead

PR #1099, opened 2021-01-12
Fix issue where users use browser back buttons

aliuk2012:
### Context
Issue https://sentry.io/organizations/dfe-get-help-with-tech/issues/2138267573/?project=5320558

I managed to replicate this error by going through the full "Get support" journey right up to the point where a zendesk ticket is created. After
I saw the "Thank you" page, I clicked on the browser back button and saw
this error.

What happens is as a user goes through the get support wizard we store the
form data in session (`session[:support_ticket]`) but once the zendesk api
has been used we clear the session. So using the back button or trying to
access other parts of the wizard may raise similar issues.

The solution I came up with is to add a `before_action` across all the
controllers which check if the `session[:support_ticket]` is present. if
it's not then we redirect the user back to the start page.

### Guidance to review

visit https://dfe-ghwt-pr-1099.herokuapp.com/get-support, 
- fill in all the forms until you get to "Thank you" page (don't worry review apps aren't configured to use Zendesk API)
- use the browser back button (no error should be raised and you should be taken back to the get support start page

### DIFF
--- a/app/controllers/support_ticket/base_controller.rb
+++ b/app/controllers/support_ticket/base_controller.rb
@@ -13,9 +13,7 @@ class SupportTicket::BaseController < ApplicationController
 
 private
 
-  def redirect_if_no_form_data_exists
-    return unless request.get?
-
+  def require_support_ticket_data!
     redirect_to support_ticket_path if session[:support_ticket].blank?
   end
 end
--- a/app/controllers/support_ticket/base_controller.rb
+++ b/app/controllers/support_ticket/base_controller.rb
@@ -10,4 +10,12 @@ class SupportTicket::BaseController < ApplicationController
   def thank_you
     render 'support_tickets/thank_you'
   end
+
+private
+
+  def redirect_if_no_form_data_exists
+    return unless request.get?
+
+    redirect_to support_ticket_path if session[:support_ticket].blank?
+  end
 end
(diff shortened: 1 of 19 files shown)


app/controllers/support_ticket/base_controller.rb
```diff
@@ -10,4 +10,12 @@ def parent_support
   def thank_you
     render 'support_tickets/thank_you'
   end
+
+private
+
+  def redirect_if_no_form_data_exists
```
jamesrose, line 16: How about:

```suggestion
  def require_support_ticket_data!
```

app/controllers/support_ticket/base_controller.rb
```diff
@@ -10,4 +10,12 @@ def parent_support
...
+    return unless request.get?
```
jamesrose, line 17: Are non-GET requests really excluded here? If they are doing `before_action :require_support_ticket_data!, only: :new` might be a little safer.

spec/controllers/support_ticket/academy_details_controller_spec.rb
```diff
@@ -3,6 +3,7 @@
 RSpec.describe SupportTicket::AcademyDetailsController, type: :controller do
```
jamesrose, line 3: Please could you add a test case for this change? Perhaps just testing the method in `BaseController` directly would be easier.

jamesrose (approved):

PR #1104, opened 2021-01-12
Increase MNO pagination to 100 items

fofr:
https://trello.com/c/0mRLaWzh/1286-mno-bulk-update-page-limit-of-requests-from-20-to-100

20 per page is not enough. MNOs have asked for more results per page so they can bulk update more at once.

Increase from 20 to 100.

### DIFF
--- a/config/initializers/pagination.rb
+++ b/config/initializers/pagination.rb
@@ -2,4 +2,5 @@ require 'pagy'
 require 'pagy/extras/items'
 require 'pagy/extras/i18n'
 
+Pagy::VARS[:items] = 100
 Pagy::VARS[:max_items] = 9999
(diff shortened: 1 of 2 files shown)


benilovj (approved):
This is fine.

One comment to the spec:

Creating 105 records isn't going to be great, performance-wise and will add considerable time to that testcase. A couple of obvious options:

- change the `Page::VARS` before and after the testcase to a smaller number, keeping the number of MNO requests needed for testing it low.
- stub it out using `stub_const` (there's prior art in the test suite already)

fofr:
@benilovj Updated in aacb5fb

benilovj (approved):

PR #1105, opened 2021-01-12
add DeleteOldSessionsJob

aldavidson:
### Context

While investigating [Trello card 1277](https://trello.com/c/uIDcEmGH/1277-mno-bug-for-providers), I found that there are many old sessions way still in the DB, way past their expiry - there has been no mechanism for clearing them. 

Add a job to run on the hour, every hour, to purge old sessions from the DB


jamesrose (approved):
Looks good! 🚢

PR #1106, opened 2021-01-12
[1280] Increased allocations interstitials

asmega:
### Context

- https://trello.com/c/TX4JVPVh/1280-interstitial-sign-in-page-for-fe-providers-who-are-already-using-the-service
- https://trello.com/c/ThIADjmi/1251-increase-allocation-for-schools-with-sixth-form-students

- My understanding is that there are 2 feature flags `increased_sixth_form_feature_flag` and `increased_fe_feature_flag`
- Depending on which is enabled for the school they get different interstitial screens
- It's implemented so if the user has both conditions it shows the sixth form version page
- There is then some conditional content in other places so if either condition is met then this is shown
- I have removed the global feature flag as it doesn't seem like its needed

#### `increased_sixth_form_feature_flag`

- Enable feature `increased_sixth_form_feature_flag` for a school
- Ensure this school can order devices
- Login as user of this school
- Should see interstitial page for sixth form allocation increase

#### `increased_fe_feature_flag`

- Enable feature `increased_fe_feature_flag` for a school
- Ensure this school can order devices
- Login as user of this school
- Should see interstitial page for fe allocation increase

### DIFF
--- a/app/services/feature_flag.rb
+++ b/app/services/feature_flag.rb
@@ -10,7 +10,6 @@ class FeatureFlag
     secondary_mass_testing_banner
     ipad_stock_message
     schools_closed_for_national_lockdown
-    sixth_form_interstitial
   ].freeze
 
   FEATURES = (PERMANENT_SETTINGS + TEMPORARY_FEATURE_FLAGS).freeze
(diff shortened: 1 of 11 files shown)


ralph-hawkins (approved):
👍  looks good to me

jamesrose (approved):

PR #1107, opened 2021-01-13
MNO request bulk-upload improvements: part 1

benilovj:
### Context

School and RB users can upload Excel spreadsheets containing MNO requests, but the mechanism is brittle and doesn't provide useful feedback to allow the user to recover when it fails.

This is the first in a series of improvements to try to make this mechanism more robust.

- Refactor the code to make further changes easier:
  - split out the logic for parsing the Excel file and its contents into separate classes (`ExtraMobileDataRequestSpreadsheet` and `ExtraMobileDataRequestRow`)
  - change `BulkUploadForm` to use validations more Rails-idiomatically
- Make `ExtraMobileDataRequestSpreadsheet` an `ActiveModel`, for the purpose of using the standard Rails validation mechanism to catch the most common errors that occur
- Introduce the couple of validations, namely catching an un-parseable `xlsx` file and a file that's missing the expected worksheet

### Guidance to review

This PR should be easier to review commit-by-commit.
Is the microcopy for the validation errors in `en.yml` appropriate?


jamesrose (approved):
Looks great. Nice work! 🚢

tonyheadford (approved):
Looks great!

PR #1114, opened 2021-01-13
make UserChange records use now as the timestamp, rather than user.updated_at

aldavidson:
### Context

See [Trello card 1275 - Multiple user changes with the same timestamp](https://trello.com/c/mKBnxiOK/1275-multiple-userchanges-being-generated-with-the-same-user-timestamp). Computacenter have reported an issue with receiving multiple userchanges for the same user, each of which has the same timestamp. This appears to be the code acting as designed - it currently uses the `user.updated_at` as the timestamp` - but it makes it difficult for them to see what the latest state is meant to be, when several changes are committed at once.

The original behaviour made sense when UserChange records were only triggered by updating the user record, but since then we've refactored and expanded functionality, to the point where they can be generated as a result of moving schools between ResponsibleBodies, or the ResponsibleBody getting a new `computacenter_reference`, or by the schools converting to Academies..... all of which do not touch the user record at all. 

Make UserChange records use the current time as the `updated_at_timestamp`, rather than user.updated_at.

### DIFF
--- a/app/models/computacenter/user_change_generator.rb
+++ b/app/models/computacenter/user_change_generator.rb
@@ -89,7 +89,7 @@ private
   def meta_attributes
     {
       user_id: user.id,
-      updated_at_timestamp: user.updated_at,
+      updated_at_timestamp: Time.zone.now.utc,
       type_of_update: type_of_update,
     }
   end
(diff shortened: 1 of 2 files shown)


jamesrose (approved):
Looks good to me.

PR #1118, opened 2021-01-14
Normalise and validate MNO request phone numbers (MNO improvements part 2)

benilovj:
### Context

School and responsible body users are seeing the bulk MNO request spreadsheets rejected because the phone numbers are missing leading zeros (due to Excel treating the phone number as an integer). Trello: https://trello.com/c/iCX5NDxr

MNO users are also slowed down by inconsistently formatted numbers being submitted to them. Trello: https://trello.com/c/xUergL2S

Switch phone number validation from a regular expression to the [`phonelib`](https://github.com/daddyz/phonelib) gem. The gem uses the regular expressions from Google's [`libphonenumber` library](https://github.com/google/libphonenumber).

Normalise phone numbers (again using the same gem) into the national format without spaces, prior to saving the MNO request record.

### Guidance to review

Is the normalisation format appropriate? i.e. is `07123456789` the right thing to surface to MNOs or is `07123 456789` better?


tonyheadford (approved):
LGTM

benilovj:
Thanks @tonyheadford for the review. I've just pushed another small commit to tighten up the validation rules to only allow UK mobile numbers. I initially overlooked a setting on the library which meant that non-UK mobiles weren't being flagged as invalid.

PR #1122, opened 2021-01-14
Add autocomplete to Multi-Academy Trust and Local Authority pages - Get support - 

aliuk2012:
- Adds autocomplete to multi-academy trust page
- Adds autocomplete to Local Authority page

This feature is just an enhanced select list so if JS were to fail or not load for whatever reason the user would still see a list of options in a select list to choose from.

### Guidance to review

- Visit https://dfe-ghwt-pr-1122.herokuapp.com/get-support
- Attempt to create a ticket on behalf of Multi-academy trust user  (NO TICKET WILL ACTUALLY BE CREATED)
- Attempt to create a ticket on behalf of a local authority user

### DIFF
--- a/app/views/support_tickets/academy_details.html.erb
+++ b/app/views/support_tickets/academy_details.html.erb
@@ -17,8 +17,9 @@
         Which academy trust do you work for?
       </h1>
 
-      <%= f.govuk_collection_select :academy_name, @form.multi_academy_trust_list,
-                                    :name, :name, label: {text: 'Academy trust name', size: 's'} %>
+      <%= f.govuk_collection_select :academy_name, @form.multi_academy_trust_list, :name, :name,
+                                    label: {text: 'Academy trust name', size: 's'},
+                                    options: { include_blank: true }%>
 
       <%= f.govuk_submit 'Next' %>
     <% end %>
--- a/app/views/support_tickets/academy_details.html.erb
+++ b/app/views/support_tickets/academy_details.html.erb
@@ -17,7 +17,8 @@
         Which academy trust do you work for?
       </h1>
 
-      <%= f.govuk_text_field :academy_name, label: {text: 'Academy trust name', size: 's'} %>
+      <%= f.govuk_collection_select :academy_name, @form.multi_academy_trust_list,
+                                    :name, :name, label: {text: 'Academy trust name', size: 's'} %>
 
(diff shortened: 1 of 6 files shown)


app/views/support_tickets/academy_details.html.erb
```diff
@@ -17,7 +17,9 @@
         Which academy trust do you work for?
       </h1>
 
-      <%= f.govuk_text_field :academy_name, label: {text: 'Academy trust name', size: 's'} %>
+      <%#= f.govuk_text_field :academy_name, label: {text: 'Academy trust name', size: 's'} %>
+      <%= f.govuk_collection_select :academy_name, SupportTicket::AcademyDetailsForm::OPTIONS,
```
benilovj, line 21:
```suggestion
      <%= f.govuk_collection_select :academy_name, f.multi_academy_trusts,
```

and then add this method to the form object (because an ActiveRecord scope in a constant feels icky):

```ruby
def options
  Trust.gias_status_open.multi_academy_trust.order(:name)
end
```

app/webpacker/scripts/responsible-bodies-autocomplete.js
```diff
@@ -4,7 +4,8 @@ const initResponsibleBodiesAutocomplete = () => {
   try {
     const inputIds = [
       "#support-user-responsible-body-form-responsible-body-id-field",
-      '#school-search-form-responsible-body-id-field'
+      '#school-search-form-responsible-body-id-field',
+      '#support-ticket-academy-details-form-academy-name-field'
```
aliuk2012, line 8: @benilovj I'm thinking of renaming this file because I cannot see anything other than the id's that might make this functionality strictly for RB only.

app/views/support_tickets/academy_details.html.erb
```diff
@@ -17,7 +17,10 @@
         Which academy trust do you work for?
       </h1>
 
-      <%= f.govuk_text_field :academy_name, label: {text: 'Academy trust name', size: 's'} %>
+      <%#= f.govuk_text_field :academy_name, label: {text: 'Academy trust name', size: 's'} %>
```
asmega, line 20: delete the label if it's not needed

asmega (approved):

jamesrose (approved):

PR #1123, opened 2021-01-14
Send virtual cap updates regardless of school status

jamesrose:
### Context

We allowed schools of any state to be added to virtual pools in https://github.com/DFE-Digital/get-help-with-tech/pull/945 Following that, this PR sends updates to CC for all schools within a pool regardless of their status.


tonyheadford (approved):
LGTM

PR #1125, opened 2021-01-15
Fix MNO request bulk upload crash when the contract type column contains an unexpected value

benilovj:
### Context

School and responsible body users can upload MNO requests in an XLSX spreadsheet. Currently, if there's an unexpected value in the 'Pay monthly or PAYG', the entire upload fails, instead of just skipping the problematic line.

https://trello.com/c/axgYg1oL

Ensure that just the single row fails when the value in the 'Pay monthly or PAYG' is unexpected.
Refactor the mechanism of how errors are fed back to the user, so it just bubbles up the errors from the `ExtraMobileDataRequest` model directly.

### Guidance to review

image

### DIFF
--- a/app/models/extra_mobile_data_request.rb
+++ b/app/models/extra_mobile_data_request.rb
@@ -3,7 +3,7 @@ class ExtraMobileDataRequest < ApplicationRecord
   before_save :normalise_device_phone_number
 
   belongs_to :created_by_user, class_name: 'User', optional: true
-  belongs_to :mobile_network
+  belongs_to :mobile_network, optional: true # set to optional as we already validate on the presence of mobile_network_id and we don't want duplicate validation errors
   belongs_to :responsible_body, optional: true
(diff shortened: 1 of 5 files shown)


jamesrose (approved):

app/models/extra_mobile_data_request.rb
```diff
@@ -3,7 +3,7 @@ class ExtraMobileDataRequest < ApplicationRecord
   before_save :normalise_device_phone_number
 
   belongs_to :created_by_user, class_name: 'User', optional: true
-  belongs_to :mobile_network
+  belongs_to :mobile_network, optional: true # set to optional as we already validate on the presence of mobile_network_id and we don't want duplicate validation errors
```
jamesrose, line 6: A bit of an oddity, but so is `govuk_error_summary` :)

jamesrose (approved):

PR #1126, opened 2021-01-15
Fix missing translation when a MNO request spreadsheet cannot be processed

benilovj:
### Context

A user reported seeing a 'missing translation' error when uploading an MNO requests spreadsheet that couldn't be processed.

There was an i18n string that was being used in two similar controllers. One controller was updated and the string moved, but I forgot to update the other controller, so users are seeing a 'missing translation' error.

I've added spec coverage around both controllers to prevent this from regressing in future.


jamesrose (approved):
Good spot!

PR #1128, opened 2021-01-15
[1317] Hide mno from certain schools

asmega:
### Context

- https://trello.com/c/8lpCnlCn/1317-hide-mno
- We need to hide mno from certain schools

- We have a list of schools to hide mno from
- This list was determined by GIAS data where `StatutoryLowAge` was 16 or higher
- We currently do not have this data in our system so this is a stop gap implementation until we import this data 

### Guidance to review

- Login as a school with the `hide_mno` `true`
- Navigate to and `/schools/URN` and `/schools/URN/internet`
- Should not see references to MNO but should see 4G routers


jamesrose (approved):
LGTM!

PR #1130, opened 2021-01-15
make staging serve requests on custom domain

aldavidson:
### Context

[Trello card 1314](https://trello.com/c/DY7lYtAT/1314-setup-staging-with-a-custom-domain) - we should have staging as prod-like as possible, which should include serving on a custom domain via Cloudfront.

The domain `staging-get-help-with-tech.education.gov.uk` has already been set up & validated (necessarily outside of this epo) and mapped to staging. I've tested it in a browser, but it's still sending emails with links pointing to the existing direct-to-paas URL 

* configure staging to use the new education.gov.uk subdomain as its main domain for links in emails
* to help reduce any user confusion between staging & prod, add a '(dev)' or '(staging)' suffix to the main header service name, only in those environments


jamesrose (approved):
Looks good. The suffix is definitely going to be handy. I get mixed up between dev and staging all the time 💯

PR #1132, opened 2021-01-15
Extract an ExtraMobileDataRequestStatusComponent view component

benilovj:
### Context

We show the status of MNO requests in 3 different places (MNO area, school are and RB area).

Extract an ExtraMobileDataRequestStatusComponent view component. This will make it easier to change the code, to test it and better follows our current patterns.


jamesrose (approved):
🥇

PR #1134, opened 2021-01-18
[1075] Finalise CC IDs/references

asmega:
### Context

- https://trello.com/c/R2235yoT/1075-spike-adding-fe-institutions-to-the-service

- Prefix RB with `FE` to match CC
- This will match spreadsheets provided to CC
- Turns outs School does not need `computacenter_identifier` as a remnant of a previous spike as URN/UKPRN are used instead

### Guidance to review

n/a

### DIFF
--- a/app/models/computacenter/responsible_body_urns.rb
+++ b/app/models/computacenter/responsible_body_urns.rb
@@ -50,7 +50,11 @@ module Computacenter::ResponsibleBodyUrns
         end
       when 'FurtherEducationCollege'
         # potential n+1
-        schools.first&.ukprn
+        if schools.first&.ukprn
+          "FE#{schools.first&.ukprn}"
+        else
+          ''
+        end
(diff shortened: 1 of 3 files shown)


app/models/computacenter/responsible_body_urns.rb
```diff
@@ -50,7 +50,7 @@ def computacenter_identifier
         end
       when 'FurtherEducationCollege'
         # potential n+1
-        schools.first&.ukprn
+        "FE#{schools.first&.ukprn}"
```
jamesrose, line 53:
```suggestion
        if schools.first&.ukprn
          "FE#{schools.first&.ukprn}"
        else
          ""
        end
```

I think we'll want to check for the UKPRN here so we don't just send `"FE"` - what do you think?

aldavidson (approved):
LGTM :+1:

PR #1135, opened 2021-01-18
Add problem statuses for MNO requests

benilovj:
### Context

MNO requests currently have a `queried` status, and a separate `problem` attribute for storing the specific nature of the problem.

MNO users would like to be able to bulk-update requests with specific problems.

Trello: https://trello.com/c/5R2cI9PT

Introduce additional statuses on `ExtraMobileDataRequest`, which correspond to the existing 5 problem types.
Update the 'report a problem' form to use these problem types.
Make the new statuses appear in the bulk-change drop-down on the requests table.

### Guidance to review

image

### DIFF
--- a/app/components/extra_mobile_data_request_status_component.rb
+++ b/app/components/extra_mobile_data_request_status_component.rb
@@ -4,14 +4,13 @@ class ExtraMobileDataRequestStatusComponent < ViewComponent::Base
   end
 
   def colour
-    {
-      requested: 'blue',
-      in_progress: 'yellow',
-      complete: 'green',
-      queried: 'red',
-      cancelled: 'grey',
-      unavailable: 'grey',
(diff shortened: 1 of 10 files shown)


benilovj:
What I'm still planning to do:

- add a service, which can be run from the Rails console, for the data migration of the `queried` MNO requests
- update the reporting dashboard to take the new statuses into account

fofr:
Can we remove "unavailable" from that list, that's not something the MNOs should ever see/use.

benilovj:
> Can we remove "unavailable" from that list, that's not something the MNOs should ever see/use.

Yup - fixed in 717deb1

benilovj:
@fofr @tonyheadford this is ready for review. I'll do the work to update the Support service dashboard in a follow-up PR.

tonyheadford (approved):
LGTM!

PR #1136, opened 2021-01-18
[1291] Suppress MNO bulk update error

asmega:
### Context

- https://trello.com/c/ks3uY0pJ/1291-mno-bulk-update-correctly-handle-processing-when-no-request-is-selected

- If no requests are selected supply an empty collection so it duck types

### Guidance to review

- Login as mno user
- Make sure there are some requests
- Do not select any requests in the checkboxes
- Hit update
- Should not see error
- No records should be updated


jamesrose (approved):
Thank you! 🚀

PR #1142, opened 2021-01-19
Add support ticket model so that we can better handle defaults

jamesrose:
### Context

- Introduce a model to decorate the support ticket session values. Add sensible for each field so that we fail gracefully from missing values.
- URN and UKPRN fields should only accept numbers.

### DIFF
--- a/app/models/support_ticket.rb
+++ b/app/models/support_ticket.rb
@@ -1,0 +1,41 @@
+class SupportTicket
+  def initialize(params:)
+    @params = ActiveSupport::HashWithIndifferentAccess.new(params)
+  end
+
+  def user_type
+    @params.fetch(:user_type, 'other_type_of_user')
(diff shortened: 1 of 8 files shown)


app/models/support_ticket.rb
```diff
@@ -0,0 +1,41 @@
+class SupportTicket
+  def initialize(params:)
```
aldavidson, line 2: is it worth including ActiveModel here, so it can do validations?

jamesrose: @aldavidson I didn't do this because @aliuk2012 had already created form objects for each stage that included `ActiveModel::Model`. They do validations as the user progresses through each of the forms.

aldavidson (approved):

aliuk2012 (approved):

PR #1143, opened 2021-01-19
Remove Slack event notifications

jamesrose:
### Context

Remove the code that posts notifiable events to Slack. I've kept the event models around for now because they seem more useful and could be repurposed to push events into other places later.


aldavidson (approved):
LGTM

fofr:
I guess we stopped using this when the events got flooded.

PR #1150, opened 2021-01-19
Mno extra mobile data requests bulk update status by csv

tonyheadford:
### Context
[Trello card](https://trello.com/c/w80Ydy7m/1288-mno-bulk-allow-mno-users-to-upload-a-csv-of-requests-to-update-their-statuses) Enable MNOs to bulk update their request statuses.

### Changes proposed in this pull request
Adds CSV upload and bulk update process with summary of changes made.

### Guidance to review
As a MNO with some `ExtraMobileDataRequest`s, sign in to the service and download your requests as CSV.
Update the status column of one or more and then upload the CSV through the new feature. Any valid changes to status will be saved to your requests and a summary of unchanged, updated and errors will be presented.

_Might need to update valid statuses and reflect these on the upload page as I think they might've changed since starting this?_

image
image
image

PR #1154, opened 2021-01-20
Refactoring of MNO request status labels and descriptions

benilovj:
### Context

MNO request statuses are presented to various users (school, RB and MNO users) and in various UI element contexts (on status tags, drop-downs and form options).

Add an explicit method which reflects which MNO request statuses are available to MNO users to use (they shouldn't be using `cancelled` and `unavailable` statuses).
DRY up how the various status labels are stored and retrieved, to make it easier if a label is missing or not.

### Guidance to review

This should help with #1150 and follow-up MNO improvements.

### DIFF
--- a/app/components/extra_mobile_data_request_status_component.rb
+++ b/app/components/extra_mobile_data_request_status_component.rb
@@ -18,10 +18,9 @@ class ExtraMobileDataRequestStatusComponent < ViewComponent::Base
   end
 
   def label
-    if @extra_mobile_data_request.status.start_with?('problem')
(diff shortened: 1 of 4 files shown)


benilovj:
@tonyheadford I pulled this out of #1150 to make that PR smaller/simpler.

I'm gonna do a follow-up PR around the `new`/`requested` status (that it appears as `new` to MNO users and as `requested` to school and RB users).

tonyheadford (approved):
LGTM

PR #1155, opened 2021-01-20
add summary of extra mobile data requests to support/responsible-body

aldavidson:
### Context

[Trello card 1305](https://trello.com/c/SYYWkvM3/1305-enable-viewing-requests-submitted-by-rbs-on-support-console) - we currently show a summary of `ExtraMobileDataRequest`s linked to a school in the support interface, but not for a `ResponsibleBody` 

Add a new summary list component `Support::ResponsibleBodyExtraMobileDataRequestsSummaryListComponent`
Render that component in `support/responsible_body/show` if any requests are present

### Guidance to review

Screenshot from 2021-01-20 11-52-17


app/views/support/responsible_bodies/show.html.erb
```diff
@@ -15,6 +15,15 @@
       <%= @responsible_body.name %>
     </h1>
 
+    <%- if @responsible_body.extra_mobile_data_requests.present? %>
+      <h2 class="govuk-heading-l govuk-!-margin-top-9">Extra Mobile Data Requests</h2>
```
benilovj, line 19:
```suggestion
      <h2 class="govuk-heading-l govuk-!-margin-top-9">MNO requests</h2>
```

aldavidson:
@fofr it's worth noting that the schools UI shows the same information in a slightly different way - one row of a summary list component, collapsed with summary/details:
Screenshot from 2021-01-20 12-10-58

@benilovj has suggested making the two consistent, and the simplest way of doing that would be to pull the MNO requests out of the `SummaryListComponent` in the schools UI, and make it a separate heading + SummaryListComponent like the ResponsibleBody interface above. 

Would welcome your opinion!

app/views/support/responsible_bodies/show.html.erb
fofr: Or "Mobile data requests"

aldavidson: They're referred to as "Extra mobile data requests" in the support/schools interface at the moment, hence labelling them as the same here. I'm happy to change that, but we should be consistent - so I'll need to change them both.

aldavidson: Done in eae32f0

aldavidson:
Just pushed ca7fb26 , which implements @benilovj 's suggestion about making the school UI consistent with this Responsible Body Ui. The `/support/schools/(urn)` page now looks like this*

Screenshot from 2021-01-20 16-14-45


* for schools which have both `show_mno?` true, and some ExtraMobileDataRequests present

benilovj:
Looking good so far, @aldavidson. Sorry I'm introducing merge conflicts into this, but it's necessary to get #1150 out urgently. I will go back and rebase/fix it up once those are out...

fofr:
Design looks good :+1:

jamesrose (approved):

PR #1156, opened 2021-01-20
Remove requests for specific circumstances

fofr:
### Context

Requests for specific circumstances were to allow schools to order devices for children not at school before disruption is reported – for example clinically extremely vulnerable children.

Now ordering is open for all schools, the "get devices early for a subset" feature doesn't apply.

Allocations do not cover children in reception to year 2, and children in those years are still eligible for devices. We will use public guidance to describe this process – many schools will have 0 allocations so might not be on service.

- Remove all links and templates for making a request for specific circumstances
- Leave in place "can order for specific circumstances" status for now

### Guidance to review

Check no references have been missed.

https://ghwt-design-history.herokuapp.com/push/#requesting-devices-for-specific-circumstances


jamesrose (approved):
Looks good to me 🚢

PR #1157, opened 2021-01-20
add "Requested" column to extra mobile data requests for schools & RBs

aldavidson:
### Context

[Trello card #1340](https://trello.com/c/A3ZBD6ab/1340-add-requested-column-to-list-of-mno-requests-shown-to-schools-and-rbs-order-table-by-that-column-newest-to-oldest) - Add "Requested" column to list of MNO requests shown to schools and RBs, order table by that column (newest to oldest)

* add the 'Requested' column the list of extra mobile data requests for schools & RBs
* update the specs

### Guidance to review

Screenshot from 2021-01-20 15-24-03

Note: there's a fair bit of existing duplication between the school / RB code and specs around this. In the interests of keeping this PR small and focussed, plus speedy delivery, I haven't done anything to address it in this PR - but I will log another Tech Debt ticket to refactor this.

### DIFF
--- a/app/controllers/responsible_body/internet/mobile/extra_data_requests_controller.rb
+++ b/app/controllers/responsible_body/internet/mobile/extra_data_requests_controller.rb
@@ -2,7 +2,7 @@ class ResponsibleBody::Internet::Mobile::ExtraDataRequestsController < Responsib
   before_action { render_404_unless_responsible_body_has_centrally_managed_schools(@responsible_body) }
 
   def index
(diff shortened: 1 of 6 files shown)


fofr:
Can we drop the timestamp from the column, so it's just the date? I'm not sure the exact time is necessary.

aldavidson:
> Can we drop the timestamp from the column, so it's just the date? Looking at it again I don't think the exact time is necessary.

Done in 0e27432 - screenshot updated

jamesrose (approved):

PR #1160, opened 2021-01-20
Remove the obsolete 'requested' status on MNO requests

benilovj:
### Context

#1158 introduced the `new` status on MNO requests and made it the default. The data migration has been run on production and the old `requested` status can be removed.

Remove the service class used for migrating the status.
Remove references to the `requested` status.


app/models/extra_mobile_data_request.rb
```diff
@@ -38,7 +37,7 @@ def self.problem_statuses
   end
 
   def self.statuses_available_to_mnos
-    statuses.keys - %w[cancelled unavailable requested]
+    statuses.keys - %w[cancelled unavailable]
```
tonyheadford, line 40: Are MNO's allowed to use 'new' ?

tonyheadford (approved):
LGTM - just one question

app/models/extra_mobile_data_request.rb
benilovj: You're right, they shouldn't, according to the designs. I'm gonna do a follow-up PR which will update the design for MNOs, I'll fix this as part of that PR.

PR #1165, opened 2021-01-21
Validate MNO request uniqueness within the same school or RB

jamesrose:
### Context

Card: https://trello.com/c/ZHeihTzs/1353-mnos-receiving-duplicate-requests

We were allowing duplicate requests via the manual entry method. This PR adds a validation to check whether a request is unique within the school or RB in which it is created.

### DIFF
--- a/app/models/extra_mobile_data_request.rb
+++ b/app/models/extra_mobile_data_request.rb
@@ -16,6 +16,7 @@ class ExtraMobileDataRequest < ApplicationRecord
   validates :agrees_with_privacy_statement, inclusion: { in: [true] }
 
   validate :validate_school_or_rb_present
+  validate :validate_request_uniqueness, on: :create
 
   enum status: {
     new: 'new',
@@ -77,11 +78,13 @@ class ExtraMobileDataRequest < ApplicationRecord
     notification.deliver_later
   end
 
-  def has_already_been_made?
+  def has_already_been_made?(extra_conditions = {})
     self.class.exists?(
-      account_holder_name: account_holder_name,
-      device_phone_number: device_phone_number,
-      mobile_network_id: mobile_network_id,
+      {
+        account_holder_name: account_holder_name,
+        device_phone_number: device_phone_number,
+        mobile_network_id: mobile_network_id,
+      }.merge(extra_conditions),
     )
   end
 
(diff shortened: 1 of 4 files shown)


app/models/extra_mobile_data_request.rb
```diff
@@ -16,6 +16,7 @@ class ExtraMobileDataRequest < ApplicationRecord
   validates :agrees_with_privacy_statement, inclusion: { in: [true] }
 
   validate :validate_school_or_rb_present
+  validate :validate_request_uniqueness, on: :create
```
jamesrose, line 19: There are some existing duplicates so we need to relax the constraint on update - otherwise MNOs they won't be able to update them at all (e.g., to update the status.)

benilovj (approved):

app/models/extra_mobile_data_request.rb
```diff
@@ -102,6 +105,11 @@ def validate_school_or_rb_present
     end
   end
 
+  def validate_request_uniqueness
+    scope = school_id.present? ? { school_id: school_id } : { responsible_body_id: responsible_body_id }
+    errors.add(:device_phone_number, 'duplicate request already made') if has_already_been_made?(scope)
```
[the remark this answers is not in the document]
jamesrose: Thanks! I've adopted your suggestions and sent the PR to @ralph-hawkins for a quick look.

tonyheadford (approved):
LGTM

app/models/extra_mobile_data_request.rb
```diff
@@ -102,6 +105,11 @@ def validate_school_or_rb_present
     end
   end
 
+  def validate_request_uniqueness
+    scope = school_id.present? ? { school_id: school_id } : { responsible_body_id: responsible_body_id }
+    errors.add(:device_phone_number, :duplicate) if has_already_been_made?(scope)
```
fofr, line 110: I'm not sure testing duplicates purely on phone number is right.

We don't allow users to edit requests, so if they make a mistake – eg wrong network – their only route is to do a new request – but this route is now being cut off.

My understanding of duplicate was same number, name and network.

fofr (dismissed):

app/models/extra_mobile_data_request.rb
jamesrose: @fofr `has_already_been_made?` checks that they have the same number, network, name and RB/school

config/locales/en.yml
```diff
@@ -576,6 +576,7 @@ en:
             device_phone_number:
               blank: Enter a UK mobile number, like 07700 900999
               invalid: Enter a UK mobile number, like 07700 900999
+              duplicate: A request for this number has already been made
```
fofr, line 579:
```suggestion
              duplicate: A request with these details has already been made
```

PR #1166, opened 2021-01-21
Use consistent who_will_order values

jamesrose:
### Context

We have a mix of `schools` on RB and `school` on PreorderInformation. This brings both inline to be the singular `schools`.

Unfortunately we currently have a mix on production:

```
ResponsibleBody.group(:who_will_order_devices).count
=> {"school"=>433, nil=>2, "schools"=>2607, "responsible_body"=>616}
```

### DIFF
--- a/db/migrate/20210121111312_change_responsible_body_who_can_order.rb
+++ b/db/migrate/20210121111312_change_responsible_body_who_can_order.rb
@@ -1,0 +1,5 @@
+class ChangeResponsibleBodyWhoCanOrder < ActiveRecord::Migration[6.0]
+  def change
+    ResponsibleBody.where(who_will_order_devices: 'schools').update_all(who_will_order_devices: 'school')
+  end
+end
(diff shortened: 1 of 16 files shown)


db/migrate/20210121111312_change_responsible_body_who_can_order.rb
```diff
@@ -0,0 +1,9 @@
+class ChangeResponsibleBodyWhoCanOrder < ActiveRecord::Migration[6.0]
+  def up
+    ResponsibleBody.where(who_will_order_devices: 'schools').update_all(who_will_order_devices: 'school')
+  end
+
+  def down
+    ResponsibleBody.where(who_will_order_devices: 'school').update_all(who_will_order_devices: 'schools')
```
asmega, line 7: I'd delete this line and make it non-reversible. once its melded in you can't take it apart

tonyheadford (approved):
LGTM

db/migrate/20210121111312_change_responsible_body_who_can_order.rb
jamesrose: Good idea, thanks.

asmega (approved):

PR #1168, opened 2021-01-21
Fix yarn to 1.22.10

jamesrose:
### Context

Our docker was failing because yarn requested a later node version that wasn't compatible with our packages.json. This fixes yarn for now.

This is blocking production builds.


aldavidson:
if this works, ship it :+1: 

We should revisit this in future, to make sure we're not version-locked for longer than needed

aldavidson (approved):

PR #1172, opened 2021-01-21
FE college users should have the same view as SAT users

jamesrose:
### Context

We have some code that narrows the views RB users have only a singular view of its school. We need to do the same thing for FE institutions so they see one school vs an index of schools.

- Update and rename `User#is_a_single_academy_trust_user?` to accommodate FE users so that they don't see the schools index view like normal RB users.

### Guidance to review

1. Log in as FE college user
2. See no list of schools, essentially looks as if you're a school user.


tonyheadford (approved):
Looks good

PR #1175, opened 2021-01-21
Fix the import of MNO request spreadsheets produced by Google Docs

benilovj:
### Context

School/RB users weren't able to upload valid MNO request spreadsheets that they had exported from Google Docs.

Trello: https://trello.com/c/LJXzm5w7

GDocs produces lots of trailing empty rows, as far as the `rubyXL` library is concerned, and the import code didn't do a good job of handling that.

Add spec coverage to prevent this issue from regressing in future.


fofr (approved):
❤️

PR #1181, opened 2021-01-22
removes feature flag as content applies to all education settings

ralph-hawkins:
### Context

We've added FE and had a feature flag for some participants in user research on the content of this page.

Removes feature flag that confuses users when they see allocation description that doesn't apply.

### DIFF
--- a/app/views/school/welcome_wizard/allocation.html.erb
+++ b/app/views/school/welcome_wizard/allocation.html.erb
@@ -9,7 +9,6 @@
         <%= title %>
       </h1>
 
-      <% if @school.increased_sixth_form_feature_flag? || @school.increased_fe_feature_flag? %>
         <p class="govuk-body">Your allocation of <%= @allocation %> is based on an estimate of the number of devices your school, college or <span class="app-no-wrap">further education provider</span> already has and the number of:</p>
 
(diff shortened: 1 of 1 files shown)


app/views/school/welcome_wizard/allocation.html.erb
```diff
@@ -18,17 +17,6 @@
           <li>students in 16 to 19 further education</li>
           <li>students who get free meals in further education</li>
         </ul>
-      <% else %>
-        <p class="govuk-body">Your allocation of <%= @allocation %> is based on:</p>
-
-        <ul class="govuk-list govuk-list--bullet govuk-!-margin-bottom-6">
-          <li>the number of children in years 3 to 11</li>
```
asmega, line 25: are we sure we want to delete this? it will now always say `students in 16 to 19 further education` what about if a primary school logs in?

ralph-hawkins: it still has:
- pupils in years 3 to 11
- pupils who get free school meals

asmega: oh sorry the diff cut the first 2 items in the list

asmega (approved):

PR #1182, opened 2021-01-22
fix version-lock on node/yarn/alpine-3.12

aldavidson:
### Context

See [Trello card 1372](https://trello.com/c/pNhXwfHA/1372-revisit-build-hack-try-to-unlock-the-alpine-version) - a recent alpine version-bump meant the Docker build couldn't find a compatible pair of versions for node & yarn.

It turns out the version-lock was coming from inside the build(ing) all along

* bump alpine version to 3.13, and bump the version requirement for node in yarn's `package.json` from 12.x to 14.x 

### Guidance to review

`make dev build` should complete succeffully

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ruby:2.7.2-alpine3.12
+FROM ruby:2.7.2-alpine3.13
 
 ARG APPNAME=get-help-with-tech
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ruby:2.7.2-alpine3.12
+FROM ruby:2.7.2-alpine3.13
```
jamesrose, line 1: Is it worth unfixing this back to `FROM ruby:2.7.2` as well?

jamesrose (approved):

Dockerfile
aldavidson: Good call - I've removed the alpine version number

PR #1190, opened 2021-01-25
1289 mno bulk add feature to search for multiple requests

tonyheadford:
### Context
[Trello card](https://trello.com/c/gQk4sysp/1289-mno-bulk-add-feature-to-search-for-multiple-requests)
Enable MNOs to search list of requests by telephone number

Add form to 'Your requests' page enabling MNO to add one phone number per line to search for.
Add 'results' page that has similar functionality to main 'Your requests' page - column sorting, bulk updates, download CSV etc.
From results page, downloading a CSV includes only matching requests

### Guidance to review
As a MNO user with some `ExtraMobileDataRequest`s, log in a view 'Your requests'
The 'Find requests by telephone number' form should now present.
Enter one or more phone numbers and click 'Find requests'.
Matching requests should be displayed.
You can interact with the requests the same as on the initial 'Your requests' page. The 'Find requests by telephone number' form should not be visible here.
Downloading a CSV should only include the matching requests in the CSV.

image
image

### DIFF
--- a/app/controllers/mno/extra_mobile_data_requests_controller.rb
+++ b/app/controllers/mno/extra_mobile_data_requests_controller.rb
@@ -139,7 +139,7 @@ private
     form_params = params.fetch(:mno_find_requests_form, {}).permit(:phone_numbers)
     # once on the find requests page we could get sorting requests for the table of results
     form_params = params.permit(:phone_numbers) if form_params.empty?
-    
+
     form_params
   end
 
--- a/app/controllers/mno/extra_mobile_data_requests_controller.rb
+++ b/app/controllers/mno/extra_mobile_data_requests_controller.rb
@@ -1,6 +1,7 @@
 class Mno::ExtraMobileDataRequestsController < Mno::BaseController
   def index
-    @extra_mobile_data_requests = extra_mobile_data_request_scope.order(safe_order)
+    @find_requests_form = Mno::FindRequestsForm.new(find_requests_params)
+    @extra_mobile_data_requests = extra_mobile_data_request_search_scope.order(safe_order)
 
     respond_to do |format|
       format.csv do
@@ -14,6 +15,11 @@ class Mno::ExtraMobileDataRequestsController < Mno::BaseController
           extra_mobile_data_request_ids: selected_extra_mobile_data_request_ids(@extra_mobile_data_requests, params),
         )
         @statuses = mno_status_options
+        if @find_requests_form.phone_number_list.any?
+          render 'search_results'
+        else
+          render
+        end
       end
     end
   end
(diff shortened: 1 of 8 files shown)


app/views/mno/extra_mobile_data_requests/index.html.erb
```diff
@@ -17,64 +17,32 @@
   <%= govuk_button_link_to 'Download requests as CSV', sort: params[:sort], dir: params[:dir], format: :csv %>
   <p class="govuk-body"><%= govuk_link_to t('page_titles.extra_mobile_data_requests_csv_update'), new_mno_extra_mobile_data_requests_csv_update_path %></p>
```
fofr, line 18: Can we make a small tweak here so it looks like this:

Screen Shot 2021-01-25 at 14 35 10



See this commit for reference: https://github.com/DFE-Digital/get-help-with-tech-prototype/commit/1a850a8b3097354302c61731fba9bbe3312f3707

Classes on `app-card`:
```app-card govuk-!-margin-bottom-6 govuk-!-margin-top-4```

File: app/views/mno/extra_mobile_data_requests/search_results.html.erb [hunk 2]
```diff
@@ -0,0 +1,23 @@
+<% content_for :title, t('page_titles.requests_for_extra_mobile_data') %>
+<% content_for :before_content, govuk_link_to('Back', mno_extra_mobile_data_requests_path, class: 'govuk-back-link') %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">
+      <%= pagy_info(@pagination, "requests").capitalize.html_safe %> found
```
fofr, line 7: I think this does this:

Screen Shot 2021-01-25 at 14 54 46

tonyheadford: Good catch. Done

app/views/mno/extra_mobile_data_requests/index.html.erb
tonyheadford: Spacing changed

fofr:
Design looks good and have tested locally :+1:

app/controllers/mno/extra_mobile_data_requests_controller.rb
```diff
@@ -48,6 +54,24 @@ def bulk_update
...
+      @extra_mobile_data_requests = extra_mobile_data_request_search_scope.order(safe_order)
+      @pagination, @extra_mobile_data_requests = pagy(@extra_mobile_data_requests)
+      @extra_mobile_data_requests_form = Mno::ExtraMobileDataRequestsForm.new(
+        extra_mobile_data_requests: @extra_mobile_data_requests,
+        extra_mobile_data_request_ids: selected_extra_mobile_data_request_ids(@extra_mobile_data_requests, params),
+      )
+      @statuses = mno_status_options
+      render :index
+    end
+  end
```
aldavidson, line 73: it looks like this has a lot in common with the :index method - if I'm reading this right, it either renders :index or redirects to it - would it make sense to just combine this into the index method, and make the default index page a search with no params?

config/routes.rb
```diff
@@ -80,6 +80,7 @@
     resources :extra_mobile_data_requests, only: %i[index show edit update], path: '/extra-mobile-data-requests' do
       put 'bulk-update', to: 'extra_mobile_data_requests#bulk_update', on: :collection
       get 'report-a-problem', to: 'extra_mobile_data_requests#report_problem', as: :report_problem
+      put 'find-requests', to: 'extra_mobile_data_requests#find_requests', on: :collection
```
aldavidson, line 83: I'm a little uneasy about using non-safe HTTP verbs for safe actions - IMHO it's almost the same as using safe verbs for unsafe actions - although I guess there could be too may numbers to fit into a URL param, but in that case it might be more conventional to fall back to POST. Was there a reason why you chose PUT?

tonyheadford: Agree, it should be a POST

app/controllers/mno/extra_mobile_data_requests_controller.rb
tonyheadford: I see what you mean. This really only validates the submitted telephone numbers to check that something valid has been submitted before passing those params off to the index method or rendering an error message. There is a bit of subtlety about how the whole thing fits together though that enables use of the params across the pages.  It felt like it would make the index method a bit unwieldy to do that there (and be submitting form data to the index method that isn't required but you'd like to validate makes it more complex)

aldavidson: OK, I'll approve this PR as it's not a blocker - maybe add a Tech Debt card to see if this can be refactored in the firebreak?

aldavidson (approved):

PR #1191, opened 2021-01-25
Removing flag for fe and setting types

ralph-hawkins:
- removes a flag that shows conditional content for FE providers as we're showing the same content to all users
- adds FE settings for places where we need to show the type of education organisation
- change button text for users with multiple schools to show org name instead of the word 'school'

### DIFF
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -81,6 +81,12 @@ class School < ApplicationRecord
     gias_status_open.where(computacenter_change: %w[new amended]).or(gias_status_open.where(computacenter_reference: nil))
   end
 
+  def human_for_school_type
+    I18n.t(school_type, scope: %i[activerecord attributes school school_type], default: 'Other')
+  rescue I18n::ArgumentError
+    'Other'
+  end
+
   def ukprn_or_urn
     ukprn || urn
   end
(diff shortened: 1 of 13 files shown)


app/views/school/welcome_wizard/what_happens_next.html.erb
```diff
@@ -20,7 +20,7 @@
         <li>change your Chromebook settings</li>
       </ul>
 
-      <%= f.govuk_submit (@current_user.has_multiple_schools? ? 'Finish and go to school page' : 'Finish and go to homepage' ) %>
+      <%= f.govuk_submit (@current_user.has_multiple_schools? ? "Finish and go to #{@school.name} overview" : "Finish and go to homepage" ) %>
```
fofr, line 23: Schools names can be really long, this would be a big button

ralph-hawkins: Good point. I'll switch it to institution type, thanks

app/models/school.rb
```diff
@@ -81,6 +81,12 @@ def self.requiring_a_new_computacenter_reference
     gias_status_open.where(computacenter_change: %w[new amended]).or(gias_status_open.where(computacenter_reference: nil))
   end
 
+  def human_for_school_type
+    I18n.t(school_type, scope: %i[activerecord attributes school school_type], default: 'Other')
```
jamesrose, line 85: @asmega Why do we need the rescue here if we set a default with `I18n.t`?

asmega: its when `school_type` is `` or empty string from `School#school_type` let me try making that return nil to see if the tests pass

jamesrose (approved):

PR #1201, opened 2021-01-27
Display the zendesk ticket reference after a ticket has been created

aliuk2012:
### Context
The "Thank you" page content was a bit confusing for some users and as it didn't give them a reference number there was no confidence that their issue was logged even though they would have received a confirmation email with the ticket reference.

What was happened is that when we posted the form data to Zendesk, Zendesk would respond to the post request with an "OK" status along with the newly created ticket information. We just didn't relay that information back to the user.

- Display the zendesk ticket reference number on the "Thank you" page
- Improve error handling if the action fails to create a ticket.

### Guidance to review
- visit https://dfe-ghwt-pr-1201.herokuapp.com/ 
- Create a new support ticket (For the shortest journey, select "I'm none of the above")
### Before
image
### After
image

### DIFF
--- a/app/controllers/support_ticket/check_your_request_controller.rb
+++ b/app/controllers/support_ticket/check_your_request_controller.rb
@@ -8,11 +8,12 @@ class SupportTicket::CheckYourRequestController < SupportTicket::BaseController
   end
 
   def save
-    if form.valid?
-      form.create_ticket
+    if form.valid? && form.create_ticket
       session[:support_ticket] = nil
+      session[:support_ticket_number] = form.ticket_number
       redirect_to next_step
     else
+      flash[:warning] = 'There was a problem trying to log your request. Please try again.'
       render 'support_tickets/check_your_request'
     end
   end
--- a/app/views/support_tickets/thank_you.html.erb
+++ b/app/views/support_tickets/thank_you.html.erb
@@ -1,23 +1,38 @@
-<% content_for :title,  'Thank you for completing this form' %>
+<% content_for :title,  'Support request complete' %>
 
 <%- content_for :before_content do %>
   <% breadcrumbs([{ 'Home' => guidance_page_path },
                   {t('landing_pages.get_support.start') => support_ticket_path},
-                  'Thank you for completing this form',
+                  'Support request sent',
                  ]) %>
 <% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full govuk-!-margin-bottom-5">
+    <%=render GovukComponent::Panel.new(
+      title: 'Support request sent',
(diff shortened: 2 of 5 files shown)


Lloyd-K:
Minor changes to content supplied directly.

spec/form_objects/support_ticket/check_your_request_form_spec.rb
```diff
@@ -19,16 +19,49 @@
...
+        end
+      end
     end
 
     context 'Subject' do
       it 'sets the email subject line to include just school name' do
         allow(ZendeskService).to receive(:send!)
-        described_class.new(ticket: { 'school_name' => 'School 1' }).create_ticket
+
+        klass = described_class.new(ticket: { 'school_name' => 'School 1' })
```
asmega, line 63: more of a style thing `klass` typically refers to an actual class and not an instance of a class. eg. I would expect `klass.new` to return me an instance

app/controllers/support_ticket/check_your_request_controller.rb
```diff
@@ -8,11 +8,12 @@ def new
   end
 
   def save
-    if form.valid?
-      form.create_ticket
+    if form.valid? && form.create_ticket
       session[:support_ticket] = nil
+      session[:support_ticket_number] = form.ticket_number
```
asmega, line 13: non blocker but does this need to be deleted later? or can it hang about? which brings me to a possible solution that using `flash` might be a better option

asmega (approved):

app/controllers/support_ticket/check_your_request_controller.rb
aliuk2012: I don't think this needs to be deleted because if they navigate to another page and decide to use the browser back button then the page will still load up as expected.

If they go through the get support journey again then the session value would be updated

fofr:
Nice update 👍

app/views/support_tickets/thank_you.html.erb
```diff
@@ -1,23 +1,38 @@
-<% content_for :title,  'Thank you for completing this form' %>
+<% content_for :title,  'Support request complete' %>
 
 <%- content_for :before_content do %>
   <% breadcrumbs([{ 'Home' => guidance_page_path },
                   {t('landing_pages.get_support.start') => support_ticket_path},
-                  'Thank you for completing this form',
+                  'Support request complete',
```
fofr, line 6:
```suggestion
                  'Support request sent',
```

How about sent/delivered rather than complete? cc @Lloyd-K

aliuk2012: I like that!

PR #1218, opened 2021-01-28
Allow support users to find permanently closed schools

jamesrose:
### Context

Trello: https://trello.com/c/DF6lRrlF/1437-support-improvement-be-clear-when-school-is-closed

A few months ago we decided to hide closed schools in the support console.

Since then schools users have been navigating to schools directly using the URN in the URL so that they can view details on closed schools. We do not show when a school is permanently closed so this is particularly confusing.

- This PR allows support users to navigate and search to schools that are marked as permanently closed by GIAS.
- It also adds a warning to the top of the school page to indicate if a school is closed.

### Guidance to review

- Search for closed school in the support console
- You should see results including the closed school

- Navigate to a responsible body with a closed school in support
- You should see the closed school listed on the RB

- You should see a warning at the top of the page that the school is permanently closed

### DIFF
--- a/app/controllers/support/responsible_bodies_controller.rb
+++ b/app/controllers/support/responsible_bodies_controller.rb
@@ -18,7 +18,6 @@ class Support::ResponsibleBodiesController < Support::BaseController
     @schools = @responsible_body
       .schools
       .includes(:device_allocations, :preorder_information)
-      .gias_status_open
       .order(name: :asc)
   end
 end
--- a/app/form_objects/school_search_form.rb
+++ b/app/form_objects/school_search_form.rb
@@ -25,7 +25,7 @@ class SchoolSearchForm
   end
 
   def schools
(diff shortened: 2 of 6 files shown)


tonyheadford (approved):
Looks good. Not a blocker but perhaps a separate PR, I think it would be great if we could provide a link to the successor school or indicate that the school still exists somehow - just wonder if the messaging sounds very final as if the school has completely gone rather than, in most cases, it's still exactly the same just with a different URN.

jamesrose:
@tonyheadford Yeah, I agree, that would be useful. I wonder if we could do a fuzzy match on the name and suggest 'similar open schools'?

benilovj:
> @tonyheadford Yeah, I agree, that would be useful. I wonder if we could do a fuzzy match on the name and suggest 'similar open schools'?

We have the actual links between open and closed schools in the `DataStage` tables without needing to do fuzzy matching...

jamesrose:
@benilovj Even better!

PR #1223, opened 2021-01-29
[1433] Extend users session Part 1/2

asmega:
### Context

- https://trello.com/c/MgxnyGhN/1433-support-improvement-increase-log-in-time-for-support-agents

- A two part change is needed to allow different users to have different session durations
- This creates `expires_at` on the session and populates it on next use
- It cannot be populated immediately as postgres does not allow for default values based on other data on the row hence a two part change
- Next change will modify `Session#expired?`, remove `ttl` from `Session`, update `DeleteOldSessionsJob`

### Guidance to review

- Signing in, out and expiration should behave as normal as no change has been made
- We are simply setting up new data to be added which be hooked on to later

### DIFF
--- a/app/models/session.rb
+++ b/app/models/session.rb
@@ -4,6 +4,10 @@ class Session < ApplicationRecord
     || (updated_at.utc + ttl.seconds) < Time.zone.now.utc
   end
 
+  # def expired?
+  #   expires_at < Time.zone.now.utc
+  # end
+
   def ttl
     Settings.session_ttl_seconds
   end
--- a/app/services/session_service.rb
+++ b/app/services/session_service.rb
@@ -3,6 +3,9 @@ class SessionService
   class TokenNotRecognised < StandardError; end
   class InvalidTokenAndIdentifierCombination < StandardError; end
 
+  DEFAULT_USER_TTL = 3600.seconds.freeze
+  SUPPORT_USER_TTL = 7200.seconds.freeze
+
   def self.send_magic_link_email!(email_address)
     if (user = find_user_by_lowercase_email(email_address))
       user.generate_token!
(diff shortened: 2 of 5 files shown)


fofr:
Why don't we increase for all users?

app/models/session.rb
```diff
@@ -4,6 +4,10 @@ def expired?
     || (updated_at.utc + ttl.seconds) < Time.zone.now.utc
   end
 
+  # def expired?
+  #   expires_at < Time.zone.now.utc
+  # end
+
```
aliuk2012, line 10: can we drop this method?

app/services/session_service.rb
```diff
@@ -42,11 +45,19 @@ def self.find_user_by_lowercase_email(email_address)
...
     if is_signed_in?(session)
       user = User.find(session[:user_id])
-      update_session!(session[:session_id])
+      update_session!(session[:session_id], ttl_for_user(user))
       user
     end
   end
 
+  def self.ttl_for_user(user)
+    if user.is_support?
```
jamesrose, line 54:
```suggestion
    if user.is_support? || user.is_computacenter?
```

jamesrose (approved):

asmega:
> Why don't we increase for all users?
I don't know. This will give us a mechanism to change for an arbitrary user to any defined amount. Rather than a single hard-coded value for all users

app/models/session.rb
asmega: its going to be uncommented in a later commit. i was using it as a placeholder to show what the api will be

app/services/session_service.rb
asmega: I've incorporated this suggestion now

PR #1230, opened 2021-02-01
Relax validation when importing csv

tonyheadford:
### Context
[Trello card](https://trello.com/c/o92WCGO8/1466-help-reduce-the-errors-in-name-matching-for-mno-csv-bulk-upload)
MNO having issues uploading CSVs for extra mobile data requests where we validate the name is correct. There doesn't appear to be any need for us to do this, so to help reduce the errors that creep in from cut-n-paste formatting changes etc. that can occur from MS Office we remove the validation of the name and phone number.

Remove validation of account holder and device phone number.
Allow status updates as long as the status is valid and the `id` is for an `ExtraMobileDataRequest` belonging to the MNO.

### Guidance to review
As an MNO with some requests, export the requests as CSV, change one or more statuses and modify the account holder name and/or device phone number. The CSV should import correctly and update the status of the modified records.

### DIFF
--- a/app/services/extra_mobile_data_request_status_importer.rb
+++ b/app/services/extra_mobile_data_request_status_importer.rb
@@ -11,17 +11,7 @@ class ExtraMobileDataRequestStatusImporter
     @datasource.requests do |request|
       record = @mno.extra_mobile_data_requests.find_by(id: request[:id])
       if record
-        if account_holder_valid?(record, request) &&
-            phone_number_valid?(record, request) &&
-            status_valid?(record, request)
-          if record.status != request[:status]
-            record.status = request[:status]
-            record.save!
-            summary.add_updated_record(record)
-          else
-            summary.add_unchanged_record(record)
-          end
-        end
+        update_status!(record, request)
       else
         # could not find record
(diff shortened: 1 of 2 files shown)


asmega (approved):
Looks fine

fofr:
Looks good to me.

I think the risk of removing this validation only comes if the MNO manages to accidentally disassociate their IDs somehow (ie like a row shift or something). We wouldn't pick that up and we'd continue to push updates through.

I'm surprised the numbers don't match – can't we check by using a formatter first?

tonyheadford:
> Looks good to me.
> 
> I think the risk of removing this validation only comes if the MNO manages to accidentally disassociate their IDs somehow (ie like a row shift or something). We wouldn't pick that up and we'd continue to push updates through.
> 
> I'm surprised the numbers don't match – can't we check by using a formatter first?

Not sure that the numbers are an issue, we had a discussion (Tom and devs) and thought that to expedite this and try and work with MNOs on the duplication issues this afternoon we'd drop the validation as it was thought not necessary. I guess thie phone number validation could be added back later if needed

PR #1240, opened 2021-02-02
[1470] Third line can update school address

asmega:
### Context

- https://trello.com/c/P4uNQa6B/1470-limited-support-staff-can-change-addresses

- Add change link to school address if third line role
- Add page with form to update the school address
- Page has status if the address has been picked up by CC or not based off `school#computacenter_change`
- Added `role` column to `User` table

### Screenshots

image

image

image

### Guidance to review

- Login as support user
- View a school
- Cannot see link to edit school address
- Or access page to edit school address
- Set user `role` to `third_line`
- Can see link to edit school address
- If school `computacenter_change` is `none`
- see status `ACTIVE` that CC has the current address
- update address and submit form
- click `Change` address again
- Should not see address status as `pending` as CC is yet to pickup address changes

### DIFF
--- a/app/views/support/addresses/edit.html.erb
+++ b/app/views/support/addresses/edit.html.erb
@@ -1,6 +1,7 @@
 <%- title = t('page_titles.support.addresses.edit') %>
 <%- content_for :title, title_with_error_prefix(title, @school.errors.any?) %>
 <%- content_for :before_content do %>
+  <%= govuk_link_to('Back', support_school_path(@school), class: 'govuk-back-link') %>
 <%- end %>
 
 <div class="govuk-grid-row">
@@ -9,12 +10,8 @@
       <%= title %>
     </h1>
 
-    <% if @school.computacenter_change == 'none' %>
-      <p class="govuk-inset-text">
-        <strong class="govuk-tag govuk-tag--green">ACTIVE</strong> Computacenter is using the specified address
-      </p>
-    <% else %>
-      <p class="govuk-inset-text">
+    <% unless @school.computacenter_change == 'none' %>
+      <p class="govuk-body">
         <strong class="govuk-tag govuk-tag--red">PENDING</strong> Computacenter still needs to update this address
       </p>
     <% end %>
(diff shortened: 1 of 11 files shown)


fofr:
Nice!

I wonder if the address status needs to be bubbled up to the summary view – perhaps only if the address is pending CC.

config/locales/en.yml
```diff
@@ -81,6 +81,8 @@ en:
           edit: Update responsible body
       allocation:
         adjust_allocations_for_many_schools: Adjust allocations for many schools, colleges and FE providers
+      addresses:
+        edit: Update an address
```
fofr, line 85:
```suggestion
        edit: Update address
```

app/views/support/addresses/edit.html.erb
```diff
@@ -0,0 +1,35 @@
+<%- title = t('page_titles.support.addresses.edit') %>
+<%- content_for :title, title_with_error_prefix(title, @school.errors.any?) %>
+<%- content_for :before_content do %>
+<%- end %>
+
+<div class="govuk-grid-row">
```
fofr, line 6: Can we add a `Back` link to the top of the page, which returns the user to the school detail page they came from

app/views/support/addresses/edit.html.erb
```diff
@@ -0,0 +1,35 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">
+      <%= title %>
+    </h1>
+
+    <% if @school.computacenter_change == 'none' %>
+      <p class="govuk-inset-text">
```
fofr, line 13: I don't think we need `govuk-inset-text` as well as the tag.

tonyheadford (approved):
LGTM - slightly uncomfortable with saying that CC are using the address when we don't know, especially as there seem to be quite a few support requests where they don't have the right address and we do. The 'pending' label is a great idea though.

asmega:
> LGTM - slightly uncomfortable with saying that CC are using the address when we don't know, especially as there seem to be quite a few support requests where they don't have the right address and we do. The 'pending' label is a great idea though.

ok, agreed. i was not aware of this. I'll just have the pending displayed and remove the `ACTIVE` tag

asmega:
> I wonder if the address status needs to be bubbled up to the summary view – perhaps only if the address is pending CC.

Not sure yet, let's try without and we can add it later if we find out it's useful

app/views/support/addresses/edit.html.erb
asmega: This is now just the tag and the inset has been removed

PR #1242, opened 2021-02-02
Bulk request upload should should handle non-standard xlsx MIME types

jamesrose:
### Context

Sometimes Chromebook will upload xlsx files as `application/octet-stream` instead of the default `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet`.

This PR allows for uploads of that type.

Example request: https://kibana.logit.io/app/kibana#/discover/doc/8ac115c0-aac1-11e8-88ea-0383c11b333c/logstash-2021.02.02?id=FX6bY3cB_UZj64kqjsc9


tonyheadford (approved):
LGTM!

PR #1243, opened 2021-02-03
1398 add mno request page for schools and rbs

aldavidson:
### Context

[Trello card 1398 - Add a page for each MNO request in the schools/RBs journey](https://trello.com/c/nKuzGaOX/1398-add-a-page-for-each-mno-request-in-the-schools-rbs-journey)

User research has shown that users do not understand all the MNO statuses, know where to find details about what each network offer, or know how to respond to a request with a problem.

* Add a page for showing an individual request details [as given in the design history](https://ghwt-design-history.herokuapp.com/mno-request-pages/)
* Link to that page from the list of requests in the school and RB user journeys
* Refactor the MNO offer details out of `app/views/guidance/telling_about_offer.html.erb` and into a set of per-network partials, so that they can be re-used between the guidance and the request details pages
* Update the db seeds with the current set of participating MNOs

### Guidance to review

* Note that the design prototype did not cover all possible problem statuses for an `ExtraMobileDataRequest` - I've combined `problem_incorrect_phone_number`, `problem_no_match_for_number` and `problem_no_longer_on_network` into the given `(MNO) did not recognise this number` problem description, but would welcome input from @fofr and/or @ralph-hawkins on whether or not this is appropriate.
* Now that the MNO offer details are dynamic according to the current requests `MobileNetwork`, it's possible that a `MobileNetwork` could be listed in the database but not have a corresponding offer details template for some reason. It shouldn't happen at all, of course, but it is possible, It seemed to me that the best way to handle this if it was to happen, is that the view writes an `:warning` entry to our logs and just doesn't show the offer details or the corresponding header. Other suggestions are very welcome.

### Screenshots

Screenshot from 2021-02-03 13-29-36


tonyheadford (approved):
LGTM!

fofr:
> Note that the design prototype did not cover all possible problem statuses for an ExtraMobileDataRequest - I've combined problem_incorrect_phone_number, problem_no_match_for_number and problem_no_longer_on_network into the given (MNO) did not recognise this number problem description, but would welcome input from @fofr and/or @ralph-hawkins on whether or not this is appropriate.

👍  Good call

app/components/extra_mobile_data_request_status_details_component.html.erb
```diff
@@ -0,0 +1,70 @@
...
+      <%- end %>
+    </div>
+  <%- elsif @extra_mobile_data_request.problem_not_eligible_status? %>
+    <div class="govuk-inset-text">
+      <h2 class="govuk-heading-m"><%= @extra_mobile_data_request.mobile_network.brand %> told us this account is not eligible</h2>
+
+      <p>This could be because:</p>
+      <ul class="govuk-list govuk-list--bullet">
+        <li>they are a new <%= @extra_mobile_data_request.mobile_network.brand %> customer</li>
+        <li>they do not meet <%= @extra_mobile_data_request.mobile_network.brand %>'s criteria</li>
```
fofr, line 25:
```suggestion
        <li>they do not meet <%= @extra_mobile_data_request.mobile_network.brand %>’s criteria</li>
```

app/components/extra_mobile_data_request_status_details_component.html.erb
```diff
@@ -0,0 +1,70 @@
...
+      <ul class="govuk-list govuk-list--bullet">
+        <li>confirm this person needs internet access</li>
+        <li>check if there’s another mobile account they could use</li>
+        <li>request a <%= govuk_link_to('4G wireless router instead', how_to_request_4g_wireless_routers_path) %></li>
+      </ul>
+    </div>
+  <%- elsif @extra_mobile_data_request.unavailable_status? %>
+    <div class="govuk-inset-text">
+      <h2 class="govuk-heading-m"><%= @extra_mobile_data_request.mobile_network.brand %> is not offering data increases yet</h2>
+      <p>We cannot request an increase in data from a network that’s not participating in the offer.</p>
```
fofr, line 67: I'm surprised these paragraphs don't need `govuk-body` classes on them 🤔

app/components/extra_mobile_data_request_summary_list_component.rb
```diff
@@ -0,0 +1,32 @@
...
+      {
+        key: 'Mobile number',
+        value: @extra_mobile_data_request.device_phone_number,
+      },
+      {
+        key: 'Mobile network',
+        value: @extra_mobile_data_request.mobile_network.brand,
+      },
+      {
+        key: 'Pay monthly or pay as you go',
```
fofr, line 27:
```suggestion
        key: 'Pay Monthly or Pay-as-you-go',
```

app/views/responsible_body/internet/mobile/extra_data_requests/show.html.erb
```diff
@@ -0,0 +1,27 @@
+<%- content_for :before_content do %>
+  <%= govuk_link_to('Back', responsible_body_internet_mobile_extra_data_requests_path, class: 'govuk-back-link') %>
+<% end %>
+<%- title = @request.account_holder_name %>
```
fofr, line 4: We need to be careful with this – this title can end up in Google Analytics, which would be PII.

Any suggestions for an alternative title?
- Just first name? (split on first space)
- First 5 chars of phone number?

fofr (changes requested):
This looks really good! 1 significant thing around PII ending up in GA that needs to be addressed.

app/components/extra_mobile_data_request_status_details_component.html.erb
aldavidson: I copied them from the prototype pretty much as-is - but sure, I'll add those classes (done in bb5535cc)

fofr:
> Now that the MNO offer details are dynamic according to the current requests MobileNetwork, it's possible that a MobileNetwork could be listed in the database but not have a corresponding offer details template for some reason. It shouldn't happen at all, of course, but it is possible, It seemed to me that the best way to handle this if it was to happen, is that the view writes an :warning entry to our logs and just doesn't show the offer details or the corresponding header. Other suggestions are very welcome.

I see some spurious requests for Talkmobile, giffgaff and iD Mobile on live – these don't have corresponding offers – but there is a network unavailable design: https://ghwt-design-history.herokuapp.com/images/mno-request-pages/08-unavailable.png

app/views/responsible_body/internet/mobile/extra_data_requests/show.html.erb
aldavidson: Hmm, tricky!

I think if that's a concern (GA includes IP addresses which are also classed as PII, for instance) then we need to have different values for title tag vs the H1 tag on the page. The H1 as designed seems like the right choice, but if we need a non-PII title tag then we need something that uniquely identifies that page (== that request)  but not the person it represents.

First 5 chars of phone number would likely be shared among many requests. 
Likewise first name - and we if we were splitting at a space, we could easily accidentally expose the full name if they've mis-typed, or written it in native non-latin characters like Chinese.

Maybe something like "request #(id) - (_last_ 5 chars of phone number)" ?

aldavidson:
> > Now that the MNO offer details are dynamic according to the current requests MobileNetwork, it's possible that a MobileNetwork could be listed in the database but not have a corresponding offer details template for some reason. It shouldn't happen at all, of course, but it is possible, It seemed to me that the best way to handle this if it was to happen, is that the view writes an :warning entry to our logs and just doesn't show the offer details or the corresponding header. Other suggestions are very welcome.
> 
> I see some spurious requests for Talkmobile, giffgaff and iD Mobile on live – these don't have corresponding offers – but there is a network unavailable design: https://ghwt-design-history.herokuapp.com/images/mno-request-pages/08-unavailable.png

Yep, unavailable is catered for, but this is slightly different. Just noticed I missed a crucial couple of words off the description, I should have said:
> it's possible that a MobileNetwork could be listed in the database _as available_ but not have a corresponding offer details template 

in which case it will try to render a template that doesn't exist, and the resulting error will be caught & logged, and it will not show any offer details

app/views/responsible_body/internet/mobile/extra_data_requests/show.html.erb
fofr: Our GA should be using the `anonymizeIp` setting: https://developers.google.com/analytics/devguides/collection/gtagjs/ip-anonymization

Full names are against GA's terms in https://support.google.com/analytics/answer/7686480?hl=en

`[network name] request - (07 … last 4 chars of phone number)`
example: `O2 request – (07…2342)`

We need the `07` so it's clear it's a number

aldavidson: Yep, I like that - done in 9033173d :+1: 
I've added a separate [Trello card](https://trello.com/c/5GWA2q1W/1505-anonymise-ips-in-google-analytics) to make sure the IPs are anonymised

fofr (approved):

PR #1244, opened 2021-02-03
Cleanse imported request data

tonyheadford:
### Context
[Trello card](https://trello.com/c/PpCfKbkE/1485-mno-cleanse-incoming-requests)
Extra mobile data requests imported via bulk spreadsheet have leading and trailing whitespace and embedded newline characters. This has been responsible for some uploading problems later on when MNOs export their requests as CSV and then re-upload them with status updates.

Remove leading/trailing and embedded runs of whitespace and any newline characters from data imported via the bulk spreadsheet facility

### Guidance to review
Upload extra mobile data requests where some records contain leading/trailing or embedded newlines using the bulk upload facility.  These should be stripped during import and not appear in the corresponding database records.


aldavidson (approved):
LGTM :+1:

PR #1248, opened 2021-02-04
[1452] 0 users found returns users on same domain

asmega:
### Context

- https://trello.com/c/MOZGgRVr/1452-support-improvement-return-emails-in-same-domain-if-user-doesnt-exist-on-service

- When searching for users if there are no results search for users on the same domain
- If parsing of the search term fails ie is not an email address but a name simply return no related results

### Screenshots

image

### Guidance to review

- Login as support agent
- Find search for an existing user eg `user1@example.com`
- Should return user
- Search for non-existent user eg `user2@example.com`
- Should return `user1@example.com` as related user
- Searching for normal string that returns no results should not throw an error
- eg `rgioerjgierjgioerjgiejrg`

### DIFF
--- a/app/controllers/support/users_controller.rb
+++ b/app/controllers/support/users_controller.rb
@@ -1,6 +1,4 @@
 class Support::UsersController < Support::BaseController
-  SEARCH_RESULTS_LIMIT = 100
-
   before_action :set_user, except: %i[new create search results]
   before_action { authorize User }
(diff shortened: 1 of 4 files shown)


fofr:
Design looks good 👍

aldavidson (approved):
LGTM :+1:

PR #1252, opened 2021-02-04
[1460] Note that deliveries are paused over half term

fofr:
Users can continue to order throughout half term, but deliveries will not resume until Feb 22.

Screen Shot 2021-02-04 at 18 01 11

https://trello.com/c/jH7QlXDk/1460-prepare-the-service-for-pausing-deliveries-over-half-term


emmajhf (approved):
Hi @fofr, I've tweaked this to: 'All deliveries are paused during half term, but you can still order devices. Deliveries will resume on Monday 22 February.'

PR #1255, opened 2021-02-05
Allow support agents to export zendesk macro data

aliuk2012:
### Context
Originally this feature was developed using the nodejs API client but seeing as more people are requesting this information and finding it useful, it seems like a good feature to include in the service itself so that any support agents could export this live data from Zendesk without asking me to run it for them.

- Add a new section at the bottom of the support agent home page
image

- Clicking the heading fetches and downloads the latest macro data
image

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -181,6 +181,8 @@ Rails.application.routes.draw do
       resources :schools_to_add, only: %i[index show update], param: :urn, path: '/schools-to-add'
       resources :schools_to_close, only: %i[index show update], param: :urn, path: '/schools-to-close'
     end
+    get '/zendesk-statistics', to: 'zendesk_statistics#index', as: :zendesk_statistics
+    get '/zendesk-statistics/macros', to: 'zendesk_statistics#macros', as: :zendesk_macros
     resources :responsible_bodies, only: %i[index show], path: '/responsible-bodies' do
       resources :users, only: %i[new create], controller: 'users'
     end
--- a/app/controllers/support/zendesk_statistics_controller.rb
+++ b/app/controllers/support/zendesk_statistics_controller.rb
@@ -5,8 +5,15 @@ class Support::ZendeskStatisticsController < Support::BaseController
 
   def macros
     service = ZendeskMacroExportService.new
-    respond_to do |format|
-      format.csv { send_data service.csv_generator, filename: service.filename }
+    service.csv_generator
+
+    if service.valid?
+      respond_to do |format|
+        format.csv { send_data service.data, filename: service.filename }
+      end
+    else
+      flash[:warning] = service.message
+      redirect_to support_zendesk_statistics_path
     end
   end
 end
--- a/app/controllers/support/zendesk_statistics_controller.rb
+++ b/app/controllers/support/zendesk_statistics_controller.rb
@@ -1,0 +1,12 @@
+class Support::ZendeskStatisticsController < Support::BaseController
+  before_action { authorize :support }
+
+  def index; end
+
+  def macros
+    service = ZendeskMacroExportService.new
+    respond_to do |format|
+      format.csv { send_data service.csv_generator, filename: service.filename }
+    end
+  end
+end
--- a/app/policies/support_policy.rb
+++ b/app/policies/support_policy.rb
@@ -12,6 +12,7 @@ class SupportPolicy < ApplicationPolicy
   alias_method :schools?, :readable?
   alias_method :technical_support?, :editable?
   alias_method :feature_flags?, :editable?
+  alias_method :macros?, :readable?
 
   alias_method :new?, :editable?
   alias_method :create?, :editable?
--- a/app/services/zendesk_macro_export_service.rb
+++ b/app/services/zendesk_macro_export_service.rb
@@ -1,12 +1,16 @@
 require 'csv'
 
 class ZendeskMacroExportService
+  include ActionView::Helpers::SanitizeHelper
+
+  attr_reader :data, :message
+
   class << self
     delegate :send!, to: :new
   end
 
   def csv_generator
-    CSV.generate(headers: true) do |csv|
+    @data = CSV.generate(headers: true) do |csv|
       csv << ['Category',
               'Title',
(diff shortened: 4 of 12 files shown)


app/services/zendesk_statistics_service.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+  end
+
+  def filename
+    "zendesk-macros-#{Time.zone.now.strftime('%Y%m%d_%H%M')}.csv"
+  end
+
+  def all_macros
+    @all_macros ||= client.macros.include(:usage_1h).include(:usage_24h).include(:usage_7d).include(:usage_30d)
+  end
```
jamesrose, line 54:
```suggestion
  def macros
    @macros ||= client.macros(include: [:usage_1h, :usage_24h, :usage_7d, :usage_30d]).all!
  end
```

Does the API allow this? In my suggestion I also adjusted the name and integrated the `all!` method so that you get what you expect - macros!

app/services/zendesk_statistics_service.rb
```diff
@@ -0,0 +1,85 @@
...
+
+private
+
+  def client
+    @client ||= ZendeskAPI::Client.new do |config|
+      config.url = 'https://get-help-with-tech-education.zendesk.com/api/v2'
```
jamesrose, line 60: Please could you put this in `settings.yml` as well?

app/services/zendesk_statistics_service.rb
```diff
@@ -0,0 +1,85 @@
...
+      config.username = Settings.zendesk.username
+      config.token = Settings.zendesk.token
+      config.retry = true
+
+      require 'logger'
+      config.logger = Logger.new('log/zendesk.log')
+      config.logger.level = Logger::DEBUG
```
jamesrose, line 67:
```suggestion
```

Best practice not to require within a method like that. In production these logs will be lost as they won't be picked up by logstash. Please keep them in the Rails log.

app/services/zendesk_statistics_service.rb
```diff
@@ -0,0 +1,85 @@
...
+  def format_category(title)
+    title[1, title.index(']::') - 1].strip
+  end
+
+  def format_title(title)
+    title[title.index(']::') + 3, title.length].strip
+  end
+
+  def tags(actions, tag)
+    return nil if actions.blank? || actions.select { |action| action[:field] == tag }.blank?
```
jamesrose, line 80:
```suggestion
    return if actions.blank? || actions.select { |action| action[:field] == tag }.blank?
```

`nil` is implicit here

app/services/zendesk_statistics_service.rb
```diff
@@ -0,0 +1,85 @@
...
+
+    tags = actions.select { |action| action[:field] == tag }.first.value
```
jamesrose, line 82: We do `actions.select { |action| action[:field] == tag }` twice in this method - is it worth memozing as `tags = actions.select { |action| action[:field] == tag }` at the start?

config/routes.rb
```diff
@@ -176,6 +176,8 @@
     get '/technical', to: 'home#technical_support', as: :technical_support
     get '/feature-flags', to: 'home#feature_flags', as: :feature_flags
     get '/performance', to: 'service_performance#index', as: :service_performance
+    get '/zendesk-statistics', to: 'zendesk_statistics#index', as: :zendesk_statistics
+    get '/zendesk-statistics/macros', to: 'zendesk_statistics#get_macros', as: :zendesk_macros
```
jamesrose, line 180:
```suggestion
    get '/zendesk-statistics/macros', to: 'zendesk_statistics#macros', as: :zendesk_macros
```

spec/services/zendesk_statistics_service_spec.rb
```diff
@@ -0,0 +1,67 @@
+require 'rails_helper'
+
+RSpec.describe ZendeskStatisticsService, type: :model do
```
jamesrose, line 3: I think it would be useful if we had some more tests here that cover the failure cases. I'm particlarly interested in:

1. What happens when we can't communicate with the Zendesk API?
2. What happens if the format of titles, categories and tags aren't what you expect?

app/services/zendesk_statistics_service.rb
aliuk2012: 🤔 I suppose what I could do is create a `ZendeskBase` class and let this service and the other zendesk service inherit from it. What do you think?

app/services/zendesk_statistics_service.rb
aliuk2012: As far as I can see it doesn't look like it supports an array of include keys. https://github.com/zendesk/zendesk_api_client_rb#side-loading

The nodejs version of the api was different 

```
client.macros.sideLoad =['usage_1h', 'usage_24h', 'usage_7d', 'usage_30d']

client.macros.list()
```

spec/services/zendesk_statistics_service_spec.rb
aliuk2012: Does this need to as much effort into this feature? It will only be used a handful of times by support agents/us so if it doesn't work as expected we can always inspect and fix it if need be.

app/services/zendesk_statistics_service.rb
aliuk2012: I don't think it is becase of issues when actions is nil etc. Its only a small collection that it would need to work through and filter.

spec/services/zendesk_statistics_service_spec.rb
jamesrose: I still think some failure cases would be useful here. This is particularly important when making external API calls, and we've seen connectivity failures to Zendesk on the support form.

I noticed that the client configuration has `config.retry = true` — what does this mean? How many times will it retry?

app/services/zendesk_statistics_service.rb
jamesrose: I think this kind of static URL to an external API belongs in `settings.yml`

jamesrose (approved):

PR #1256, opened 2021-02-05
Remove FE allocation interstitials

fofr:
### Context

We no longer need to show these to users now ordering is open to all FE users. These pages might clash with the current uplift messages going out.

Remove FE allocation interstitials, logic and flags

### Guidance to review

Please check my migrations.

https://trello.com/c/EFVoZytK/1510-remove-interstitial-screens-for-sixth-form-and-the-fe-providers-already-using-the-service


db/migrate/20210205132951_remove_allocation_feature_flags_from_schools.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveAllocationFeatureFlagsFromSchools < ActiveRecord::Migration[6.1]
```
jamesrose, line 1: Please could you delete the columns in separate PR that is deployed later? Our migrations are ran _before_ the application is updated so some users may see errors. The database wouldn't have the field but the code would be expecting it.

fofr: Migration removed

jamesrose (approved):

PR #1258, opened 2021-02-05
Adds guidance to on checking what to do and remove 'do not delay' message

ralph-hawkins:
### Context

We’ve noticed that a few users are submitting their orders and then having to amend or cancel them.

- add a few lines to help users check device specification and invite others to order
- remove the 'Do not delay' message as this may be causing users to feel its better to place and order before they're certain of what they need


aldavidson (approved):
LGTM :+1:

PR #1265, opened 2021-02-08
Gias updates

tonyheadford:
### Context
[Trello card](https://trello.com/c/ufcDwSN6/1413-automate-more-gias-updates-where-possible)
Add some support front-end visibility and automation for adding and closing schools

Add a 'GIAS updates' area to the support area, visible only to 3rd-line support roles.
List all schools waiting to be added to the service
See detail of school to be added, along with any "school links" and button to add school
List all schools that are waiting to be closed
See detail of school to close along with any "school links" and button to close school.
Adding a school will create new school, set up preorder information if responsible body has answered the "who will order devices" question, move any remaining allocation (std and coms) to new school from predecessors and adjust totals.  Move users from predecessors to new school. Mark predecessors as closed.

### Guidance to review
Given there are some GIAS updates
Sign in as a support user with a third-line role
You should see a 'GIAS updates' link on the support home page
Visiting that link will show links with totals for adding and closing schools
From there you will see lists of schools to be added / closed
Clicking on a school shows the DataStage information for the school and the predecessor/successor school links. These will have 'View' actions if they are available on the system or in the DataStage area.
Clicking on the 'Add School' button will add the school and move allocations and users from any predecessors and close the predecessor (if available on the service) and remove it from the list.
Clicking the 'Close school' button will mark the school as closed and removed it from the list.

image
image
image
image
image
image
image
image
image
image


fofr:
Would an Add all or Close all feature be useful?

tonyheadford:
> Would an Add all or Close all feature be useful?

Perhaps, generally if we stay on top of these there wouldn't be many to do and it might be beneficial to look at them one by one but it could be useful. Would you place that button under the title before the table?

aldavidson:
> > Would an Add all or Close all feature be useful?
> 
> Perhaps, generally if we stay on top of these there wouldn't be many to do and it might be beneficial to look at them one by one but it could be useful. Would you place that button under the title before the table?

....or checkboxes with a 'select all' option?

app/form_objects/support/gias_info_form.rb
```diff
@@ -0,0 +1,29 @@
+class Support::GiasInfoForm
```
aldavidson, line 1: I'm not sure about this being a FormObject - it doesn't submit any data, doesn't have any attributes, and doesn't have any need for validations. Would it be better to call this something else (maybe a Presenter?) or even just call the wrapped methods directly on the SchoolUpdateService from the controller?

tonyheadford: Good shout. Originally though this might need to hold search params etc.

tonyheadford:
> > > Would an Add all or Close all feature be useful?
> > 
> > 
> > Perhaps, generally if we stay on top of these there wouldn't be many to do and it might be beneficial to look at them one by one but it could be useful. Would you place that button under the title before the table?
> 
> ....or checkboxes with a 'select all' option?

Good ideas. Discussed with @ralph-hawkins earlier and we felt because we cannot guarantee success (typically missing RBs or RB name issues) we'd have to provide additional means to show the results of the operation and list failures etc.  Decided to keep it simple for now and we can iterate later if necessary.

aldavidson (approved):
I think this is much cleaner now :+1: :shipit:

PR #1266, opened 2021-02-08
[1515] Stage SPIs from GIAS

asmega:
### Context

- https://trello.com/c/4h5TboLZ/1515-able-to-onboard-fe-schools-from-gias
- First step to get SPIs (Speical Post 16 institutions) from GIAS into the service
- Second step will be to move from staging area to actual records

- No longer exclude SPIs when staging schools

### Guidance to review

-


asmega:
closing as no longer needed

PR #1281, opened 2021-02-10
Signed in users shouldn't need to complete "Contact details" page of "Get support" journey

aliuk2012:
### Context
If a user is not signed in and they try to contact us using the contact form, they should not notice any changes.

If a user is signed in and they try to contact us using the contact form, they go through exactly the same journey as they would before except they will not be asked to provide their contact details.

### DIFF
--- a/app/controllers/support_ticket/contact_details_controller.rb
+++ b/app/controllers/support_ticket/contact_details_controller.rb
@@ -2,15 +2,27 @@ class SupportTicket::ContactDetailsController < SupportTicket::BaseController
   before_action :require_support_ticket_data!, only: :new
 
   def new
-    @form ||= SupportTicket::ContactDetailsForm.new(set_params)
-    render 'support_tickets/contact_details'
+    if current_user.id.present?
+      session[:support_ticket].merge!({
+        full_name: helpers.sanitize(current_user.full_name),
+        email_address: helpers.sanitize(current_user.email_address),
+        telephone_number: helpers.sanitize(current_user.telephone),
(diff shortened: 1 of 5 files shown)


app/controllers/support_ticket/contact_details_controller.rb
```diff
@@ -2,15 +2,27 @@ class SupportTicket::ContactDetailsController < SupportTicket::BaseController
   before_action :require_support_ticket_data!, only: :new
 
   def new
-    @form ||= SupportTicket::ContactDetailsForm.new(set_params)
-    render 'support_tickets/contact_details'
+    if current_user.id.present?
```
aldavidson, line 5:
```suggestion
    if current_user.persisted?
```

app/views/support_tickets/check_your_request.html.erb
```diff
@@ -40,14 +40,15 @@
...
 
-      <% component.slot(
-         :row,
-         key: "How can we contact you?",
-         value: "#{ @support_ticket.full_name}<br>#{ @support_ticket.email_address}<br>#{ @support_ticket.telephone_number}".html_safe ,
-         action: govuk_link_to("Change", support_ticket_contact_details_path),
-         classes: "Name".to_s.downcase.gsub(/ /, "-"),
-       ) %>
-
+      <% if current_user.id.nil? %>
```
aldavidson, line 43:
```suggestion
      <% unless current_user.persisted? %>
```

aldavidson (approved):
Looks good, just a couple of minor non-blocking stylistic suggestions :+1:

PR #1289, opened 2021-02-11
updates banner with half-term dates and techsource downtime message

ralph-hawkins:
### Context

TechSource will stop deliveries over half-term
TechSource will have downtime Saturday 13 February

Update half-term banner content to let users know

### Guidance to review

half_term_delivery_suspension flag needs to be active


asmega (approved):

Lloyd-K (commented):
Approved

PR #1294, opened 2021-02-11
1410 rb access flag for cc part 2

tonyheadford:
### Context
[Trello card](https://trello.com/c/EWxSCmOl/1410-investigate-putting-an-rb-access-flag-on-all-sat-users)
[Part 1 PR](https://github.com/DFE-Digital/get-help-with-tech/pull/1288)

Part 2 to tie in passing the `rb_level_access` flag to Computacenter to indicate that a user should have RB level access at Techsource.

There is no front-end exposure of this attribute so far. I think this might get complex as we expose and edit user info in several places so it will need design/content input. The value is also dependent on the `orders_devices` value as it will only be passed to CC when that is set and the user has a techsource account and has seen the privacy statement.

Adds the `cc_rb_user` and `original_cc_rb_user` to the `Computacenter::UserChange` model
Passes that value to the CC API with user changes as `u_rb_user` and `u_original_rb_user`.
Sets the `User.rb_level_access` to `true` when creating a responsible body user in the `CreateUserService`
Adds `:rb_level_access_notification` feature flag around exposing the `Computacenter::UserChange` attributes to Computacenter.

### Guidance to review
Create a user that orders devices and has seen the privacy notice.
Set the `:rb_level_access_notification` feature flag
Observe that `cc_rb_user` and `original_cc_rb_user` are populated in `Computacenter::UserChange.latest_for_user(user)`
Changing the `rb_level_access` will generate new UserChanges and populate the `cc_rb_user` values accordingly

### DIFF
--- a/app/jobs/notify_computacenter_of_latest_change_for_user_job.rb
+++ b/app/jobs/notify_computacenter_of_latest_change_for_user_job.rb
@@ -1,11 +1,6 @@
 class NotifyComputacenterOfLatestChangeForUserJob < ApplicationJob
   queue_as :default
 
-  # NOTE: we only pass the allocation_ids, not the amounts to update the caps to.
-  # This way, the new caps get read at the time the job is performed rather than
-  # when it was scheduled - so if there are several jobs queued up, it doesn't
-  # matter if the jobs get processed in the right order or not, the last job
-  # processed will always set the latest value.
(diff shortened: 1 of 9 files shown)


aldavidson (approved):
Looks good :+1:

jamesrose (approved):

PR #1306, opened 2021-02-17
Update DM breadcrumbs and copy

jamesrose:


app/views/school/donated_devices/interest/opted_in.html.erb
```diff
@@ -13,9 +13,9 @@
       <h1 class="govuk-panel__title">
         <%= title %>
       </h1>
-      <p class="govuk-panel__body">
+      <div class="govuk-panel__body">
```
jamesrose, line 16: @tonyheadford This fixes a spacing issue. The RB version should use the same.

PR #1308, opened 2021-02-17
[1548] FE MNO

asmega:
### Context

- https://trello.com/c/LlvyVpmg/1548-mno-rollout-to-fe-by-weds-17th

- Adds MNO to FEs
- `School#hide_mno?` has now been re-purposed. Previously these schools could not see the MNO, they now can. Instead the flag is used to determine if only a subset of mobile networks are permitted
- On manual requests remove excluded networks for FE
- On bulk upload show errors if user attempt to upload FE excluded network. we should be rejected and not persisted 

### Screenshots

image

### Guidance to review

- Probably want to side by side compare current school journey vs FE journey
- For FE flag is `hide_mno` set to true. as there are some FE with `hide_mno` false as they have under 16s
- Perform manual request
- School journey should see all available networks
- FE journey should see all networks minus giffgaff and o2
- Perform bulk upload
- School journey should accept all available networks
- FE should not accept giffgaff and o2 - see screenshot above
- There are also various copy changes thru the journeys

### DIFF
--- a/app/models/mobile_network.rb
+++ b/app/models/mobile_network.rb
@@ -14,7 +14,7 @@ class MobileNetwork < ApplicationRecord
     participating
   end
 
-  scope :excluded_fe_networks, -> { where(brand: %w[giffgaff O2]) }
+  scope :excluded_fe_networks, -> { where(brand: %w[giffgaff O2 Three SMARTY]) }
   scope :fe_networks, -> { participating.where.not(id: excluded_fe_networks.pluck(:id)) }
 
   def pathsafe_brand
--- a/app/models/mobile_network.rb
+++ b/app/models/mobile_network.rb
@@ -14,6 +14,9 @@ class MobileNetwork < ApplicationRecord
     participating
   end
 
+  scope :excluded_fe_networks, -> { where(brand: %w[giffgaff O2]) }
+  scope :fe_networks, -> { participating.where.not(id: excluded_fe_networks.pluck(:id)) }
+
   def pathsafe_brand
     brand.to_s
          .downcase
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -74,6 +74,18 @@ class School < ApplicationRecord
     type == 'FurtherEducationSchool'
   end
 
+  def available_mobile_networks
+    if hide_networks_not_supporting_fe?
+      MobileNetwork.fe_networks
+    else
+      MobileNetwork.participating
+    end
+  end
+
+  def hide_networks_not_supporting_fe?
+    hide_mno?
+  end
+
   def has_ordered?
     device_allocations.to_a.any? { |alloc| alloc.devices_ordered.positive? }
   end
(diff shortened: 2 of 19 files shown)


app/views/shared/internet/_mno_guidance.html.erb
```diff
@@ -9,9 +9,12 @@
   <li>cannot afford additional data</li>
 </ul>
 
+<p class="govuk-body">giffgaff, O2 and Tesco Mobile will not accept requests for students over the age of 16.</p>
```
fofr, line 12:
```suggestion
<p class="govuk-body">giffgaff and O2 will not accept requests for students over the age of 16.</p>
```

asmega: good spot i will change to read from model instead

app/models/school.rb
```diff
@@ -74,6 +74,14 @@ def is_further_education?
     type == 'FurtherEducationSchool'
   end
 
+  def available_mobile_networks
+    if hide_mno?
```
fofr, line 78: I wonder if we should have something like this?

```ruby
def hide_networks_not_supporting_fe?
  hide_mno?
end```

`hide_mno` doesn't do the heavy lifting in explaining

File: app/views/school/internet/mobile/bulk_requests/new.html.erb [hunk 3]
```diff
@@ -28,6 +28,15 @@
...
     %>
+
+    <div class="govuk-warning-text">
+      <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
+      <strong class="govuk-warning-text__text">
+        <span class="govuk-warning-text__assistive">Warning</span>
+        <%= MobileNetwork.excluded_fe_networks.pluck(:brand).to_sentence(last_word_connector: ' and ') %> will not accept requests for students over the age of 16
+      </strong>
+    </div>
+
```
fofr, line 39: This block should only show to schools/colleges etc that we know have students over 16

app/models/school.rb
asmega: yea probably a good call. i did consider renaming the entire column. Aliasing seems a good middle ground

app/views/school/internet/mobile/bulk_requests/new.html.erb
asmega: this can't be done at the moment and we will have to show for all. we simply do not have the data for it. there is a backlog ticket titled `import `StatutoryLowAge` from GIAS`. if we do that first we can use that as the toggle for this

app/models/mobile_network.rb
```diff
@@ -14,6 +14,9 @@ def self.participating_in_pilot
     participating
   end
 
+  scope :excluded_fe_networks, -> { where(brand: %w[giffgaff O2 Three SMARTY]) }
```
tonyheadford, line 17: This feels a bit icky, I think it would've been better as a flag/data item on the model?

tonyheadford (approved):
Looks good, I've made a comment about the `excluded_fe_networks` but don't want that to hold this up.

app/models/mobile_network.rb
asmega: sure, this works so i'll merge this now and make the change once everything is sorted.

PR #1309, opened 2021-02-17
1571 iterate support console for mno requests

aldavidson:
### Context

[Trello card 1571](https://trello.com/c/3SK2pEbr/1571-iterate-support-console-for-mno-requests) - We need a way for support users to be able to find individual mobile data requests that they have been contacted about by the users, whilst not exposing personally-identifying information if it is at all possible to avoid doing so.

In discussion with @fofr we agreed to try a limited redaction that would just expose (say) the first character of account holder name, and the last 4 digits of the mobile number.  

* add a new top-level support section to find & list mobile data requests
* add a basic search form to allow filtering of the list
* add a policy to only allow support users to view this list - explicitly _excluding_ any Computacenter users who may also have read-only access to support

I chose to do this as a top-level section so that later, we can link directly into it from a school or RB page in the support UI, passing the relevant search params to show specific subsets of requests

### Guidance to review

I've included all points of design feedback, and added the 'Request ID' search field, so that users can go straight a particular request ID. The specs are now completed and clean, so I think this is ready to go. Un-drafted.

Screenshots:
Screenshot from 2021-02-22 10-57-05

### DIFF
--- a/app/views/support/extra_mobile_data_requests/index.html.erb
+++ b/app/views/support/extra_mobile_data_requests/index.html.erb
@@ -1,0 +1,58 @@
+<%- title = t('page_titles.support_extra_mobile_data_requests') %>
+<%- content_for :title, title %>
+<%- content_for :before_content do %>
+  <% breadcrumbs([
+                   { "Home" => support_home_path },
+                   title
+                 ]) %>
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">
+      <%= title %>
+    </h1>
+
+    <% if @requests.any? %>
+      <%= render partial: 'shared/internet/help_with_statuses', locals: { statuses_with_descriptions: @statuses_with_descriptions } %>
+    <% end %>
+  </div>
+</div>
+
+<% if @requests.any? %>
+  <p class="govuk-body">Showing <%= 'request'.pluralize(@requests.count) %> <%= @pagination.from %> - <%= @pagination.to %> of <%= @pagination.count %></p>
+  <table class="govuk-table requests">
+    <thead class="govuk-table__head">
+      <tr class="govuk-table__row">
+        <th class="govuk-table__header">Request ID</th>
+        <th class="govuk-table__header">Account holder</th>
+        <th class="govuk-table__header">Mobile number</th>
+        <th class="govuk-table__header">Requested</th>
+        <th class="govuk-table__header">Mobile Network</th>
+        <th class="govuk-table__header">Status</th>
+      </tr>
+    </thead>
+    <tbody class="govuk-table__body">
+      <%- for emd_request in @requests do %>
+        <tr class="govuk-table__row" id="request-<%= emd_request.id %>">
+          <td class="govuk-table__cell"><%= emd_request.id %></td>
+          <td class="govuk-table__cell"><%= redact(emd_request.account_holder_name, first: 1) %></td>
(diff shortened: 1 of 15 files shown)


fofr:
Can we show the first two digits of the number too? They should all be `07...`, but it sets the context that this is a mobile number. I think we also do this in the page title.

app/views/support/extra_mobile_data_requests/index.html.erb
```diff
@@ -0,0 +1,85 @@
...
+    </thead>
+    <tbody class="govuk-table__body">
+      <%- for emd_request in @requests do %>
+        <tr class="govuk-table__row" id="request-<%= emd_request.id %>">
+          <td class="govuk-table__cell"><%= emd_request.id %></td>
+          <td class="govuk-table__cell"><%= redact(emd_request.account_holder_name, first: 1) %></td>
+          <td class="govuk-table__cell"><%= redact( emd_request.device_phone_number, last: 4 ) %></td>
+          <td class="govuk-table__cell">
+            <%= emd_request.created_at.to_date.to_s(:long_ordinal) %>
+            by
```
fofr, line 68:
```suggestion
            <br />by
```

fofr:
Can we make the column order be:

`ID, Mobile Network, Name, Number, etc.`

We can show the full network details, so I'd like to front load that.

The account name feels a little odd with just one character.

lib/string_utils.rb
```diff
@@ -8,4 +8,8 @@ def split_string(string, limit:)
     end
     [words1.join(' '), (words2.join(' ')[0...limit])]
   end
+
+  def redact(string, first: 0, last: 0, redaction: '...')
```
fofr, line 12:
```suggestion
  def redact(string, first: 0, last: 0, redaction: '…')
```

Can we use the ellipsis character everywhere, rather than three dots.

fofr:
I think the Search requests form might need to be a bit more obvious that you don't need to fill in _all the fields_. @ralph-hawkins any suggestions?

fofr:
I think this is a really good start 💯

aldavidson:
> I think the Search requests form might need to be a bit more obvious that you don't need to fill in _all the fields_. @ralph-hawkins any suggestions?

I've added a short bit of text explaining that you can fill in as many or as few fields as you like. All edits welcome!

ralph-hawkins:
> > I think the Search requests form might need to be a bit more obvious that you don't need to fill in _all the fields_. @ralph-hawkins any suggestions?
> 
> I've added a short bit of text explaining that you can fill in as many or as few fields as you like. All edits welcome!

Sorry, missed that originally.

Looks good — can we change 'All fields are optional - you can fill in as many or few as you like' to 'Search by 1 or more piece of information'.

And it looks like the dates are ordinal numbers, can we remove the 'th's and 'nds'?

aldavidson:
> > > I think the Search requests form might need to be a bit more obvious that you don't need to fill in _all the fields_. @ralph-hawkins any suggestions?
> > 
> > 
> > I've added a short bit of text explaining that you can fill in as many or as few fields as you like. All edits welcome!
> 
> Sorry, missed that originally.
> 
> Looks good — can we change 'All fields are optional - you can fill in as many or few as you like' to 'Search by 1 or more piece of information'.
> 
> And it looks like the dates are ordinal numbers, can we remove the 'th's and 'nds'?

OK, I've changed the text, and used the `:govuk` format for the dates :+1:

ralph-hawkins:
👍  looks good to me

ralph-hawkins (commented):
looks good to me 👍

tonyheadford (approved):
Looks great, nice one!

PR #1313, opened 2021-02-18
[1597] Add test personas for heroku

asmega:
### Context

-  https://trello.com/c/cNeHTPti/1597-add-personas-to-heroku-environments-for-user-testing

- Modify heroku setup to start sidekiq worker. Since I added the flash messaging afterwards this change isn't actually needed but I decided to leave it as it maybe useful in the future.
- Heroku nows loads test personas into the database
- Users can then login as these personas in the heroku review apps
- Sign in links are then presented back as flash messages

### Guidance to review

- View the review app of this PR https://dfe-ghwt-pr-1313.herokuapp.com/
- Sign in as the different 3 personas `support.user.1@example.com`, `school.user.1@example.com`, `trust.user.1@example.com`

### DIFF
--- a/app/controllers/sign_in_tokens_controller.rb
+++ b/app/controllers/sign_in_tokens_controller.rb
@@ -46,7 +46,7 @@ class SignInTokensController < ApplicationController
     if @sign_in_token_form.email_is_user?
       token = SessionService.send_magic_link_email!(@sign_in_token_form.email_address)
 
-      if ENV['HEROKU'] == 'heroku'
+      if FeatureFlag.active?(:display_sign_in_token_links)
         user = User.find_by(email_address: @sign_in_token_form.email_address)
         identifier = user.sign_in_identifier(user.sign_in_token)
         flash[:success] = validate_sign_in_token_url(token: token, identifier: identifier)
--- a/app/controllers/sign_in_tokens_controller.rb
+++ b/app/controllers/sign_in_tokens_controller.rb
@@ -45,6 +45,13 @@ class SignInTokensController < ApplicationController
 
     if @sign_in_token_form.email_is_user?
       token = SessionService.send_magic_link_email!(@sign_in_token_form.email_address)
+
+      if ENV['HEROKU'] == 'heroku'
+        user = User.find_by(email_address: @sign_in_token_form.email_address)
+        identifier = user.sign_in_identifier(user.sign_in_token)
+        flash[:success] = validate_sign_in_token_url(token: token, identifier: identifier)
+      end
+
       redirect_to sent_token_path(token: token)
     else
       redirect_to email_not_recognised_path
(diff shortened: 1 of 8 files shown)


app/controllers/sign_in_tokens_controller.rb
```diff
@@ -45,6 +45,13 @@ def create
 
     if @sign_in_token_form.email_is_user?
       token = SessionService.send_magic_link_email!(@sign_in_token_form.email_address)
+
+      if ENV['HEROKU'] == 'heroku'
```
aldavidson, line 49: Could this be a `FeatureFlag` that spells out the behaviour it governs,  rather than a reference to a hosting environment? 
Something like `FeatureFlag.active?(:display_sign_in_token_links)` or similar?

aldavidson (changes requested):
Code-wise all fine, but I'm just a bit nervous about this functionality being behind a seemingly-innocent-sounding name ('HEROKU'). I'd prefer it to be more explicit & obvious about what it does

app/controllers/sign_in_tokens_controller.rb
aliuk2012: Feature flag sounds good and I like the idea of using a config setting, but one thing that is concerning me with using that approach here is that we don't want someone to accidentally enable the feature flag and expose the whole service. So maybe it shouldn't be seen as another feature flag and the dev should go out of their way to set it?

- if we want to use config settings perhaps it should be something more generic so that we can reuse it for other features we might want to only be allowed to use/see in review apps. Maybe something along the lines of `FeatureFlag.active?(:review_app_enabled)`?

asmega: So I think both your comments are polar opposites. I'm with Mr. Davidson on this one, as I think one is less likely to make a mistake if the intent is explicit and direct as opposed to vague and indirect.

asmega: I've added this as a new commit. it now uses the `FeatureFlag` service and explicitly named. I've also added some coverage for this

aldavidson (approved):
Looks good - thanks!

aliuk2012 (approved):
Thanks @asmega this is going to be a game changer to be able to use review apps during code review or prototyping. I know I'll be using this for local dev as well.

Only suggestion I could make and it doesn't need to be in this PR but perhaps a mention of this in the readme along with a list of personas we could use would be good for new devs/designers.

asmega:
> Thanks @asmega this is going to be a game changer to be able to use review apps during code review or prototyping. I know I'll be using this for local dev as well.
> 
> Only suggestion I could make and it doesn't need to be in this PR but perhaps a mention of this in the readme along with a list of personas we could use would be good for new devs/designers.

I don't think I want to add to the README yet because it only just got added. I think it needs a few uses and iterations so it stabilises. As these current persona i just knocked up with little thought.

PR #1319, opened 2021-02-19
fix edge case in occasional spec fail

aldavidson:
### Context

I saw an intermittent spec failure due to what looks like a time rounding issue:

```
1) DonatedDeviceRequestsExporter when given a filename includes the request information in the CSV file
     Failure/Error: expect(record.school.created_at.to_s).to eq(request['created_at'])
       expected: "2021-02-19 16:08:52 +0000"
            got: "2021-02-19 16:08:51 +0000"
```

Although the error goes away on re-run, it shouldn't be happening in the first place

* change the comparison to use `be_within(1.second).of(...)` to allow for the very occasional rounding edge case


tonyheadford (approved):
Good catch

PR #1327, opened 2021-02-23
Guidance for people who don't have access to or cant use our service

ralph-hawkins:
### Context

[Trelllo card](https://trello.com/c/YGRbFY6Q/1405-support-users-who-cannot-sign-in-to-the-service)

The support team have data on high volumes of request coming from users who can't access the service.

Improve content to give clarity to the existing sign-in flow.
Adds a page to give guidance for users who could have an account but don't
Adds a page for users who can't access the service.

### Guidance to review

Sign in using an email address that the service doesn't recognise.

localhost_3000_sign-in (2)

localhost_3000_token_email-not-recognised (3)

localhost_3000_how-to-access-the-get-help-with-technology-service (3)

localhost_3000_what-to-do-if-you-cannot-get-laptops-tablets-or-internet-access-from-dfe (3)

### DIFF
--- a/app/views/sign_in_tokens/email_not_recognised.html.erb
+++ b/app/views/sign_in_tokens/email_not_recognised.html.erb
@@ -6,19 +6,20 @@
       <%= t('page_titles.email_not_recognised') %>
     </h1>
     <p class="govuk-body">
-      We did not recognise the email address you entered.
-      You may have mistyped it, or you may not have access.
+      If you mistyped your email address, <%= govuk_link_to 'try to sign in again', sign_in_path %>.
     </p>
-    <p class="govuk-body">
-      <%= govuk_button_link_to 'Try again', sign_in_path %>
-    </p>
-    <p class="govuk-body">
-      If you think your email address is correct and you should have access,
-      you can <%= govuk_link_to "contact our support team", support_ticket_path %>.
+
+    <h2 class="govuk-heading-m">
+      How to access this service to order laptops, tablets and internet connectivity
+    </h2>
+    <p class="govuk-body govuk-!-margin-bottom-6">
+      This service is for schools, colleges, <span class="app-no-wrap">FE institutions</span>, trusts and local authority education teams only. If you work for one of these institutions, <%= govuk_link_to 'find out how to get access', how_to_access_the_get_help_with_technology_service_path %>.
     </p>
     <h2 class="govuk-heading-m">
-      Parents, carers, pupils and students
+      Help for parents, pupils, social care workers, independent schools, and schools outside of England
     </h2>
-    <p class="govuk-body">Parents, carers, pupils and students do not have access to this service. You should contact your school, college or further education institution for help accessing remote education.</p>
+    <p class="govuk-body">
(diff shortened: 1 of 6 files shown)


emmajhf:
Looks great @ralph-hawkins. My only suggested change is to the 'We didn't recognise that email address' page: I'd make the en-dash into a comma, ie 'If you mistyped your email address, try to sign in again'.

emmajhf (approved):

app/views/sign_in_tokens/email_not_recognised.html.erb
```diff
@@ -6,19 +6,20 @@
       <%= t('page_titles.email_not_recognised') %>
     </h1>
     <p class="govuk-body">
-      We did not recognise the email address you entered.
-      You may have mistyped it, or you may not have access.
+      If you mistyped your email address, <%= govuk_link_to 'try to sign in again', sign_in_path %>.
```
aliuk2012, line 9: Not sure if you saw my last comment I left about this line, but I think we should play back the email address they supplied otherwise there's no way for them to double check what was submitted. They maybe a registered user but just had a typo in the email address they supplied . so by playing it back it gives them the opportunity to check why it didn't allow them access.

Perhaps something along the lines of "The email address you supplied was x, if this is incorrect please try again"( with a link back to the sign in page. It also shows empathy by not not explicitly calling the user out and blaming them for supplying the wrong email address.

app/views/sign_in_tokens/email_not_recognised.html.erb
```diff
@@ -6,19 +6,20 @@
...
+      This service is for schools, colleges, <span class="app-no-wrap">FE institutions</span>, trusts and local authority education teams only. If you work for one of these institutions, <%= govuk_link_to 'find out how to get access', how_to_access_the_get_help_with_technology_service_path %>.
     </p>
     <h2 class="govuk-heading-m">
-      Parents, carers, pupils and students
+      Help for parents, pupils, social care workers, independent schools, and schools outside of England
     </h2>
-    <p class="govuk-body">Parents, carers, pupils and students do not have access to this service. You should contact your school, college or further education institution for help accessing remote education.</p>
+    <p class="govuk-body">
+      You cannot use this service to request support. <%= govuk_link_to 'Find out what to do if you cannot get laptops, tablets or internet access from DfE.', what_to_do_if_you_cannot_get_laptops_tablets_or_internet_access_from_dfe_path %>
+    </p>
```
aliuk2012, line 23: If we are changing this should we not update the parent banner too? 
Could we not remove this and playback the parent banner as we do across the majority of guidance pages?

aliuk2012:
The changes suggested in this PR can be tested here, https://dfe-ghwt-pr-1327.herokuapp.com/

app/views/sign_in_tokens/email_not_recognised.html.erb
ralph-hawkins: I'd argue for removing the parent banner, it adds a lot of noise and repetition to the service.

By adding this content hopefully, we'll see a reduction in users being confused and can remove that banner completely.

aliuk2012: Fair point about removing the parent banner. We should check to see if we are still getting as many queries through from parents and if not then remove the banner

PR #1329, opened 2021-02-23
[1478] fix edge case allowing duplicating ExtraMobileDataRequests

aldavidson:
### Context

[Trello card 1478](https://trello.com/c/k6UsEZa6/1478-mno-prevent-duplicate-requests-part-2) - Despite there being code to look for the existence of `ExtraMobileDataRequests` for the same name & number (within the scope of school, RB, mobile network etc) we're still seeing a lot of duplicates making it through to the MNOs. This is causing them wasted time & frustration.

On a closer look, the code that normalises the phone number was running `before_save`, which is called _after_ validation. So if a user submits a phone number with spaces in, the first will be normalised on save, then the second will still pass validation as it's only looking for the exact phone number as entered, not the normalised version.
 
Change the normalising hook to `before_validation`
Add a spec for this behaviour

### DIFF
--- a/app/models/extra_mobile_data_request.rb
+++ b/app/models/extra_mobile_data_request.rb
@@ -2,7 +2,7 @@ class ExtraMobileDataRequest < ApplicationRecord
   has_paper_trail
 
   after_initialize :set_defaults
-  before_save :normalise_device_phone_number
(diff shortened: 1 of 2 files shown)


tonyheadford (approved):
Looks good

PR #1336, opened 2021-02-24
Spike - adding LA funding places school

tonyheadford:
### Context
[Trello card](https://trello.com/c/phfNsAcy/1607-spike-to-set-up-placeholder-school-named-la-funded-places)

__SPIKE__ to add a school to represent LA funding places - an allocation that can be used by LAs for children they support in institutions such as independent schools that are not included in the service.

Add a school type `LaFundedPlace` based on `CompulsorySchool`, and an `establishment_type` of `la_funded_places`
Adds policy based on `CompulsorySchool`
Adds `create_la_funded_places!` method to a `LocalAuthority` to create and populate a funding places school for that RB.

### Notes / Questions
* We need to determine what we use for a URN - e.g. start at 999000?
* Would the URN scheme need agreeing with CC?
* None of the LAs in the service have any of the address fields populated
* Some content changes may be required:
  * LA Funding Places school appears in school lists for the RB along with all the other schools - that presumably would need to be communicated to the LAs
  * Should we remove the ability to change who orders devices, or whether they want Chromebooks?
 
image
image

### DIFF
--- a/app/models/local_authority.rb
+++ b/app/models/local_authority.rb
@@ -15,4 +15,30 @@ class LocalAuthority < ResponsibleBody
   }
 
   validates :organisation_type, presence: true
+
+  def la_funded_places
+    schools.la_funded_places_establishment_type.first
+  end
+
+  def create_la_funded_places!(urn:, device_allocation: 0, router_allocation: 0, extra_args: {})
+    return unless la_funded_places.nil?
+    attrs = {
+      responsible_body: self,
+      urn: urn,
+      name: 'LA Funded Places',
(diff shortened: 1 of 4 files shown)


app/models/local_authority.rb
```diff
@@ -15,4 +15,30 @@ class LocalAuthority < ResponsibleBody
   }
 
   validates :organisation_type, presence: true
+
+  def la_funded_places
```
asmega, line 19: should this be singular then. not actually sure...

nigel-lowry (approved):

PR #1339, opened 2021-02-25
[1449] Opening schools adds related schools

asmega:
### Context

- https://trello.com/c/A7CZ5t0K/1449-support-improvement-link-predecessor-and-successor-schools

- Adds a new model `Relation` and table 'relations'
- This stores relationships between schools from GIAS
- These are added when a school is opened from the web interface which hooks into the service objects
- Added a new tab when support views a school. this shows the related schools 

### Screenshots

No related schools

image

A closed school

image

Opened school

image

### Guidance to review

- Open a school. best option is academisation which closes another school in the process
- View the closed school
- Should link to new school
- View new school
- Should link to old school

### DIFF
--- a/app/views/support/responsible_bodies/show.html.erb
+++ b/app/views/support/responsible_bodies/show.html.erb
@@ -15,106 +15,141 @@
       <%= @responsible_body.name %>
     </h1>
 
-    <%- if @responsible_body.extra_mobile_data_requests.present? %>
-      <h2 class="govuk-heading-l govuk-!-margin-top-9">Mobile data requests</h2>
-      <%= render Support::ExtraMobileDataRequestsSummaryListComponent.new(requests: @responsible_body.extra_mobile_data_requests) %>
-    <%- end %>
+    <div class="govuk-tabs" data-module="govuk-tabs">
+      <h2 class="govuk-tabs__title">
+        Contents
+      </h2>
 
-    <h2 class="govuk-heading-l govuk-!-margin-top-9">Users</h2>
+      <ul class="govuk-tabs__list">
+        <li class="govuk-tabs__list-item govuk-tabs__list-item--selected">
+          <a class="govuk-tabs__tab" href="#schools-colleges">
+            Schools and colleges
+          </a>
+        </li>
 
-    <% if policy(User).new? %>
-      <%= govuk_button_link_to 'Invite a new user', new_support_responsible_body_user_path(@responsible_body) %>
-    <% end %>
+        <li class="govuk-tabs__list-item">
+          <a class="govuk-tabs__tab" href="#users">
+            Users
+          </a>
+        </li>
 
-    <% if @users.present? %>
-      <% @users.each do |user| %>
-        <div class="user">
-          <h3 class="govuk-heading-m govuk-!-font-size-27 govuk-!-margin-bottom-2">
-            <%= govuk_link_to user.full_name, support_user_path(user) %>
-          </h3>
-          <% if policy(user).edit? %>
-            <p class="govuk-body">
-              <%= govuk_link_to "Edit user<span class=\"govuk-visually-hidden\"> #{user.full_name}</span>".html_safe, edit_support_user_path(user) %>
-            </p>
-          <% end %>
-          <%= render Support::UserPreviewSummaryListComponent.new(user: user) %>
-        </div>
-      <% end %>
-    <% else %>
-      <p class="govuk-body">None</p>
-    <% end %>
+        <%- if @responsible_body.extra_mobile_data_requests.present? %>
+          <li class="govuk-tabs__list-item">
+            <a class="govuk-tabs__tab" href="#mobile-data-requests">
+              Mobile data requests
+            </a>
+          </li>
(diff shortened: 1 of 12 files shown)


app/models/relation.rb
```diff
@@ -0,0 +1,3 @@
+class Relation < ApplicationRecord
```
tonyheadford, line 1: Could this be something more meaningful? e.g. `SchoolRelation` or `RelatedSchool` or even just `SchoolLink` as they are named in GIAS?  `Relation` sounds very low level and `ActiveRecord` like

db/migrate/20210224145258_create_relations.rb
```diff
@@ -0,0 +1,11 @@
+class CreateRelations < ActiveRecord::Migration[6.1]
+  def change
+    create_table :relations do |t|
```
tonyheadford, line 3: Same comment regarding naming - is `relations` too generic?

tonyheadford (commented):
Looks good other than question about naming of the table/model

app/views/support/responsible_bodies/show.html.erb
```diff
@@ -15,106 +15,141 @@
...
-      <div class="govuk-grid-row">
-          <div class="govuk-grid-column-two-thirds">
-            <h2 class="govuk-heading-l govuk-!-margin-top-9">Centrally managed schools and colleges</h2>
-            <p class="govuk-body"><%= @responsible_body.name %>:</p>
-            <ul id="responsible-body-centrally-managed-stats" class="govuk-list govuk-list--bullet">
-              <li>manages ordering for <%= centrally_managing_count_or_all_schools(responsible_body: @responsible_body) %> of its schools and colleges</li>
-              <li>has <%= what_to_order_allocation_list(allocations: @virtual_cap_pools) %> available to order right now</li>
-              <li>has ordered <%= what_to_order_state_list(allocations: @virtual_cap_pools) %> so far</li>
-            </ul>
+    <div class="govuk-tabs" data-module="govuk-tabs">
```
aliuk2012, line 18: Could we switch this out to use govuk_components? 
https://dfe-digital.github.io/govuk-components/#tabs

app/models/relation.rb
aldavidson: Agree with @tonyheadford here - `Relations` sounds too much like a generic `ActiveRecord` thing. I'd favour `SchoolLink`, but any of the options Tony suggests would be fine

db/migrate/20210224145258_create_relations.rb
aldavidson: ditto

app/models/relation.rb
asmega: Agreed, this is now `SchoolLink`

db/migrate/20210224145258_create_relations.rb
asmega: agreed, this is now `school_links`

app/views/support/responsible_bodies/show.html.erb
asmega: Yes, but i think that's outside the scope of this ticket

db/schema.rb
```diff
@@ -10,11 +10,20 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_02_19_145448) do
+ActiveRecord::Schema.define(version: 2021_02_25_165034) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
+  create_table "allocation_requests", force: :cascade do |t|
```
tonyheadford, line 18: Should this be in this PR?

db/schema.rb
```diff
@@ -193,6 +202,26 @@
     t.index ["status"], name: "index_preorder_information_on_status"
   end
 
+  create_table "recipients", force: :cascade do |t|
```
tonyheadford, line 205: Should this be in this PR?

db/schema.rb
```diff
@@ -374,6 +415,30 @@
     t.index ["status"], name: "index_staged_trusts_on_status"
   end
 
+  create_table "support_tickets", force: :cascade do |t|
```
tonyheadford, line 418: Should this be in this PR?

db/schema.rb
```diff
@@ -447,5 +512,6 @@
   add_foreign_key "school_virtual_caps", "virtual_cap_pools"
   add_foreign_key "school_welcome_wizards", "users", column: "invited_user_id"
   add_foreign_key "schools", "responsible_bodies"
+  add_foreign_key "support_tickets", "users"
```
tonyheadford, line 515: Should this be in this PR?

tonyheadford (changes requested):
Looks good other than the schema

db/schema.rb
asmega: good catch. thought I fixed these but they've come back to haunt me. schema should now reflect only changes in this PR

tonyheadford (approved):
Looks good!

PR #1347, opened 2021-02-27
[1634] add duplicate status for ExtraMobileDataRequests

aldavidson:
### Context

[Trello card 1634](https://trello.com/c/why6PX23/1634-add-a-new-status-for-extramobiledatarequest-problemduplicate) - We've had a lot of duplicate requests in the system, due at least in part to a bug fixed in #1329. I've removed most of the duplicate requests where it was clear which ones should be kept and which deleted, but there will still be some (around a couple of hundred) where it was not clear. 

MNOs have been using various statuses for the duplicates - some `problem_no_match_for_number`, some `problem_ineligible`, some `problem_other`, etc - but it would be much clearer for everyone involved if they had an explicit `duplicate` status to use for them.

Add `problem_duplicate` status to `ExtraMobileDataRequests`

### Guidance to review

Needs content review from @Lloyd-K or @emmajhf 

Screenshots:

Screenshot from 2021-02-27 17-04-51
Screenshot from 2021-03-01 17-36-11
Screenshot from 2021-02-27 17-09-00

### DIFF
--- a/app/components/extra_mobile_data_request_status_component.rb
+++ b/app/components/extra_mobile_data_request_status_component.rb
@@ -11,7 +11,7 @@ class ExtraMobileDataRequestStatusComponent < ViewComponent::Base
     when :new then 'blue'
     when :in_progress then 'yellow'
     when :complete then 'green'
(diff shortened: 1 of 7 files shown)


jamesrose (approved):
Code LGTM 👍

Lloyd-K (approved):

PR #1351, opened 2021-03-01
[1619] Support ticket persists in db not session

asmega:
### Context

- https://trello.com/c/JZEzGIUF/1619-convert-support-tickets-to-use-activerecord-instead-of-storing-the-details-temporary-in-session-state

- Support ticket journey now persists progress in database rather than the session
- The model is linked via the current session id of the requester
- On check answers page inverted when urn should show
- Session sweeper now calls `destroy_all` and not `delete_all` so AR callbacks fire and `SupportTickets` are removed

### Guidance to review

- Create support ticket as non logged in user
- Create support ticket as logged in user


app/controllers/support_ticket/base_controller.rb
```diff
@@ -16,12 +14,10 @@ def thank_you
...
-    redirect_to support_ticket_path if session[:support_ticket].blank?
+    redirect_to support_ticket_path unless support_ticket.started?
   end
 
-  def set_support_ticket
-    return unless session[:support_ticket]
-
-    @support_ticket = SupportTicket.new(params: session[:support_ticket])
+  def support_ticket
+    @support_ticket ||= SupportTicket.find_or_create_by!(session_id: session.id.to_s)
```
aldavidson, line 21: Just one thing to watch out for, we might get some spurious SupportTicket records that get created by the user clicking on the link, but never updated because the user never fills in the form. Shouldn't be many, but just something to keep an eye on

app/controllers/support_ticket/check_your_request_controller.rb
```diff
@@ -2,15 +2,18 @@ class SupportTicket::CheckYourRequestController < SupportTicket::BaseController
...
   end
 
   def save
-    if form.valid? && form.create_ticket
-      session[:support_ticket] = nil
-      session[:support_ticket_number] = form.ticket_number
+    @form ||= SupportTicket::CheckYourRequestForm.new(support_ticket: support_ticket)
+
+    if @form.valid? && @form.create_ticket
+      support_ticket.destroy!
```
aldavidson, line 14: I'm not sure I follow the logic here - it's probably a problem with me rather than your code, but could you explain why you destroy the `support_ticket` if `@form.create_ticket` is true?

spec/controllers/support_ticket/contact_details_controller_spec.rb
```diff
@@ -1,66 +1,74 @@
...
-      it 'redirects the user to the next step' do
-        expect(response).to redirect_to(support_ticket_support_needs_path)
+      it 'redirects to the get support home page' do
+        expect(response).to redirect_to(support_ticket_path)
       end
     end
   end
 
   describe '#save' do
     it 'stores the data in session state' do
```
aldavidson, line 60:
```suggestion
    it 'stores the data in the database' do
```

spec/controllers/support_ticket/local_authority_details_controller_spec.rb
```diff
@@ -33,13 +36,16 @@
 
   describe '#save' do
     it 'stores the data in session state' do
```
aldavidson, line 38:
```suggestion
    it 'stores the data in the database' do
```

spec/controllers/support_ticket/support_details_controller_spec.rb
```diff
@@ -33,13 +36,14 @@
 
   describe '#save' do
     it 'stores the data in session state' do
```
aldavidson, line 38:
```suggestion
    it 'stores the data in the database' do
```

aldavidson (approved):
This mostly looks fine, just a couple of minor naming suggestions and the bit that seems counter-intuitive (destroying a ticket if create_ticket is true). None of them are blockers, so I've approved it

app/controllers/support_ticket/base_controller.rb
asmega: I get your point but it's not quite so trivial. There would have to be other changes that would be needed on the first step to check if it's persisted or not as `update` requires a persisted record. The user actually has to hit the start button on the landing page for the record to be persisted which i think is acceptable.

app/controllers/support_ticket/check_your_request_controller.rb
asmega: the old code didn't quite have things in the right place. essentially `create_ticket` sends to zendesk then `destroy!` is called as we no longer need to persist this temporary data. I will:
- move `create_ticket` to `SupportTicket` and maybe rename
- `SupportTicket::CheckYourRequestForm` isn't really form so i may kill that and replace with just `SupportTicket`

asmega: I did the first bullet point but decided to keep the form object, despite the fact it hardly does anything just to keep the similarity to the other pages

app/controllers/support_ticket/base_controller.rb
aldavidson: OK, we can add some kind of clean-up job to delete old + empty SupportTickets in a separate card if it becomes an issue

asmega: it's associated with a session. the session sweeper now calls `destroy_all` so will fire AR callbacks which will delete the associated `SupportTicket` which was the old behaviour when a session was lost

PR #1354, opened 2021-03-01
Fix infer_status to accommodate coms devices

jamesrose:
### Context

This method would progress statuses of PreorderInformation when a school had ordered _some_ of its devices. This PR fixes a bug where we weren't taking into account router devices.


tonyheadford (approved):
LGTM

PR #1355, opened 2021-03-01
Split closed schools off into separate table for support RB view

tonyheadford:
### Context
[Trello card](https://trello.com/c/nbI7zpoQ/1568-do-not-show-closed-schools-in-schools-list-for-rbs-might-be-virtual-cap-related)
Do not include closed schools in the school list shown in the RB view of the support area. As an aid for support, I've moved closed schools into a separate table under the list of open schools with some additional tags to indicate whether the school is (still) in the RBs virtual cap pool or not and whether there are any users still associated with the school.

Remove closed schools from current list of RB's schools
Add closed schools under the school list
Added indicator tags to show whether the school is still in the virtual cap pool and/or still has users still associated with it.
The Allocation, Cap and Devices Ordered are the 'raw' values not the virtual cap pool values if applicable for the closed schools.

### Guidance to review
As a support user, select a responsible body to view that has one or more (GIAS status) closed schools.
The closed schools should not be visible in the main "Schools and colleges" list
There should be a "Closed schools and colleges" table under the "Schools and colleges" table containing the closed schools.
There are tags to highlight whether the school is still in a virtual pool (or not applicable) and a count of users or "No users" in each row.

image

image

### DIFF
--- a/app/controllers/support/responsible_bodies_controller.rb
+++ b/app/controllers/support/responsible_bodies_controller.rb
@@ -17,6 +17,12 @@ class Support::ResponsibleBodiesController < Support::BaseController
     @users = policy_scope(@responsible_body.users).not_deleted.order('last_signed_in_at desc nulls last, updated_at desc')
     @schools = @responsible_body
       .schools
+      .gias_status_open
+      .includes(:device_allocations, :preorder_information)
+      .order(name: :asc)
+    @closed_schools = @responsible_body
+      .schools
+      .gias_status_closed
       .includes(:device_allocations, :preorder_information)
       .order(name: :asc)
   end
--- a/app/controllers/support/responsible_bodies_controller.rb
+++ b/app/controllers/support/responsible_bodies_controller.rb
@@ -24,6 +24,9 @@ class Support::ResponsibleBodiesController < Support::BaseController
       .schools
       .gias_status_closed
       .includes(:device_allocations, :preorder_information)
+      .left_outer_joins(:user_schools)
+      .select('schools.*, count(user_schools.id) AS user_count')
+      .group('schools.id')
       .order(name: :asc)
   end
 end
--- a/app/views/support/responsible_bodies/show.html.erb
+++ b/app/views/support/responsible_bodies/show.html.erb
@@ -115,6 +115,66 @@
           </tbody>
         </table>
 
+        <% if @closed_schools.any? %>
+          <h2 class="govuk-heading-l govuk-!-margin-top-9">Closed schools and colleges</h2>
+          <table id="responsible-body-closed-schools" class="govuk-table">
+            <caption class="govuk-table__caption govuk-visually-hidden">Closed schools and colleges</caption>
+
+            <thead class="govuk-table__head">
+              <tr class="govuk-table__row">
+                <th scope="col" class="govuk-table__header govuk-!-width-one-third">Name and URN</th>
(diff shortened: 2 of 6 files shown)


app/views/support/responsible_bodies/show.html.erb
```diff
@@ -116,5 +116,64 @@
...
+              <td class="govuk-table__cell">
+                <%= dongles_allocations_data&.raw_allocation.to_i %> allocated<br>
+                <%= dongles_allocations_data&.raw_cap || 0 %> caps<br>
+                <%= dongles_allocations_data&.raw_devices_ordered || 0 %> ordered
+              </td>
+              <td class="govuk-table__cell">
+                <%= school&.preorder_information&.who_will_order_devices_label || 'Not decided' %>
+              </td>
+              <td class="govuk-table__cell">
+                <%- user_count = school.users.count %>
```
aldavidson, line 166: n+1 queries here - some RBs have over 500 schools, and the overhead might get significant. See my suggestion above for how to include the count in the query
```suggestion
                <%- user_count = school.user_count %>
```

app/controllers/support/responsible_bodies_controller.rb
```diff
@@ -17,6 +17,12 @@ def show
     @users = policy_scope(@responsible_body.users).not_deleted.order('last_signed_in_at desc nulls last, updated_at desc')
     @schools = @responsible_body
       .schools
+      .gias_status_open
+      .includes(:device_allocations, :preorder_information)
+      .order(name: :asc)
```
aldavidson, line 22: If you include these extra lines in the query, it will allow you to refer to `school.user_count` without n+1 queries:
```suggestion
      .gias_status_open
      .includes(:device_allocations, :preorder_information)
      .left_outer_joins(:user_schools)
      .select('schools.*, count(user_schools.id) AS user_count')
      .group('schools.id')
      .order(name: :asc)
```

app/controllers/support/responsible_bodies_controller.rb
```diff
@@ -17,6 +17,12 @@ def show
...
+    @closed_schools = @responsible_body
+      .schools
+      .gias_status_closed
       .includes(:device_allocations, :preorder_information)
       .order(name: :asc)
```
aldavidson, line 27: and as before:
```suggestion
      .gias_status_closed
      .includes(:device_allocations, :preorder_information)
      .left_outer_joins(:user_schools)
      .select('schools.*, count(user_schools.id) AS user_count')
      .group('schools.id')
      .order(name: :asc)
```

aldavidson (changes requested):
All good, but there is an n+1 query issue with the schools user count. I've added some suggestions for how to deal with it

app/controllers/support/responsible_bodies_controller.rb
tonyheadford: Great spot, not needed in this query but it is for closed schools, thanks.

app/controllers/support/responsible_bodies_controller.rb
tonyheadford: Great spot, it only applies to the closed schools query though as it's not needed for open schools.

app/views/support/responsible_bodies/show.html.erb
tonyheadford: Great spot, thanks.

aldavidson (approved):

PR #1361, opened 2021-03-02
[1655] On deleting user, orders_devices => false

asmega:
### Context

- https://trello.com/c/ZYKzLXeR/1655-when-deleting-a-user-ensure-they-cannot-order

- On soft deleting users set `orders_device` to false
- This is so they do not contribute to the count of TechSource accounts per org
- This way should they be re-instated they do not automatically go back into the pool of TechSource accounts causing too many accounts per org

### Guidance to review

- Sign in as support user
- Soft delete a user that can order devices
- Should set their `deleted_at` timestamp and toggle `orders_devices` to false

### DIFF
--- a/app/controllers/support/users_controller.rb
+++ b/app/controllers/support/users_controller.rb
@@ -73,7 +73,7 @@ class Support::UsersController < Support::BaseController
   def confirm_destroy; end
 
   def destroy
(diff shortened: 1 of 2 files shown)


jamesrose (approved):
Nice spot!

PR #1366, opened 2021-03-03
[1628] Include routers in allocations csv export

asmega:
### Context

- https://trello.com/c/EHwrsQLG/1628-list-of-fe-providers-on-the-service

- After a school search when exporting data to csv include router data

### Guidance to review

- Sign in as support agent
- Perform a schools search with multiple schools preferably with combination of devices and router allocations, in and out of virtual pools
- Export the data to csv
- Should see router data along with device data

### DIFF
--- a/app/services/allocations_exporter.rb
+++ b/app/services/allocations_exporter.rb
@@ -38,6 +38,12 @@ class AllocationsExporter
       'Pool devices allocation',
       'Pool devices cap',
       'Pool devices ordered',
+      'Routers allocation',
+      'Routers cap',
+      'Routers ordered',
+      'Pool routers allocation',
+      'Pool routers cap',
+      'Pool routers ordered',
     ].freeze
   end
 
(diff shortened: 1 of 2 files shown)


app/services/allocations_exporter.rb
```diff
@@ -60,6 +66,12 @@ def render(csv, query)
         (school.in_virtual_cap_pool? ? school.std_device_allocation&.allocation : nil),
         (school.in_virtual_cap_pool? ? school.std_device_allocation&.cap : nil),
         (school.in_virtual_cap_pool? ? school.std_device_allocation&.devices_ordered : nil),
+        school.coms_device_allocation&.raw_allocation,
```
aldavidson, line 69: might also want add `:coms_device_allocation` to the `.includes(...)` of the `schools` query, to avoid n+1 queries

aldavidson (approved):
Looks fine, just one suggestion about including :coms_device_allocation to the includes

app/services/allocations_exporter.rb
asmega: good spot, i'll add that in now

PR #1367, opened 2021-03-03
[1590] Re-write 'stock availability' and 'before you order' content

fofr:
To prepare for limited stock, we need to rewrite these sections to:

- make it clear what device types are out of stock, or low on stock
- provide easy access to information about different device types (but not necessarily recommend any alternatives)
- be transparent about devices that will come back into stock
- soften language about being certain what device types are needed (as their first and second choices may be unavailable)

### DIFF
--- a/app/views/school/devices/can_order.html.erb
+++ b/app/views/school/devices/can_order.html.erb
@@ -29,7 +29,6 @@
       <li><%= govuk_link_to 'give someone else access so they can place orders', school_users_path %></li>
     </ul>
 
-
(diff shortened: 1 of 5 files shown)


Lloyd-K (changes requested):
Add: <p class="govuk-body">
     <%= link_to_devices_guidance_subpage 'Read more information about the devices and specifications available.', 'device-specification' %>.
 </p>

Lloyd-K (changes requested):
Remove line 32

Lloyd-K (approved):

PR #1371, opened 2021-03-03
Remove ordinals from dates in MNO requests tables

fofr:
Use recommended GOV.UK date formatting, add helpers for consistency


ralph-hawkins (approved):
👍  nice

PR #1372, opened 2021-03-03
Add missing breadcrumb from Daily Mail opted in page

fofr:
For RBs the opted in confirmation page is one level deeper, coming off the "Get laptops" page

Screen Shot 2021-03-03 at 15 49 48


aldavidson (approved):
LGTM  :+1:

PR #1375, opened 2021-03-03
Do not show order action on devolved schools page

fofr:
Remove the "Order" button from school pages for responsible bodies that have devolved ordering to schools.

Also update the prefix to highlight that "they" ordered, rather than the RB

| Before | After |
|--|--|
|Screen Shot 2021-03-03 at 16 43 39|Screen Shot 2021-03-03 at 16 43 16|


aldavidson (approved):
LGTM :+1:

PR #1377, opened 2021-03-04
[1585] Suggest support requests use IDs not personal data

fofr:
We do not want personally identifiable information in our support system

Screen Shot 2021-03-04 at 11 32 24


aldavidson (approved):
looks fine code-wise

PR #1378, opened 2021-03-04
Update replace_a_faulty_device.md

dannychadburndfe:


Lloyd-K (changes requested):
Suggest changing line 9 to 'School, local authority and academy trust contacts can raise a request on the [Support Portal](https://computacenterprod.service-now.com/dfe) to get a replacement for a faulty device.

First H3 could also have a positive contraction: If it's not working within 3 days of receiving it

Lloyd-K (approved):

PR #1385, opened 2021-03-05
[1669] Disable worker in Heroku review apps

aldavidson:
### Context

[Trello card 1669](https://trello.com/c/iYPafQTH/1669-remove-the-worker-from-heroku-which-runs-overnight-jobs-to-import-school-data-from-gias) When a Heroku review app lives for more than a few hours, it can
trigger the overnight GIAS updates. This brings in up to 70k rows
to the DB, and triggers warnings from Heroku about the database
size exceeding the maximum available on the free plan.

Since we are now flashing login token links to the user on Heroku,
there's no reason that we can think of for actually keeping the
worker process for review apps.

Remove the worker process.

### Guidance to review

Click around the review app for this PR, make sure it works

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,1 @@
 web: bundle exec puma -p $PORT
-worker: bundle exec sidekiq -C config/sidekiq.yml
--- a/app.json
+++ b/app.json
@@ -23,9 +23,6 @@
   "formation": {
     "web": {
       "quantity": 1
-    },
-    "worker": {
-      "quantity": 1
     }
   },
   "addons": [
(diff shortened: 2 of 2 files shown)


asmega (approved):

aliuk2012:
I think you can remove ` "heroku-redis"` from line 30 too.

aliuk2012:
There are a few other things we could remove to tidy it up like the mention of the worker in `Procfile`

aliuk2012:
I was looking at what was introduced in this PR https://github.com/DFE-Digital/get-help-with-tech/pull/1385

aliuk2012 (approved):

PR #1396, opened 2021-03-08
[1661] Reject example mobile number

asmega:
### Context

- https://trello.com/c/AGdMGI1F/1661-investigate-sample-spreadsheet-rows-present-as-extramobiledatarequests

- Spreadsheet uploads with example number are rejected
- Validations added so users can see the problem to rectify

### Screenshots

image

### Guidance to review

- Sign in as school or RB
- Download the MNO template spreadsheet
- Change the name of the example user in the template spreadsheet and upload
- Should see validation error message about example number

### DIFF
--- a/spec/models/extra_mobile_data_request_spec.rb
+++ b/spec/models/extra_mobile_data_request_spec.rb
@@ -30,6 +30,15 @@ RSpec.describe ExtraMobileDataRequest, type: :model do
     end
   end
 
+  describe 'validates example number' do
+    subject(:model) { described_class.new(device_phone_number: '07123456789') }
+
+    it 'is not valid' do
+      expect(model.valid?).to be_falsey
+      expect(model.errors[:device_phone_number]).to be_present
+    end
+  end
+
   describe 'validate RB or school must be present' do
     let(:school) { create(:school) }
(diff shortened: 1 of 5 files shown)


fofr:
Looks good :+1:

tonyheadford (approved):
LGTM

spec/models/extra_mobile_data_request_spec.rb
```diff
@@ -30,6 +30,15 @@
     end
   end
 
+  describe 'validates example number' do
```
aldavidson, line 33:
```suggestion
  describe 'validates device_phone_number is not the example number' do
```

aldavidson (approved):
Code looks fine, no blockers, just a suggestion about the name of the spec example. Don't know if you've had any content design input on the error message, but it might be worth asking Emma or Kate for an opinion?

asmega:
> Code looks fine, no blockers, just a suggestion about the name of the spec example. Don't know if you've had any content design input on the error message, but it might be worth asking Emma or Kate for an opinion?

I spoke to @fofr about it so should be good enough for now. We can tweak again if it's not quite right

PR #1400, opened 2021-03-09
SLA update

Lloyd-K:
https://trello.com/c/byU8JEhp

### DIFF
--- a/app/views/devices_guidance/support_and_maintenance.md
+++ b/app/views/devices_guidance/support_and_maintenance.md
@@ -102,4 +102,4 @@ We cannot increase data limits if a pupil has used 20GB within a month. Advise t
 
 ## How to contact DfE for help
 
-If you work for a local authority, academy trust, school or college and need help with any of the devices provided by this programme, [contact us](/get-support). We aim to respond within 5 working days.
+If you work for a local authority, academy trust, school or college and need help with any of the devices provided by this programme, [contact us](/get-support). We aim to respond within 3 working days.
--- a/app/views/devices_guidance/support_and_maintenance.md
+++ b/app/views/devices_guidance/support_and_maintenance.md
@@ -102,4 +102,4 @@ We cannot increase data limits if a pupil has used 20GB within a month. Advise t
 
 ## How to contact DfE for help
 
-If you work for a local authority, academy trust, school or college and need help with any of the devices provided by this programme, [contact us](/get-support). We aim to respond within 3 working days.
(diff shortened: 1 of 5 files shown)


app/views/devices_guidance/support_and_maintenance.md
```diff
@@ -102,4 +102,4 @@ We cannot increase data limits if a pupil has used 20GB within a month. Advise t
 
 ## How to contact DfE for help
 
-If you work for a local authority, academy trust, school or college and need help with any of the devices provided by this programme, [contact us](/get-support). We aim to respond within 5 working days.
+If you work for a local authority, academy trust, school or college and need help with any of the devices provided by this programme, [contact us](/get-support). We aim to respond within 3 working days.
```
aliuk2012, line 105: Do we need to mention the SLA here? It is mentioned on the get support home page. I think when the user is reading this all they need to know at this point is that they can contact us. I don't think they should be concerned about how long it's going to take us to respond. When they try to contact us, that's when they will want to know how long it may be before they can expect a response from us.

app/views/support_tickets/thank_you.html.erb
```diff
@@ -23,11 +23,11 @@
     </h2>
 
     <p class="govuk-body">
-      Our support team has received your message. We aim to respond to you by email within 5 working days.
+      Our support team has received your message. We aim to respond to you by email within 3 working days.
       If you have not heard from us within that time, please check your spam or junk folder.
     </p>
     <p class="govuk-body">
-      We’re experiencing high demand at the moment. We know that 5 days is a bit of a long time to ask you to wait, but we appreciate your patience.
+      We’re experiencing high demand at the moment. We know that 3 days is a bit of a long time to ask you to wait, but we appreciate your patience.
```
aliuk2012, line 30: Is this needed now? I know this was added as a result of the huge spike of tickets back in Jan and asking the user to wait a week before getting a response.

aliuk2012 (changes requested):

jamesrose (approved):

app/views/devices_guidance/support_and_maintenance.md
Lloyd-K: Agreed

app/views/support_tickets/thank_you.html.erb
Lloyd-K: Removal approved

aliuk2012 (approved):

aliuk2012 (approved):

PR #1403, opened 2021-03-09
Update the home page design 

aliuk2012:
- Adds a new start page banner component
- Adds the banner to the start page
- removes the parent banner (only from the home page)
- Removes the "Guidance" link in the header nav for users who have not signed in. This was confusing because the service name and the "Guidance" link took the unsigned in user to the home page. This shouldn't affect users when they are signed in.

### Guidance to review

https://dfe-ghwt-pr-1403.herokuapp.com/


aliuk2012:
[Add sign-in link and this styling to homepage: https://ghwt-prototype.herokuapp.com/guidance-reorg](https://trello.com/c/dtYj8FET/1648-add-sign-in-link-and-this-styling-to-homepage-https-ghwt-prototypeherokuappcom-guidance-reorg)

tonyheadford (approved):
LGTM

app/components/start_page_banner/view.rb
```diff
@@ -0,0 +1,2 @@
+class StartPageBanner::View < ViewComponent::Base
```
asmega, line 1: any reason why these need to be namespaced?

aliuk2012: Good point, personally I feel it's much easier to spot what files are used for a particular component and it's easier to maintain rather than having a LONG list of components all lumped in together.

I'll remove the namespace to follow other components and increase the existing list of files in the components folder

asmega (approved):

app/webpacker/styles/application.scss
```diff
@@ -20,6 +20,7 @@ $govuk-image-url-function: frontend-image-url;
 @import "related";
 @import "card";
 @import "components/cookie-banner";
+@import "components/start_page_banner";
```
fofr, line 23: Can you rename this to `start-page-banner` to match the hyphenation used by the other components

app/webpacker/styles/inverted.scss
```diff
@@ -0,0 +1,79 @@
...
+}
+
+.app-tag--inverted {
+  background-color: govuk-colour("white");
+  color: govuk-colour("blue");
+
+  > :last-child { margin-bottom: 0 !important; margin-right: 0 !important; }
+}
+
+// PSD homepage (not signed-in)
```
fofr, line 70: `PSD homepage`?

app/webpacker/styles/inverted.scss
```diff
@@ -0,0 +1,79 @@
...
+.app-inverted--masthead {
+  @include govuk-responsive-padding(6, "bottom");
+}
+
+// Introduction slides
```
fofr, line 75: Intro slides?

fofr: Is there some copy and paste here we don't need?

PR #1413, opened 2021-03-11
add task to create users for pentest

aldavidson:
### Context

We last had a pentest 6 months ago, at which point I created users manually. We're setting up for the next test, and the work involved in identifying appropriate organisations to link the users to has been fiddly enough that I don't want to do this manually every time.   

Add a rake task to create users of all required flavours given a name and email address.
 
### Guidance to review

`rake db:create_pentest_users NAME="Jane Doe" EMAIL_ADDRESS=jane.doe@example.com`

It will create users with suffixes of `+school`, `+responsible_body` etc, and return their suffixed email addresses.
If it can't find suitable schools or responsible bodies to link them to, it will raise an error and create nothing.

### DIFF
--- a/app/services/personas/pentest_users.rb
+++ b/app/services/personas/pentest_users.rb
@@ -1,0 +1,65 @@
+class Personas::PentestUsers
+  attr_accessor :email_address, :school, :responsible_body, :mobile_network, :users
+
+  def initialize(email_address:, name:, school: nil, responsible_body: nil, mobile_network: nil)
+    @email_address = email_address
+    @name = name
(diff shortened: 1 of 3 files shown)


tonyheadford (commented):
Looks good except that I think all the responsible body and school queries should be qualified with `.gias_status_open` or `where(status: 'open')`

aldavidson:
> Looks good except that I think all the responsible body and school queries should be qualified with `.gias_status_open` or `where(status: 'open')`

Good call - I've made those changes in e247489

tonyheadford (approved):
Nice!

PR #1421, opened 2021-03-12
[1668] Enable tracking of notify emails

asmega:
### Context

- https://trello.com/c/w2iRzC2P/1668-spike-store-status-of-email-notifications-sent

- When sending emails via notify we attach attach a reference on the api call which is the `id` of the `EmailAudit`
- When notify sends back the callback sometime in the future it includes the above reference
- We then link the above reference to an existing `EmailAudit` and update the status
- Some emails we send do not create an `EmailAudit`
- In these cases when notify hits the callback endpoint we will create a new `EmailAudit`
- NOTE: Notify will also send us callbacks for emails sent via there web interface so any failed batched jobs will now be recorded by our app 

### TODO

- [x] before merging. add api keys via env vars
- [ ] after deployment configure endpoint in notify to start using the now available endpoint

### Screenshots

image

### Guidance to review

- Sign in as support user
- Find a school with users and make them able to order, this should send an email
- Check `EmailAudit` table, should have new row with `govuk_notify_id` set with a guid by notify
- Fake notify callback with the following:

```
curl --header "Content-Type: application/json" \
  --header "Authorization: Bearer password" \
  --request POST \
  --data '{"id":"GUID","reference":"EMAIL_AUDIT_ID","to":"EMAIL_ADDRESS","status":"delivered","notification_type":"email"}' \
  http://localhost:3000/notify_callbacks
```

- `EmailAudit` status should now be updated

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -98,6 +98,14 @@ RSpec/PredicateMatcher:
 RSpec/DescribeClass:
   Enabled: false
 
+RSpec/VariableDefinition:
+  Exclude:
+    - spec/mailers/log_email_delivery_observer_spec.rb
+
+RSpec/VariableName:
+  Exclude:
+    - spec/mailers/log_email_delivery_observer_spec.rb
+
 Lint/AmbiguousBlockAssociation:
   Exclude:
     - 'spec/system/support_interface/daily_report_spec.rb'
--- a/app/mailers/log_email_delivery_observer.rb
+++ b/app/mailers/log_email_delivery_observer.rb
@@ -1,9 +1,29 @@
 class LogEmailDeliveryObserver
   def self.delivered_email(message)
+    log_to_debugger(message)
+    link_message_with_audit(message)
+  end
+
+  def self.log_to_debugger(message)
     Rails.logger.debug <<~END_LOG_MSG
       To: #{message.to}
 
       #{message.preview.try(:body)}
     END_LOG_MSG
+  end
+
+  def self.link_message_with_audit(message)
+    if message.delivery_method.is_a?(Mail::Notify::DeliveryMethod)
+      govuk_notify_id = message.delivery_method.response.id
+      audit_id = message.delivery_method.response.reference
+      audit = EmailAudit.find_by(id: audit_id)
+
+      if audit
+        audit.update!(govuk_notify_id: govuk_notify_id)
+      end
+    end
+  rescue StandardError => e
+    Rails.logger.warn(e)
+    Rails.logger.warn(e.message)
   end
 end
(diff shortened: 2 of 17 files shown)


app/mailers/log_email_delivery_observer.rb
```diff
@@ -1,9 +1,29 @@
...
+    if message.delivery_method.is_a?(Mail::Notify::DeliveryMethod)
+      govuk_notify_id = message.delivery_method.response.id
+      audit_id = message.delivery_method.response.reference
+      audit = EmailAudit.find_by(id: audit_id)
+
+      if audit
+        audit.update!(govuk_notify_id: govuk_notify_id)
+      end
+    end
+  rescue StandardError => e
```
asmega, line 25: I don't really have a testing strategy so would rather be on the safe side as this is a blocking call

.rubocop.yml
```diff
@@ -98,6 +98,14 @@ RSpec/PredicateMatcher:
 RSpec/DescribeClass:
   Enabled: false
 
+RSpec/VariableDefinition:
+  Exclude:
+    - spec/mailers/log_email_delivery_observer_spec.rb
+
+RSpec/VariableName:
+  Exclude:
+    - spec/mailers/log_email_delivery_observer_spec.rb
```
asmega, line 107: rubocop seems to treat `subject` as an rspec keyword instead of `subject` as an attribute/method on an email

app/models/email_audit.rb
```diff
@@ -3,7 +3,8 @@ class EmailAudit < ApplicationRecord
   belongs_to :school, optional: true
 
   validates :message_type, presence: true
-  validates :template, presence: true
 
   scope :sent_to, ->(email_address) { where(email_address: email_address) }
+
+  scope :problematic, -> { where(govuk_notify_status: %w[permanent-failure temporary-failure technical-failure]) }
```
asmega, line 9: not used at the moment but i envisage the emails audits web interface to show problematic emails only in the near future so they can be triaged and actioned

app/mailers/log_email_delivery_observer.rb
jamesrose: @asmega Can we be more specific about what you might expect here?

jamesrose (approved):
Looks good, nice work! Just one question re: a generic rescue.

app/mailers/log_email_delivery_observer.rb
asmega: I don't think so. I've not been able to do a full integration test with notify as we only have the production environment in notify. i have been using a dev key to simulate but never a live key so I don't have 100% certainty this will behave as I expect it to. I don't know what errors it could raise.

If we don't update the audit at the moment it's not really important since it's a new feature anyway and it's currently not doing that. but if it does hypothetically go wrong i'd rather the error be swallowed silently than throw an error potentially to a user if there are any synchronous calls.

PR #1422, opened 2021-03-12
[1684] Build view of unclaimed devices count per day

tonyheadford:
### Context
[Trello card](https://trello.com/c/NBVc0nsi/1684-build-view-on-unclaimed-devices-count-per-day)
Record and display daily totals of unclaimed devices.
__NOTE__: the view of the unclaimed devices is not included in this PR, it will be done as part of the updates to the performance dashboard

Adds a new table/model `RemainingDeviceCount` to record daily totals
Adds `RemainingDevicesCalculator` service to calculate the totals
Adds `UpdateRemainingDevicesJob` which is scheduled to record the daily total at 11:55pm
Adds ability to download a CSV of all `RemainingDeviceCount` records

### Guidance to review
The current unclaimed totals can be retrieved via `RemainingDevicesCalculator.new.current_unclaimed_totals`. This returns a `RemainingDeviceCount` instance that has not been persisted.
There are no safe guards to prevent multiple `RemainingDeviceCount` records being generated for the same day, however the only place they _should_ actually be persisted is from the `UpdateRemainingDevicesJob`.
This enables us to always have a dynamic current view of the unclaimed devices during the day while only recording the value at the end of each day.
CSV can be downloaded from `/support/performance/remaining-device-counts.csv`

### DIFF
--- a/app/controllers/support/remaining_device_counts_controller.rb
+++ b/app/controllers/support/remaining_device_counts_controller.rb
@@ -1,0 +1,21 @@
+class Support::RemainingDeviceCountsController < Support::BaseController
+  before_action { authorize Support::ServicePerformance }
+
+  def index
+    respond_to do |format|
+      format.csv do
+        send_data RemainingDeviceCount.to_csv, filename: "remaining_device_counts-#{file_date}.csv"
+      end
+    end
+  end
+
+private
+
+  def file_date
+    RemainingDeviceCount.most_recent.date_of_count.strftime('%Y%m%d')
+  end
+
+  def service
+    @service ||= RemainingDevicesCalculator.new
+  end
+end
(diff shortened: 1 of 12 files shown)


app/controllers/support/remaining_device_counts_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+      format.csv do
+        send_data RemainingDeviceCount.to_csv, filename: "remaining_device_counts-#{file_date}.csv"
+      end
+    end
+  end
+
+private
+
+  def file_date
+    RemainingDeviceCount.most_recent.date_of_count.strftime('%Y%m%d')
```
jamesrose, line 15: I think `RemainingDeviceCount.most_recent` would be nil on the day you deploy this because there won't be any records until that evening.

jamesrose (approved):

app/controllers/support/remaining_device_counts_controller.rb
tonyheadford: Good spot!

PR #1425, opened 2021-03-15
Allow ssh-ing to a specific PROCESS & INSTANCE from make

aldavidson:
### Context

For a while now, we've had the convenient command `make (env) ssh` for sshing to the first web process, but if you want to connect to a different instance and/or process, you have to run the more verbose `make (env) set_cf_target  && cf ssh (fully-scoped app name) --process (process) -i (instance)`.

Recently I added the `make (env) upload|download` tasks which support `PROCESS`and `INSTANCE` parameters - we should add the same support to the ssh command.

* support the same `PROCESS` and `INSTANCE` parameters that we allow in `make (env) upload|download`
* explicitly tell you which instance and process you're connecting to, to prevent confusion
* document the new parameters in  /README.md and remove some now-redundant lines
* also add documentation for the `make (env) upload|download` tasks, which it appears I did not get time to do when adding the functionality (whoops)

### Guidance to review

Read the updated docs on the `ssh`, `upload` and `download` tasks
Try and follow them

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,8 @@ prod:
 	@true
 
 .PHONY: require_env_stub build push deploy setup_paas_env setup_paas_db setup_paas_app promote ssh \
-				logs logs-recent remote-docker-tags push-tag rollback-to timestamp-latest
+				logs logs-recent remote-docker-tags push-tag rollback-to timestamp-latest setup_ssh_params \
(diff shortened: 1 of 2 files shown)


tonyheadford (approved):
LGTM!

PR #1427, opened 2021-03-15
[1686] store & detect duplication of normalised name for ExtraMobileDataRequests

aldavidson:
### Context

[Trello card 1686](https://trello.com/c/y6UnyaGS/1686-store-and-detect-duplication-of-normalised-accountholdername-on-extramobiledatarequest) -  we're currently only looking for _exact_ matches of account_holder_name exactly as it was entered. In analysis for the first official publication of our success metrics, we found lots of closely-related names in the dataset - some all in upper or lower case vs mixed case, some with a middle name or title, or a dot after 'Mr.' vs not, etcetc . 

We should be able to help the situation by running a simple normalisation process.

* Migration to add the field `normalised_name` to `ExtraMobileDataRequest`
* Add a method to normalise the name on validation
* Change the current check for existing `account_holder_name` to instead check `normalised_name`
* Add a separate migration to populate the field

### DIFF
--- a/app/models/extra_mobile_data_request.rb
+++ b/app/models/extra_mobile_data_request.rb
@@ -2,7 +2,7 @@ class ExtraMobileDataRequest < ApplicationRecord
   has_paper_trail
 
   after_initialize :set_defaults
-  before_validation :normalise_device_phone_number
+  before_validation :normalise_device_phone_number, :normalise_name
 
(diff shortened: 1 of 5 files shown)


tonyheadford (approved):
Looks great! (bar the rubocop issue)

PR #1431, opened 2021-03-16
[1720] Hide 0 allocation on device count component

asmega:
### Context

- https://trello.com/c/szUJw8JR/1720-do-not-show-0-of-0-routers-or-0-routers-available-when-the-org-has-no-router-allocation

- Do not show any zero allocations for the device count component

### Screenshots

before
image

after
image

### Guidance to review

- Login as school user that can order
- Give them a zero device allocation
- Should not see `0 of 0`
- Give then a non zero router allocation
- Should continue to not see `0 of 0`

### DIFF
--- a/app/components/device_count_component.rb
+++ b/app/components/device_count_component.rb
@@ -49,10 +49,6 @@ private
     end
   end
 
-  def allocations
-    school.device_allocations
-  end
-
   def non_zero_allocations
     school.device_allocations.reject { |alloc| alloc.cap.zero? }
   end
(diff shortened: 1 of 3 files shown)


tonyheadford (approved):

app/components/device_count_component.rb
```diff
@@ -52,4 +52,8 @@ def availability_suffix
   def allocations
     school.device_allocations
   end
```
jamesrose, line 54:
```suggestion
```

jamesrose (approved):

PR #1434, opened 2021-03-17
Remove unused banner feature flags

fofr:
These banners were used in January as we opened up ordering to everyone in stages, and increased allocations back to their original values. They are no longer needed.

- Remove increased_allocations_banner
- Remove secondary_mass_testing_banner


asmega (approved):
once this is in. `increased_allocations_feature_flag` db column on schools might be able to go too

PR #1435, opened 2021-03-17
[1725] Give impersonation to all support users

asmega:
### Context

- https://trello.com/c/bJGKdBhW/1725-roll-out-impersonation-to-all-support-agents

- Impersonation available to all support users minus computacenter users

### Guidance to review

- Sign in as support user that is not third line
- Should be able to impersonate a user
- Sign in as a computacenter user with support capabilities
- Should not be able to impersonate someone

### DIFF
--- a/app/policies/impersonation_policy.rb
+++ b/app/policies/impersonation_policy.rb
@@ -1,9 +1,13 @@
 class ImpersonationPolicy < ApplicationPolicy
   def create?
-    user.third_line_role?
+    return false if user.is_computacenter?
+
+    user.is_support?
   end
 
   def destroy?
-    user.third_line_role?
+    return false if user.is_computacenter?
+
+    user.is_support?
   end
 end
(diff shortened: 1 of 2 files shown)


app/policies/impersonation_policy.rb
```diff
@@ -1,9 +1,13 @@
 class ImpersonationPolicy < ApplicationPolicy
   def create?
-    user.third_line_role?
+    return false if user.is_computacenter?
```
tonyheadford, line 3: Is there a way that third_line roles with computacenter access could still be permitted to do this?
I know there are a few of us...

asmega: There's only a handful of users that fit your criteria. I don't recommended enabling the `is_computacenter` flag personally as it hides items from support views when you probably want to be able to see them. Personally I only toggle this flag if and when i need it

tonyheadford (approved):

PR #1439, opened 2021-03-17
[1708] add reportable_events table & record EMDR completions

aldavidson:
### Context

Full context is in the [Trello card](https://trello.com/c/bl6bdgcM/1708-improve-the-way-we-report-on-significant-events-by-recording-them-as-reportable-events) - TL;DR is that when we're running reports, we can't rely on current status & updated_at as a reliable indicator of _when_ the current status was set. We need a better way of recording reportable events.

* Add reportable_events table and `ReportableEvent` model class
* Record `completion` events for `ExtraMobileDataRequests`

### Guidance to review

I'd like to run this as-is on production for a few days, just to double-check that the events are being recorded reliably when triggered by all known methods. Once we're happy that that's the case, I can add a migration to backfill completion events for the existing data, switch over the ad-hoc reporting code to use that table, and bring that ad-hoc code into the main codebase.

### DIFF
--- a/app/models/extra_mobile_data_request.rb
+++ b/app/models/extra_mobile_data_request.rb
@@ -4,6 +4,8 @@ class ExtraMobileDataRequest < ApplicationRecord
   after_initialize :set_defaults
   before_validation :normalise_device_phone_number, :normalise_name
 
+  after_save :record_completion_if_needed!
+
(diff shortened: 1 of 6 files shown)


jamesrose (approved):
Looks great! 🚢

PR #1451, opened 2021-03-19
[1736] Support can opt school out of notifications

asmega:
### Context

- https://trello.com/c/DegS78mW/1736-create-a-flag-in-service-to-not-contact-users-who-told-us-they-do-not-need-devices

- Added db column `opted_out_at` on `School`
- Support users can opt a school out of comms which will set the above value
- They can also remove the value to opt them back in
- `SchoolCanOrderDevicesNotifications` takes the above value into consideration and will not send any notifications if set

### Screenshots

Support school show page when opted in (default state)
image

Change page on default state (opted in and about to opt them out)
image

Support school show page when opted out
image

### Guidance to review

- Sign in as support user
- Find a school
- Should be opted in
- Opt them out
- Should be opted out
- Change order state so they get a nudge email
- Should no longer receive a nudge email

### DIFF
--- a/app/views/support/schools/opt_outs/edit.html.erb
+++ b/app/views/support/schools/opt_outs/edit.html.erb
@@ -1,0 +1,31 @@
+<%- title = t('page_titles.support.schools.opt_outs.edit') %>
+<%- content_for :title, title %>
+<%- content_for :before_content do %>
+  <%= govuk_link_to('Back', support_school_path(@school), class: 'govuk-back-link') %>
+<%- end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">
+      <span class="govuk-caption-xl"><%= @school.name %></span>
+      <%= title %>
+    </h1>
+
+    <div class="govuk-body">
+      By opting outing of communications a user from the school has expressed that they no longer wish to use the remainder of their allocation. This will prevent us from sending any further updates to them about unclaimed devices.
+    </div>
+
+    <%= form_with model: Support::OptOutForm.new(school: @school), url: support_school_opt_out_path, method: :put, scope: :school do |f| %>
+      <%= f.govuk_error_summary %>
+
+      <% opt_out_options = [OpenStruct.new(id: 1, name: 'Opt out and no longer receive communications'),
+                            OpenStruct.new(id: 0, name: 'Opt in and receive communcations')] %>
+
+      <%= f.govuk_collection_radio_buttons :opt_out, opt_out_options, :id, :name,
+        legend: { text: 'Opt school out of further notifications' }
+      %>
+
+      <%= f.govuk_submit 'Save' %>
+    <%- end %>
+  </div>
+</div>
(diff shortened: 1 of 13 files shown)


app/views/support/schools/opt_outs/edit.html.erb
```diff
@@ -0,0 +1,26 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">
+      <span class="govuk-caption-xl"><%= @school.name %></span>
+      <%= title %>
+    </h1>
+
+    <%= form_with model: Support::OptOutForm.new(school: @school), url: support_school_opt_out_path, method: :put, scope: :school do |f| %>
+      <%= f.govuk_error_summary %>
+
+      <% opt_out_options = [OpenStruct.new(id: 1, name: 'Yes, opt out and no longer receive communications'), OpenStruct.new(id: 0, name: 'No, opt in and receive communcations')] %>
```
fofr, line 17:
```suggestion
      <% opt_out_options = [OpenStruct.new(id: 1, name: 'Opt out and no longer receive communications'), OpenStruct.new(id: 0, name: 'Opt in and receive communications')] %>
```

(Note there's a type on the no option)

fofr: The "yes, opt out" is a little odd

app/views/support/schools/opt_outs/edit.html.erb
```diff
@@ -0,0 +1,26 @@
+<%- title = t('page_titles.support.schools.opt_outs.edit') %>
+<%- content_for :title, title %>
+<%- content_for :before_content do %>
+  <%= govuk_link_to('Back', support_school_path(@school), class: 'govuk-back-link') %>
+<%- end %>
+
+<div class="govuk-grid-row">
```
fofr, line 7: I think we need something on this page about this being explicitly about not wanting allocations as the reason for opting out.

db/schema.rb
```diff
@@ -349,6 +349,7 @@
     t.integer "ukprn"
     t.text "fe_type"
     t.boolean "hide_mno", default: false
+    t.datetime "opted_out_at"
```
jamesrose, line 352: Sorry to be a pain, but do you think we could rename this to something a bit more descriptive? `opted_out_of_communications` ?

app/views/support/schools/opt_outs/edit.html.erb
asmega: The yes/no bit has gone and screenshot has been updated

app/views/support/schools/opt_outs/edit.html.erb
asmega: I've added the following above the form and the screenshot has been updated
```
By opting outing of communications a user from the school has expressed that they no longer wish to use the remainder of their allocation. This will prevent us from sending any further updates to them about unclaimed devices.
```

db/schema.rb
asmega: I've changed to `opted_out_of_comms_at` as it's a timestamp not a bool and a bit shorter.
No pain, that's the point of review 🙂

jamesrose (approved):
Code LGTM - deferring to @fofr for the rest.

PR #1452, opened 2021-03-19
1721 build the service dashboard

tonyheadford:
### Context
[Trello card](https://trello.com/c/YqzQCGGs/1721-build-the-service-dashboard)
Implement the redesign of the service performance dashboard

New service performance dashboard.

### Guidance to review
There is a lot of complexity in here, switching between queries about devolved schools to responsible body's centrally managed schools, so checking the values displayed against the your local data, one tab at a time, is probably the way to review this.
image
image
image
image

### DIFF
--- a/app/models/school_device_allocation.rb
+++ b/app/models/school_device_allocation.rb
@@ -21,6 +21,22 @@ class SchoolDeviceAllocation < ApplicationRecord
     where(device_type: device_type)
   end
 
+  def self.has_fully_ordered
+    where('devices_ordered > 0 AND cap = devices_ordered')
+  end
+
+  def self.has_partially_ordered
+    where('devices_ordered > 0 AND cap > devices_ordered')
+  end
+
+  def self.has_not_ordered
+    where(devices_ordered: 0)
+  end
+
+  def self.has_not_fully_ordered
+    where('cap > devices_ordered')
+  end
+
   def self.by_computacenter_device_type(cc_device_type)
     by_device_type(Computacenter::CapTypeConverter.to_dfe_type(cc_device_type))
   end
--- a/app/models/school_device_allocation.rb
+++ b/app/models/school_device_allocation.rb
@@ -13,33 +13,13 @@ class SchoolDeviceAllocation < ApplicationRecord
 
   validates_with CapAndAllocationValidator
 
-  def self.can_order_std_devices_now
-    by_device_type('std_device').where('cap > devices_ordered')
-  end
-
-  def self.by_device_type(device_type)
-    where(device_type: device_type)
-  end
-
-  def self.has_fully_ordered
-    where('devices_ordered > 0 AND cap = devices_ordered')
-  end
-
-  def self.has_partially_ordered
-    where('devices_ordered > 0 AND cap > devices_ordered')
-  end
-
-  def self.has_not_ordered
-    where(devices_ordered: 0)
-  end
-
-  def self.has_not_fully_ordered
-    where('cap > devices_ordered')
-  end
-
-  def self.by_computacenter_device_type(cc_device_type)
-    by_device_type(Computacenter::CapTypeConverter.to_dfe_type(cc_device_type))
-  end
+  scope :has_fully_ordered, -> { where('devices_ordered > 0 AND cap = devices_ordered') }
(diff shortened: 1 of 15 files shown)


app/models/support/service_performance.rb
```diff
@@ -1,4 +1,318 @@
...
+  def number_of_centrally_managed_schools
+    @number_of_centrally_managed_schools ||=
+      School
+      .gias_status_open
+      .that_are_centrally_managed
+      .count
+  end
+
+  def total_devices_available
+    SchoolDeviceAllocation.std_device.sum(:allocation)
```
jamesrose, line 37: Would using `cap` here and elsewhere in this file be marginally safer?

app/models/support/service_performance.rb
```diff
@@ -1,4 +1,318 @@
...
+
+  def total_devices_ordered
+    SchoolDeviceAllocation.std_device.sum(:devices_ordered)
+  end
+
+  def total_devices_remaining
+    SchoolDeviceAllocation.std_device.sum('allocation - devices_ordered')
+  end
+
+  def total_routers_available
```
jamesrose, line 48: Was there anything on the card on whether this should show the number that can order vs schools in any order state?

app/models/school_device_allocation.rb
```diff
@@ -21,6 +21,22 @@ def self.by_device_type(device_type)
     where(device_type: device_type)
   end
 
+  def self.has_fully_ordered
```
asmega, line 24: can we convert these class methods to scopes?

app/models/support/service_performance.rb
```diff
@@ -1,4 +1,318 @@
...
+      .that_are_centrally_managed
+      .count
+  end
+
+  def total_devices_available
+    SchoolDeviceAllocation.std_device.sum(:allocation)
+  end
+
+  def total_devices_ordered
+    SchoolDeviceAllocation.std_device.sum(:devices_ordered)
```
asmega, line 41:
```
SchoolDeviceAllocation.joins(:school).where(schools: { status: 'closed' }).sum(:cap)
```
The above returns a non-zero number. do we need to exclude these?

app/models/support/service_performance.rb
```diff
@@ -1,4 +1,318 @@
...
+  end
+
+  def total_devices_remaining
+    SchoolDeviceAllocation.std_device.sum('allocation - devices_ordered')
```
asmega, line 45:
```
SchoolDeviceAllocation.joins(:school).where(schools: { status: 'closed' }).sum('allocation - devices_ordered')
```
The above returns a non-zero number. do closed schools need to be excluded?

app/models/support/service_performance.rb
tonyheadford: It could be, I don't know tbh, I thought `allocation` was the 'real' figure and `cap` just how much of that was available now but it has become a bit blurred?

app/models/school_device_allocation.rb
tonyheadford: Yes, we are quite inconsistent in this, good shout.

app/models/support/service_performance.rb
tonyheadford: I'm not sure, it depends on the meaning of the values we're displaying I guess. If we're showing the total devices then those closed schools that ordered still had devices that are presumably still in use and the successor school may well have had an adjustment to their allocation to account for it. Also we have to leave closed school in virtual pools at the moment so their figures are still used.

app/models/support/service_performance.rb
tonyheadford: There was something on the card about RBs that have provided all the information, but the lists on the cards seemed more like a wish list than requirements. I have implemented the design that @fofr developed, perhaps that's a question for @fofr ?

app/models/support/service_performance.rb
jamesrose: Yeah, I'm not sure what the best option is here. The difference between cap and devices ordered is what you can order now, not allocation and devices ordered...

fofr: The total for the blue box should equal the sum of the green and orange boxes below it

app/models/school_device_allocation.rb
tonyheadford: Done.

asmega (approved):

PR #1458, opened 2021-03-22
Update _privacy_notice_text.html.erb

Lloyd-K:
Case change on 'Get help with technology'


emmajhf (approved):

aliuk2012:
There are a few more instead where we use "Get Help..." instead of "Get help..."

PR #1459, opened 2021-03-22
Fix locale collision

asmega:
### Context

- Locale collision causing incorrect copy to show up

- Fix locale collision

### Guidance to review

- Sign in as support
- Visit `/support/schools`
- Should see correct h1


aldavidson (approved):
LGTM

PR #1460, opened 2021-03-22
1731 redirect /start to `/`

aliuk2012:
### Context
As we have a home page and it has a CTA, there is no need to have `/start` page. Although we have mentioned this in a few places like email templates & guidance, we need to still keep the path but instead used should be redirected to the service home page.

- redirects `/start` to `root_path`
- removes old start page view
### Guidance to review

Visit https://dfe-ghwt-pr-1460.herokuapp.com/start and it should redirect you back to the service home page


aliuk2012:
[Redirect https://get-help-with-tech.education.gov.uk/start to https://get-help-with-tech.education.gov.uk](https://trello.com/c/gN1pQpHz/1731-redirect-https-get-help-with-techeducationgovuk-start-to-https-get-help-with-techeducationgovuk)

spec/features/signing_in_as_different_types_of_user_spec.rb
```diff
@@ -34,7 +34,7 @@
     scenario 'visiting start page when already signed in redirects user to home page' do
```
fofr, line 34: Should this whole scenario change?

aliuk2012: Good spot, no need for this spec

asmega:
I expected a controller method to be removed to. can this be removed? https://github.com/DFE-Digital/get-help-with-tech/blob/882881571963a3875288945d372a83780f0dfb32/app/controllers/pages_controller.rb#L10

asmega (approved):

PR #1465, opened 2021-03-23
[1778] Email audits only shows problematic emails

asmega:
### Context

- https://trello.com/c/4p5Bo5nW/1778-tweak-email-audits-to-show-only-problematic-emails

- Only show emails with delivery issues
- As we don't care about emails that were sent successfully
- Also added `Sent at` so triage is easier

### Guidance to review

- Sign as support user
- View email audits
- Should only show problematic ones

### DIFF
--- a/app/views/support/home/show.html.erb
+++ b/app/views/support/home/show.html.erb
@@ -69,7 +69,7 @@
 
     <p class="govuk-body">Use this section to:</p>
     <ul class="govuk-list govuk-list--bullet">
-      <li>view emails the service has recently sent</li>
+      <li>view recent emails with delivery issues</li>
     </ul>
 
     <% if policy(:support).technical_support? %>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -91,7 +91,7 @@ en:
       addresses:
         edit: Update address
       email_audits:
-        index: Email audits
+        index: Recent emails with delivery issues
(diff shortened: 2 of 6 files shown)


app/views/support/home/show.html.erb
```diff
@@ -69,7 +69,7 @@
 
     <p class="govuk-body">Use this section to:</p>
     <ul class="govuk-list govuk-list--bullet">
-      <li>view emails the service has recently sent</li>
+      <li>view recent emails with delivery issues</li>
```
aldavidson, line 72: Is it worth making a similar change to the title (`page_titles.support.email_audits.index`) of the index page, to make it clear you're only seeing mails with problems?

aldavidson (approved):
LGTM - just one non-blocking suggestion about the title of the index page

app/views/support/home/show.html.erb
asmega: good point i'll update with the above as title

jamesrose (approved):

config/locales/en.yml
```diff
@@ -91,7 +91,7 @@ en:
       addresses:
         edit: Update address
       email_audits:
-        index: Email audits
+        index: Recent emails with delivery issues
```
aldavidson, line 94: :+1:

aldavidson (approved):

PR #1467, opened 2021-03-23
[1766] use the national web archives cache of the local authorities register

aldavidson:
### Context

[Trello card](https://trello.com/c/1pnxPVA4/1766-find-an-alternative-authoritative-list-of-local-authorities-in-england) - The Register of Local Authorities in England, from which we were pulling an authoritative list of English LAs for import into the service, no longer exists. [The GOV.UK Registers service has been shut down, and all data moved to the National Archives and/or Github](https://dataingovernment.blog.gov.uk/2021/02/18/new-guidance-for-publishing-data/).  

Update the URL used to pull in the data to point to the cached copy in the National Web Archives - this means we do not need to make any other changes to the import code, as the format is exactly the same as before.

### Guidance to review

`bundle exec rake import:local_authorities_in_england`

### DIFF
--- a/app/models/local_authorities_in_england_register.rb
+++ b/app/models/local_authorities_in_england_register.rb
@@ -1,7 +1,7 @@
 require 'open-uri'
 
 class LocalAuthoritiesInEnglandRegister
-  URL = 'https://local-authority-eng.register.gov.uk/records.json?page-size=5000'.freeze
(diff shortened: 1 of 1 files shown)


benilovj:
Sad times 😢

jamesrose (approved):

PR #1482, opened 2021-03-25
[1785] Remove mimemagic dependency

asmega:
### Context

- https://trello.com/c/ygU488Vz/1785-fix-blocking-mimemagic-issue

- This is to deal with licensing issues
- Further discussion can be found at https://github.com/rails/rails/issues/41750
- `canonical-rails` is pinned to a fork with needed changes
- There is a PR from that fork to the original source

### Guidance to review

- no changes to app and should function as before


Gemfile
```diff
@@ -6,16 +6,25 @@ ruby File.read('.ruby-version').chomp
 # Reduces boot times through caching; required in config/boot.rb
 gem 'bootsnap', '>= 1.1.0', require: false
 
+gem 'activesupport', '~> 6.1.3'
```
jamesrose, line 9: Why do you have to split these out?

jamesrose (approved):

Gemfile
asmega: it was suggested at https://github.com/rails/rails/issues/41750#issuecomment-805913270 which removes the dependencies that depend on mimemagic, therefore allowing us to remove mimemagic

asmega: it was suggested at https://github.com/rails/rails/issues/41750#issuecomment-805913270 which removes the dependencies that depend on mimemagic, therefore allowing us to remove mimemagic

PR #1485, opened 2021-03-25
[1788] fix issue with ReportableEvents not being recorded

aldavidson:
### Context

[Trello card](https://trello.com/b/JOgdYLg8/team-board-ghwt-service-team) - we have `after_save` hooks in place to record completion events when an `ExtraMobileDataRequest` is updated to `complete`. However, the most common usage pattern is  `bulk_update` which uses `update_all` to update many records at once in pure SQL, bypassing ActiveRecord hooks - so we're not actually recording events yet.

* Make the bulk_update method update the individual records one-by-one (there can never be more than 100 records per page, so this should not be too much of an overhead)
* Add explicit controller/service-level tests for correct behaviour 

### Guidance to review

* Log in as a user with a non-nil mobile_network_id, and some requests present on that MobileNetwork (you can use the `FakeDataService` to generate some if needed)
* From the console, note the initial value of `ReportableEvent.count`
* Use the bulk update to set some requests to `complete`
* From the console, check that the value of `ReportableEvent.count` has increased correctly
* Repeat using the download/upload spreadsheet update mechanism

### DIFF
--- a/app/controllers/mno/extra_mobile_data_requests_controller.rb
+++ b/app/controllers/mno/extra_mobile_data_requests_controller.rb
@@ -48,9 +48,11 @@ class Mno::ExtraMobileDataRequestsController < Mno::BaseController
       ids = (bulk_update_params[:extra_mobile_data_request_ids] || []).reject(&:empty?)
       extra_mobile_data_request_scope
                .where('extra_mobile_data_requests.id IN (?)', ids)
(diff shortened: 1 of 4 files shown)


asmega:
@aldavidson if you rebase your build should go green

tonyheadford (approved):
LGTM!

PR #1487, opened 2021-03-25
1737 update internet access pages

aliuk2012:


aliuk2012:
[Update "Internet access" second level page](https://trello.com/c/3Bm6G5Br/1737-update-internet-access-second-level-page)

PR #1491, opened 2021-03-27
Specify version to allow development on M1 Apple Silicon

nigel-lowry:
### Context

Couldn't get `rails` commands to work with current version of `http-parser` on newer Apple Silicon.

Update `http-parser` version.

### Guidance to review

https://github.com/cotag/http-parser/issues/12

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,6 +15,9 @@ gem 'activerecord', '~> 6.1.3'
 gem 'actionmailer', '~> 6.1.3'
 gem 'railties', '~> 6.1.3'
 
+# Allows development on M1 Apple Silicon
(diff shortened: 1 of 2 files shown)


asmega:
Hi

If you do `bundle update http` does that fix your problem? If so can we do that instead

I would prefer that as it looks like they've recently removed the `http-parser` dependency in a unreleased versioned https://github.com/httprb/http/pull/651/files

aldavidson (approved):
Thanks for the PR!

nigel-lowry:
[Rework `http` gem update](https://trello.com/c/Rufz8zEY/1842-rework-http-gem-update)

PR #1495, opened 2021-03-29
[1653] Use more govuk components

asmega:
### Context

- https://trello.com/c/IXSDAOYw/1653-switching-out-hardcoded-markup-to-use-govuk-components

- Convert the majority of hard crafted govuk components to use govuk components gem
- There are a few remaining that have odd quirks that I have left as is
- Minor tweaks:
- Adding missing translations
- Fixed missing back links or breadcrumbs
- Fix smart quotes

### Guidance to review

- Where hardcoded components have been swapped out these should render the same as the replaced versions


aldavidson (approved):
:medal_sports:

PR #1497, opened 2021-03-29
Enable change who will order

tonyheadford:
### Context
[Trello card](https://trello.com/c/y1u5RdWk/1235-allow-centrally-managed-rbs-to-change-who-will-order-devices-on-schools-from-school-to-responsible-body)
Enable RBs to switch a school from devolved ordering to being centrally managed - but not the reverse.
Enable same functionality for support users

Enable the change link on a schools details except when RB has `vcap_feature_flag == true` and school is centrally managed.
Guard against accessing the change who orders page directly when the conditions are not met

### Guidance to review
As a RB user, try with/without `vcap_feature_flag` set to view your schools and change who will order devices.
When `vcap_feature_flag` is set, you should see warning text on the 'change who will order' page informing that the change to centrally managed is irreversible. This is not present when the `vcap_feature_flag` is not set
When `vcap_feature_flag` is set and school is already centrally managed there should be no 'Change' link to change who will order devices.
Repeat the above as a support user. 

image
image
image
image


jamesrose (approved):
Looks great. 🚢

PR #1501, opened 2021-03-30
[1798] fix undefined method is_responsible_body_user? for nil

aldavidson:
### Context

[Trello card](https://trello.com/c/yOWUsU32/1798-fix-undefined-method-isresponsiblebodyuser-for-nilnilclass) - there's some occasional edge case that results in an error on the sign_in_token sent page, whereby the user has a valid session but no `current_user`. On investigation, I found that this can happen when a user signs in successfully, but then goes back to the "Check your email" / sign-in token sent page. That page is uses the token given on the URL to assign `@current_user`, as there is no valid session yet, but if the user goes _back_ to an already-user token sent page, the token will not be valid anymore. 

Stop this edge case throwing an error
Add a spec example for it

### Guidance to review

Try signing in, then using your browser back button to go back to the 'Check your email' page.
It should not throw an error
It should display "We didn't recognise that link"

### DIFF
--- a/app/controllers/sign_in_tokens_controller.rb
+++ b/app/controllers/sign_in_tokens_controller.rb
@@ -60,6 +60,7 @@ class SignInTokensController < ApplicationController
 
   def sent
     @current_user = User.where(sign_in_token: params[:token]).first
+    render(:token_not_recognised, status: :bad_request) unless @current_user.present?
   end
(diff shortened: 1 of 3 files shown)


jamesrose (approved):
Looks good. Thanks!

PR #1502, opened 2021-03-30
Message telling users to get permission before inviting someone

ralph-hawkins:
Context:
https://trello.com/c/V2czUlbD/1760-notify-or-display-a-privacy-notice-when-a-new-user-has-been-added

Because we sometimes send an invited user’s details to TechSource before they have signed-in and seen the privacy policy we need the inviting user to show the privacy policy to the invited user before they invite them.

Manage users:

localhost_3000_schools_106621_users_new(screenshot for design history) (1)

School wizard flow (if ordering is devolved to the school and user has TechSource account):

localhost_3000_schools_106603_will-other-order(screenshot for design history)

We might want to consider removing the option to add a user from the school wizard flow because it doesn't seem likely a user will speak to the person they're inviting mid-flow.

### DIFF
--- a/app/views/school/users/index.html.erb
+++ b/app/views/school/users/index.html.erb
@@ -15,7 +15,7 @@
       <li>change who places orders for devices</li>
     </ul>
 
-    <p class="govuk-body">You can also choose to let up to 3 users:</p>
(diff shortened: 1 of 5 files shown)


jamesrose (approved):
Looks great, thanks @ralph-hawkins

PR #1503, opened 2021-03-30
[1820] sentry upgrade

asmega:
### Context

- https://trello.com/c/Wymbaha7/1820-upgrade-sentry

- We were not using the latest API from sentry
- This upgrades the gems to latest api is in use (from 3 to 4)
-


app/controllers/application_controller.rb
```diff
@@ -27,6 +28,10 @@ def identify_user!
     @current_user ||= (SessionService.identify_user!(session) || User.new)
   end
 
+  def set_sentry_user
+    Sentry.set_user(id: current_user.id)
```
jamesrose, line 32:
```suggestion
    Sentry.set_user(id: current_user&.id)
```

`SessionService.identify_user!` can return nil if the user is not signed in.

asmega: A few lines above
```
@current_user ||= (SessionService.identify_user!(session) || User.new)
```
This should get rid of the need for a null/nil guard but i can add anyway. There's also a test for this to make sure it all works

asmega: done

jamesrose (approved):

PR #1522, opened 2021-04-01
[1821] allow support to query ExtraMobileDataRequest completions

aldavidson:
### Context

[Trello card](https://trello.com/c/AA7dPOWD/1821-add-ability-to-query-for-extramobiledatarequest-completions-up-to-a-given-date-time-to-support-ui) - now that we have an agreed monthly schedule for reporting MNO request completions, and we also have a working method of recording completions in a separate table as they happen, we can and should add the ability to query for completions between dates into the support UI. 

* Add a bit of UI to `/support/performance` to explicitly display the completions
* Add a form for querying between dates, tucked away behind a summary/details component until you need it

#### Data setup
You will need some `ReportableEvent` records, with `event_name: 'completion'` and `record_type: 'ExtraMobileDataRequest'`
The easiest way is to generate some `ExtraMobileDataRequest` records (either using the `FakeDataService`, or actually submitting them in the school / responsible body UI) and then either complete them in the `/mno/` UI), or run something like this:

```
ExtraMobileDataRequest.where(status: 'complete').each{ |r| ReportableEvent.create!(event_name: 'completion', record: r, event_time: r.updated_at) }
```
If you are creating them all at once, you might want to stagger the event_times, something like: `event_time: r.updated_at - ((rand*10).to_i * 1.day) `

#### Testing the UI

* Login as a support user
* Go to Support home > Service performance > MNO
* Scroll down to just below the tiles of counts by status
* Try entering different Chronic-compatible/incompatible date formats, make sure it doesn't break on anything unexpected (it should just treat unparsable strings as nils) 

### Screenshots

Screenshot from 2021-04-01 16-52-40

Screenshot from 2021-04-01 16-54-14

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,6 +20,7 @@ gem 'http-parser', '~> 1.2.3'
 
 # Canonical meta tag
 gem 'canonical-rails', git: 'https://github.com/asmega/canonical-rails.git', ref: '1b9426229dfa5136326f5ce4963e8e8d1cb3b23c'
+gem 'chronic'
 gem 'dotenv-rails'
 # Having Faker here rather than in dev/test lets us still create
 # fake data in the deployed Docker container
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -76,6 +76,7 @@ GEM
       mail
     childprocess (3.0.0)
     choice (0.2.0)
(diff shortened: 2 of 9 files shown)


jamesrose (approved):
Nice work!

PR #1525, opened 2021-04-06
Remove "Parents/Carers" banner from the service

aliuk2012:
- Removes the Parents banner from the service as we are not seeing as main parents contacting us and we have more guidance for these types of users.

image

### Guidance to review
visit https://dfe-ghwt-pr-1525.herokuapp.com/ and go to a couple guidance pages and you should not see the above banner appear on any of the pages.


aliuk2012:
[Review parents banner - can we point to the new guidance page instead?](https://trello.com/c/ma2wBK7J/1616-review-parents-banner-can-we-point-to-the-new-guidance-page-instead)

tonyheadford (approved):

PR #1531, opened 2021-04-07
[1238] Can disable notifications in update service

asmega:
### Context

- https://trello.com/c/c8nB3ikw/1238-allow-bulk-opening-page-to-work-with-many-thousands-of-urns

- Can disable notifications in `SchoolOrderStateAndCapUpdateService`

### Guidance to review

- Code review


tonyheadford (approved):
Non-blocking comment, could the flag be called `disable_user_notifications` to make it clear that it is only those notifications that are disabled?

jamesrose (approved):

PR #1540, opened 2021-04-09
[1238] Batch allocations upload

asmega:
### Context

- https://trello.com/c/c8nB3ikw/1238-csv-upload-many-allocations-to-be-updated
- Allows batch updating of allocations and order state via CSV upload
- Previously this would be done via the console which would be time consuming, error prone and limited to developers

- In support area bulk enabling schools to be able to order their full allocation has been removed.
- This is because most school already order their full allocation so no longer needed
- Modified bulk adjust allocation page
- Previously this was a textarea it is now a CSV file upload so data can be structured beforehand
- This use to be a synchronous request and timeout with large volumes and has now been made asynchronous
- Upon uploading the CSV, each row is parsed and persisted to database. A job is also enqueued for each record 
- Each CSV upload is associated with an auto generated UUID to form a `batch_id`
- User is redirected to the batch showing jobs that will be worked on
- When uploading CSV user can select whether not user notifications will be sent
- Accessing to previous batch reports are accessible in the `Technical support` area by support users
- When viewing an uploaded batch user can press button to send notification to schools in the batch

#### CSV upload page
image

#### Batch report after processing has finished
image

#### Previous available batch reports in technical support area
image

### Guidance to review

- Sign in as support user
- Navigate to `/support/schools`
- Click on `Bulk upload allocation updates`
- Upload different CSV with different datasets to test
- Check notifications are or are not sent based on the option selected when uploading CSV
- Check progress report summary updates correctly
- Check batch reports available in technical area `/support/technical`


jamesrose:
This looks really great @asmega. I did have one question around notifications. The common use case with CC is the following:

1. Allocations increased
2. Wait for manual confirmation from CC
3. Send notifications

Essentially they are not done immediately following an allocation update. Does this process support that? Can I decide to send notifications at a later date for a batch?

app/jobs/allocation_job.rb
```diff
@@ -0,0 +1,34 @@
+class AllocationJob < ApplicationJob
+  queue_as :default
+
+  def perform(allocation_batch_job)
+    school = allocation_batch_job.school
+    order_state = allocation_batch_job.order_state
+    current_allocation = school.std_device_allocation
+
+    current_allocation_value = current_allocation&.allocation || 0
```
jamesrose, line 9:
```suggestion
    current_allocation_value = current_allocation&.raw_allocation || 0
```

`allocation` and `cap` can be from a virtual pool or a singular pool. In this instance I think we'll want to raise relative to the `raw_allocation`. Same for `cap` below.

app/jobs/allocation_job.rb
```diff
@@ -0,0 +1,34 @@
...
+    new_allocation_value = current_allocation_value + allocation_batch_job.allocation_delta
+
+    current_cap_value = current_allocation&.cap || 0
```
jamesrose, line 12:
```suggestion
    current_cap_value = current_allocation&.raw_cap || 0
```

asmega: I'll make the change although I can't actually make a failing test for this. I reckon as the state changes to either:
- `cannot_order` cap will get zeroed down to `devices_ordered`
- `can_order` cap gets set to the max allocation
So we could set `cap` to anything, I don't think it'd actually make a difference

app/jobs/allocation_job.rb
asmega: nice spot, done

app/jobs/allocation_job.rb
jamesrose: The test for this would be:

- Virtual pool with two schools at 10 devices each.
- Using this service object, update allocation for school A by 2.
- Does the virtual pool increase to 22 or 32? (It should be the former.)

nigel-lowry (approved):

app/jobs/allocation_job.rb
asmega: That line does set `current_cap_value` to the incorrect value with `cap` and correct value with `raw_cap`. However by the end of it all the outcome is the same and makes no difference hence why I can't get a test to fail

asmega: > * pool with two schools at 10 devices each.
> * Using this service object, update allocation for school A by 2.
> * Does the virtual pool increase to 22 or 32? (It should be the former.)

`22` with either `cap` or `raw_cap`

tonyheadford (approved):
Great stuff!

asmega:
> This looks really great @asmega. I did have one question around notifications before going deeper into the code here. The common use case with CC is the following:
> 
> 1. Allocations increased
> 2. Wait for manual confirmation from CC
> 3. Send notifications
> 
> Essentially they are not done immediately following an allocation update. Does this process support that? Can I decide to send notifications at a later date for a batch?

@jamesrose you got yourself a shiny new button. When viewing a batch you can now click a button to enqueue jobs to send notifications to schools in the batch. The jobs checks and update a flag to prevent double sending.

jamesrose (approved):

markhallen:
[Bulk Update Allocations should allow negative values](https://trello.com/c/xFxgUms1/1916-bulk-update-allocations-should-allow-negative-values)

PR #1542, opened 2021-04-09
Updated text on 'Your Organisations' page to be less terse

nigel-lowry:
### Context

Text on the 'Your Organisations' page was seen as being too terse.

Apply updated text

### Guidance to review

https://ghwt-prototype.herokuapp.com/your-organisations

### DIFF
--- a/app/views/schools/index.html.erb
+++ b/app/views/schools/index.html.erb
@@ -5,6 +5,9 @@
   <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-xl"><%= title %></h1>
 
+    <p class="govuk-body">You have access to manage ordering on behalf of more than one organisation.</p>
+    <p class="govuk-body">Select which one you want to view:</p>
+
     <%- @schools.each do |school| %>
       <h2 class="govuk-heading-l govuk-!-font-size-27">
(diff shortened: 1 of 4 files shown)


nigel-lowry:
[Update content on 'Your organisations'](https://trello.com/c/uFSYjeeH/1771-update-content-on-your-organisations)

app/views/schools/index.html.erb
```diff
@@ -5,6 +5,9 @@
   <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-xl"><%= title %></h1>
 
+    <p class="govuk-body">You have access to manage ordering on behalf of more than one organisation.</p>
+    <p class="govuk-body">Select which one you want to view:</p>
```
aliuk2012, line 9: I know you were a little worried about not including this in the locales file but I don't think you need to stress about it because it's very unlikely we will reuse this copy on other pages and if you look at the other pages you will see a lot of content isn't in the locale files. We tend to use a locales file if we likely to reuse the copy. For example, keeping browser title,  page headings, and breadcrumbs consistent then locales are great to use.

aliuk2012 (approved):

jamesrose (approved):

PR #1543, opened 2021-04-11
Add global `Timecop.return` hook

nigel-lowry:
### Context

Specs which used the Timecop gem had there own `after` hooks to make sure `Timecop.return` was used so system time changes didn't bleed into other specs. 

This PR adds a global `before` hook so `Timecop.return` is always called so nobody has to remember to do it themselves anymore. 

This commit has been manually verified to work with RSpec features as well as unit specs.

Adds global `before` hook. Having this in a `before` hook is safer than having it in an `after` hook because an `after` hook isn't guaranteed to be run (when a spec errors).

There are some calls to `Timecop.return` which are kept in because they're called multiple times in the middle of a spec to set and reset specific behaviour.

### DIFF
--- a/spec/controllers/techsource_launcher_controller_spec.rb
+++ b/spec/controllers/techsource_launcher_controller_spec.rb
@@ -32,10 +32,6 @@ RSpec.describe TechsourceLauncherController, type: :controller do
         sign_in_as user
       end
 
-      after do
-        Timecop.return
-      end
-
       it 'renders the unavailable page' do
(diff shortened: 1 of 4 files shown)


markhallen (approved):
LGTM

tariqhamid (approved):
looks fine

jamesrose (approved):

nigel-lowry:
[Timecop global reset hook](https://trello.com/c/pzEqm1GZ/1840-timecop-global-reset-hook)

PR #1579, opened 2021-04-20
Promote git action text update

nigel-lowry:
### Context

In the promote Github Action it's not clear with our production environment is called `production` or `prod`. The correct one is `prod` and whilst there seems to be no way to constrain users to this, this PR adds descriptive text to inform users of this instead. 

* Fix URL typo in README
* Add text to Github Action description to prompt users to use the correct environments


aldavidson (approved):
looks good! :+1:

PR #1584, opened 2021-04-20
One "Import data" task to rule them all

aliuk2012:
### Guidance to review
Run the following

```
bundle exec rails db:drop
bundle exec rails db:create
bundle exec rails db:schema:load
bundle exec rails db:migrate
bundle exec rails db:seed
bundle exec rails import:all
```
It will take some time to run the import but once its done you should have all the schools, RB and some user accounts setup.


README.md
```diff
@@ -38,14 +38,12 @@ To perform manual releases, you will need:
...
+### Setting up local development data
 
-There are several tasks you can run to get schools and responsible bodies:
+These are the tasks that you will need to run to set up your local db:
 
-- Run `bundle exec rake import:responsible_bodies` to import local authorities and trusts
-- Run `bundle exec rake import:schools` to import schools into staging area
-- Then open up a rails console with `bundle exec rails c`
-- Run `service = SchoolUpdateService.new; DataStage::School.all.each { |staged| puts "importing #{staged.urn}..."; service.create_school!(staged) }` to convert staged schools into schools
+- Run `bundle exec rails db:seed` to insert mobile network operators
```
aliuk2012, line 45: I'm in two minds as to whether this should still be a seed or whether we should move it into the `import` task namespace. What do others think?

lib/tasks/import_schools.rake
```diff
@@ -1,9 +1,24 @@
...
     StageGiasDataJob.perform_now
   end
 
+  desc 'Process schools records from DataStage::School to School'
+  task process_schools_data: :environment do
+    service = SchoolUpdateService.new
+
+    # TODO: Can be removed once we have data for these RB
+    DataStage::School.gias_status_open.where('responsible_body_name like ?', '%Northampton%').destroy_all
+    DataStage::School.gias_status_open.where('responsible_body_name like ?', '%Overseas Establishments%').destroy_all
```
aliuk2012, line 14: @tonyheadford this is what I had to do so that I can successfully create the other schools.

aliuk2012:
[Simplifly local dev db setup process](https://trello.com/c/xy5FC0MC/1656-simplifly-local-dev-db-setup-process)

README.md
tonyheadford: Feels like we should be consistent and either make them all seeds or all rake tasks. I don't know if there is a reason this needs to be in the seeds file rather than a rake task, maybe someone else knows?

nigel-lowry (approved):

tonyheadford (approved):
Nice one. Only comment would be that should we note somewhere in the README that `import:all` should not be used in production?

jamesrose (approved):

nigel-lowry (approved):

PR #1589, opened 2021-04-21
Remove hard-coded date string

nigel-lowry:
### Context

When a TechSource maintenance period was active we were displaying the hardcoded string that `You will be able to order devices when it reopens on Saturday 28 November` which is wrong.

https://trello.com/c/d7XZ7U4y

Dynamically set the end of the maintenance period (and include the time)


nigel-lowry:
[Wrong hard-coded date during TechSource maintenance window](https://trello.com/c/d7XZ7U4y/1886-wrong-hard-coded-date-during-techsource-maintenance-window)

aliuk2012 (approved):

PR #1605, opened 2021-04-23
[1712] Support interface to add support users

asmega:
### Context

- https://trello.com/c/lkhFVwWu/1712-third-line-support-can-add-other-support-users
- Save the need for a developer to onboard additional support users

- Third line support users can:
  - view privileged users: support users, computacenter users and mno users
  - can also add other support users - these are restricted to certain domains for the email addresses
  - revoke all privileges of one of these users

### Screenshots

New section on support homepage 
image

Index view
image

Adding another user
image

### Guidance to review

- Sign in as non third line support
- Cannot see or access this areas
- Sign in as third line support
- Should be able to see and access new area
- Can see existing privileged users
- Can add privileged users
- Can remove privileges of users

### DIFF
--- a/app/form_objects/support/privileged_user_form.rb
+++ b/app/form_objects/support/privileged_user_form.rb
@@ -1,0 +1,45 @@
+class Support::PrivilegedUserForm
+  include ActiveModel::Model
+
+  attr_accessor :full_name, :email_address, :privileges
+
+  validates :full_name, presence: { message: 'Enter the name of the user' }
+  validates :email_address, presence: { message: 'Enter the email address of the user' },
+                            format: { with: /\A.*@computacenter.com|.*@digital.education.gov.uk|.*@education.gov.uk\z/,
+                                      message: 'Email address must be end with @computacenter.com, @digital.education.gov.uk or @education.gov.uk' }
+  validate :validate_email_not_taken
+  validate :validate_atleast_one_privilege
+  validate :validate_acceptable_privileges
+
+  def create_user!
+    User.create!(full_name: full_name, email_address: email_address, is_support: is_support?, is_computacenter: is_computacenter?)
+  end
+
+private
+
+  def is_support?
+    privileges.include?('support')
+  end
+
+  def is_computacenter?
+    privileges.include?('computacenter')
+  end
+
+  def validate_acceptable_privileges
+    if privileges.union(['', 'support', 'computacenter']).size > 3
+      errors.add :privileges, :inclusion, message: 'Only select from available privileges'
+    end
+  end
+
+  def validate_atleast_one_privilege
+    if privileges.reject(&:blank?).empty?
+      errors.add :privileges, :at_least_one, message: 'Select at least one privilege to apply'
+    end
+  end
+
+  def validate_email_not_taken
+    if User.exists?(['lower(email_address) = ?', email_address.downcase])
+      errors.add :email_address, :duplicate, message: 'This email adddress has already been taken'
+    end
+  end
+end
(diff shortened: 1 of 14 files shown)


app/form_objects/support/privileged_user_form.rb
```diff
@@ -0,0 +1,45 @@
...
+  include ActiveModel::Model
+
+  attr_accessor :full_name, :email_address, :privileges
+
+  validates :full_name, presence: { message: 'Enter the name of the user' }
+  validates :email_address, presence: { message: 'Enter the email address of the user' },
+                            format: { with: /\A.*@computacenter.com|.*@digital.education.gov.uk|.*@education.gov.uk\z/,
+                                      message: 'Email address must be end with @computacenter.com, @digital.education.gov.uk or @education.gov.uk' }
+  validate :validate_email_not_taken
+  validate :validate_atleast_one_privilege
```
nigel-lowry, line 11: I'd probably add an underscore so the method becomes `validate_at_least_one_privilege`

app/form_objects/support/privileged_user_form.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+  end
+
+  def validate_atleast_one_privilege
+    if privileges.reject(&:blank?).empty?
+      errors.add :privileges, :at_least_one, message: 'Select at least one privilege to apply'
+    end
+  end
+
+  def validate_email_not_taken
```
nigel-lowry, line 40: Can't this use the standard uniqueness validation?

app/form_objects/support/privileged_user_form.rb
```diff
@@ -0,0 +1,45 @@
...
+  validate :validate_atleast_one_privilege
+  validate :validate_acceptable_privileges
+
+  def create_user!
+    User.create!(full_name: full_name, email_address: email_address, is_support: is_support?, is_computacenter: is_computacenter?)
+  end
+
+private
+
+  def is_support?
```
nigel-lowry, line 20: If these predicate methods were renamed to something like `support?` then we could use RSpec syntactic sugar on the assertions (see other comment in pull request).

spec/form_objects/support/privileged_user_form_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    subject(:form) { described_class.new(full_name: 'some body', email_address: 'user@digital.education.gov.uk', privileges: %w[support computacenter]) }
+
+    it 'sets up privileges' do
+      expect {
+        form.create_user!
+      }.to change(User, :count).by(1)
+
+      user = User.last
+
+      expect(user.is_support?).to be_truthy
```
nigel-lowry, line 82: If we changed the method name, the assertion could become `expect(user).to be_support`

nigel-lowry (approved):

spec/form_objects/support/privileged_user_form_spec.rb
asmega: unfortunately these are already present as db column names. if we want to change these i think it needs to be part of a different ticket and addressed holistically as i touches lots of different parts. but yes i agree we can have these switched over

app/form_objects/support/privileged_user_form.rb
asmega: See my other comment. this is private so damage is limited but can be changed when we tackle the whole problem

app/form_objects/support/privileged_user_form.rb
asmega: This is not a an ActiveRecord model, so I can't use the standard uniqueness validator as it doesn't hook up to a data source.

nigel-lowry (approved):

PR #1611, opened 2021-04-26
[1909] Add SocialCareLeaversProvision

asmega:
### Context

- https://trello.com/c/PT2OlVWG/1909-add-social-care-leavers-school-type
- We need a start point to hook on behaviour for social care leavers provision

- Add `SocialCareLeaversProvision`
- Add method at attach such a provision to an LA

### Guidance to review

- Should be able to attach an `SocialCareLeaversProvision` to an LA via the rails console

### DIFF
--- a/app/concerns/school_type.rb
+++ b/app/concerns/school_type.rb
@@ -18,6 +18,7 @@ module SchoolType
       special: 'special',
       other_type: 'other_type',
       la_funded_place: 'la_funded_place',
+      social_care_provision: 'social_care_provision',
     }, _suffix: true
 
     def human_for_school_type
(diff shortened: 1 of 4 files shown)


app/models/social_care_leavers_provision.rb
```diff
@@ -0,0 +1,9 @@
+class SocialCareLeaversProvision < CompulsorySchool
+  def institution_type
+    'local_authority'
+  end
+
+  def techsource_urn
+    "SCL_#{responsible_body.gias_id}"
```
asmega, line 7: will be used at some point in the future

asmega:
Will be superceded by https://github.com/DFE-Digital/get-help-with-tech/pull/1612/files

PR #1612, opened 2021-04-26
Responsible body updates for iss

tonyheadford:
### Context
[Trello card](https://trello.com/c/EINnKlVj/1905-iss-add-maidenhead-la-and-cornwall-la-schools) - when adding these initial independent special school provisions for manual processing we identified extra functionality that wasn't in place yet that was needed and also confirmed the URN format and added provision for Social care

Add creation of ISS and SCL provision methods to the `LocalAuthority` model.
Adds `provision_urn` and `provision_type` to school DB schema
Prevents `LaFundedPlace` schools from appearing in the responsible bodies list of schools
Uses `provision_urn` in place of `urn` where applicable

### Guidance to review
Add ISS or SCL provisions via the `LocalAuthority#create_iss_provision!` or `LocalAuthority#create_scl_provision!`. This will add a `LaFundedPlace` with the `provision_type` set to `iss` or `scl`.
A RB user should not see either of these in their list of schools and colleges.

### DIFF
--- a/app/models/local_authority.rb
+++ b/app/models/local_authority.rb
@@ -16,15 +16,39 @@ class LocalAuthority < ResponsibleBody
 
   validates :organisation_type, presence: true
 
-  has_one :la_funded_place, foreign_key: 'responsible_body_id'
+  has_many :la_funded_provisions, foreign_key: 'responsible_body_id', class_name: 'LaFundedPlace'
+  has_one :iss_provision, -> { iss_provision }, foreign_key: 'responsible_body_id', class_name: 'LaFundedPlace'
+  has_one :scl_provision, -> { scl_provision }, foreign_key: 'responsible_body_id', class_name: 'LaFundedPlace'
 
-  def create_la_funded_places!(urn:, device_allocation: 0, router_allocation: 0, extra_args: {})
-    return la_funded_place if la_funded_place.present?
+  def create_iss_provision!(device_allocation: 0, router_allocation: 0, extra_args: {})
+    create_provision!(urn: "ISS#{gias_id}",
+                      name: 'State-funded pupils in independent special schools and alternative provision',
+                      provision_type: :iss,
+                      device_allocation: device_allocation,
+                      router_allocation: router_allocation,
+                      extra_args: extra_args)
+  end
+
+  def create_scl_provision!(device_allocation: 0, router_allocation: 0, extra_args: {})
+    create_provision!(urn: "SCL#{gias_id}",
+                      name: 'Care leavers',
+                      provision_type: :scl,
+                      device_allocation: device_allocation,
+                      router_allocation: router_allocation,
+                      extra_args: extra_args)
+  end
+
+private
+
+  def create_provision!(urn:, name:, provision_type:, device_allocation: 0, router_allocation: 0, extra_args: {})
+    existing_provision = la_funded_provisions.find_by(provision_type: provision_type)
+    return existing_provision unless existing_provision.nil?
 
     attrs = {
       responsible_body: self,
-      urn: urn,
-      name: 'State-funded pupils in independent special schools and alternative provision',
+      provision_urn: urn,
+      name: name,
+      provision_type: provision_type,
       establishment_type: 'la_funded_place',
       address_1: address_1,
       address_2: address_2,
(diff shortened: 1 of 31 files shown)


app/models/local_authority.rb
```diff
@@ -18,21 +18,39 @@ class LocalAuthority < ResponsibleBody
 
   has_one :la_funded_place, foreign_key: 'responsible_body_id'
 
-  def create_la_funded_places!(urn:, device_allocation: 0, router_allocation: 0, extra_args: {})
+  def create_iss_places!(device_allocation: 0, router_allocation: 0, extra_args: {})
+    create_la_funded_places!(urn: "ISS_#{gias_id}",
```
asmega, line 22: should `urn` be set? should we be setting `la_funded_place_urn` instead

app/models/local_authority.rb
```diff
@@ -18,21 +18,39 @@ class LocalAuthority < ResponsibleBody
...
+                            name: 'State-funded pupils in independent special schools and alternative provision',
+                            device_allocation: device_allocation,
+                            router_allocation: router_allocation,
+                            extra_args: extra_args)
+  end
+
+  def create_scl_places!(device_allocation: 0, router_allocation: 0, extra_args: {})
+    create_la_funded_places!(urn: "SCL_#{gias_id}",
+                            name: 'Social care',
```
asmega, line 31: needs to be `Care leavers`

app/models/local_authority.rb
```diff
@@ -18,21 +18,39 @@ class LocalAuthority < ResponsibleBody
...
-  def create_la_funded_places!(urn:, device_allocation: 0, router_allocation: 0, extra_args: {})
+  def create_iss_places!(device_allocation: 0, router_allocation: 0, extra_args: {})
+    create_la_funded_places!(urn: "ISS_#{gias_id}",
+                            name: 'State-funded pupils in independent special schools and alternative provision',
+                            device_allocation: device_allocation,
+                            router_allocation: router_allocation,
+                            extra_args: extra_args)
+  end
+
+  def create_scl_places!(device_allocation: 0, router_allocation: 0, extra_args: {})
```
asmega, line 29: actually can we remove this method for this PR. I think having the same instance type of `LaFundedPlace` for Social Care Leavers will be problematic as that just leaves the name of the school as the only anchor point to hook and attach behaviour to which I think is quite brittle. I'd rather introduce another type to remove any ambiguity

db/migrate/20210426150053_add_la_funded_place_urn_to_schools.rb
```diff
@@ -0,0 +1,5 @@
+class AddLaFundedPlaceUrnToSchools < ActiveRecord::Migration[6.1]
+  def change
+    add_column :schools, :la_funded_place_urn, :string, index: true
```
asmega, line 3: when we introduce Social Care Leavers i assume we want to use the same column? if so would this be called `other_urn` or to that extent. Or would we introduce another column such as `social_care_urn`?

app/models/local_authority.rb
tonyheadford: Done

db/migrate/20210426150053_add_la_funded_place_urn_to_schools.rb
tonyheadford: Using `provision_urn` now as discussed

app/models/local_authority.rb
tonyheadford: Updated to use `provision_type` on `LaFundedPlace` to differentiate

nigel-lowry (approved):

db/migrate/20210426150053_add_la_funded_place_urn_to_schools.rb
```diff
@@ -0,0 +1,6 @@
+class AddLaFundedPlaceUrnToSchools < ActiveRecord::Migration[6.1]
+  def change
+    add_column :schools, :provision_urn, :string, index: true
```
asmega, line 3: this `index: true` does not look to be correct. because if you look at the schema no index is added. maybe use `add_index` as an easy fix?

nigel-lowry (approved):

db/migrate/20210426150053_add_la_funded_place_urn_to_schools.rb
tonyheadford: Great spot, I've changed this to use `add_index` as suggested

asmega (approved):

PR #1616, opened 2021-04-27
Update README to say how to run spec files or spec directories in parallel

nigel-lowry:
### Context

`README.md` was missing instructions on how to run RSpec in parallel without running entire suite

Add instructions to run RSpec in parallel on spec directories or individual spec files


markhallen (approved):
LGTM

PR #1633, opened 2021-04-29
1857 handle missing ship to

tonyheadford:
### Context
[Trello card](https://trello.com/c/PzH0Z61T/1857-computacenter-cant-handle-cap-updates-when-missing-shipto)
Prevent cap changes being sent to CC when the ship-to is not present and therefore CC cannot process the request.

Prevent cap updates from being sent if the school or it's responsible body does not have a `computacenter_reference`
When the `computacenter_reference` is updated (via the CC support interface) then send cap updates (if possible). This applies to "sold-to" and "ship-to" updates via the CC support interface.

### Guidance to review
Given a RB with a `computacenter_reference` and one of its schools without a `computacenter_reference`.
Enable ordering for the school via the `SchoolOrderStateAndCapUpdateService`
No cap updates should be sent.
Mark the school as `computacenter_change_amended` so that it appears in the CC area.
As a CC user, find the school and set a ship-to reference
Cap updates should be now sent for that school.
You should be able to observe similar by removing the reference from the RB and setting one on a school, and then adding a sold-to the RB via the CC area.  In this case cap updates are sent to all the RB schools that have a ship-to.


asmega (approved):
Code seems fine

PR #1650, opened 2021-05-04
Update build-push-and-deploy-container.yml

markhallen:
### Context

The github action build-and-push-container has the following warning:
Input 'repository' has been deprecated with message: v2 is now available through docker/build-push-action@v2

We can use build-push-action v2

### Guidance to review

Confirm successful build and push of container to heroku

### DIFF
--- a/.github/workflows/build-push-and-deploy-container.yml
+++ b/.github/workflows/build-push-and-deploy-container.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - main
+      - docker-build-push-action-v2
 
 jobs:
   build-and-push-container:
--- a/.github/workflows/build-push-and-deploy-container.yml
+++ b/.github/workflows/build-push-and-deploy-container.yml
@@ -25,7 +25,7 @@ jobs:
       # Github actions. Helpfully, there's an off-the-shelf 'official' docker
       # action that does it for us
       - name: Build and push Docker image
-        uses: docker/build-push-action@v1
+        uses: docker/build-push-action@v2
         with:
           build_args: GIT_COMMIT_SHA=${{ steps.vars.outputs.sha_short }},GIT_BRANCH=${{ steps.vars.outputs.branch }}
           username: ${{ secrets.DOCKER_USERNAME }}


.github/workflows/build-push-and-deploy-container.yml
```diff
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - main
+      - docker-build-push-action-v2
```
aldavidson, line 8: I just added this line so that we could test it running in situ - please feel free to remove it when it's ready to merge

.github/workflows/build-push-and-deploy-container.yml
```diff
@@ -25,7 +26,7 @@ jobs:
       # Github actions. Helpfully, there's an off-the-shelf 'official' docker
       # action that does it for us
       - name: Build and push Docker image
-        uses: docker/build-push-action@v1
+        uses: docker/build-push-action@v2
```
aldavidson, line 29: Although this gets rid of the original deprecation warning, we now get a [warning about unexpected arguments](https://github.com/DFE-Digital/get-help-with-tech/runs/2510410399?check_suite_focus=true#step:4:1):

`Warning: Unexpected input(s) 'build_args', 'username', 'password', 'repository', valid inputs are ['allow', 'build-args', 'builder', 'cache-from', 'cache-to', 'context', 'file', 'labels', 'load', 'network', 'no-cache', 'outputs', 'platforms', 'pull', 'push', 'secrets', 'secret-files', 'ssh', 'tags', 'target', 'github-token']`

Looking at the [upgrade notes](https://github.com/docker/build-push-action/blob/master/UPGRADE.md#v1-to-v2) it seems several inputs have been renamed or restructured, so I'm not sure whether the values got passed through or not. The `build_args` certainly don't seem to have been passed through, as the `GIT_COMMIT_SHA` we ended up with on dev after [this run](https://github.com/DFE-Digital/get-help-with-tech/runs/2510410399?check_suite_focus=true#step:4:1) didn't match what I expected.

aldavidson (changes requested):

PR #1667, opened 2021-05-06
[1955] fix Trusts with nil addresses

aldavidson:
### Context

Full context is on [Trello card 1955](https://trello.com/c/hfWrhp57/1955-82-new-trusts-exposed-without-addresses). A quick summary:

- We have lots of Trusts with no address.
- The addresses have been pulled through from GIAS and are present in the staged data, but have not been propagated to the actual Trust records
- The code to propagate the changes is already written, but has never been wired up to a job, so it has never been called. 

Add migration to one-off-update all Trusts, and job to make sure future updates are propagated

### Guidance to review

First, stage the updates with `StageTrustDataJob.new.perform`
Check your local database count of `Trust.where(address_1: nil).count`
Run the migration
Check your local database count of `Trust.where(address_1: nil).count` again - it should have decreased (note: it's unlikely to decrease to zero, there are some Trusts which have no address in GIAS too )

### DIFF
--- a/app/jobs/apply_staged_attribute_changes_to_trusts_job.rb
+++ b/app/jobs/apply_staged_attribute_changes_to_trusts_job.rb
@@ -1,0 +1,5 @@
+class ApplyStagedAttributeChangesToTrustsJob < ApplicationJob
+  def perform
+    TrustUpdateService.new.update_trusts
+  end
+end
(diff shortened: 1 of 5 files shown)


markhallen (approved):
LGTM

PR #1668, opened 2021-05-06
add redis status to the healthcheck JSON

aldavidson:
### Context

The `/healthcheck.json` endpoint only shows the status of the database service, and does not include redis

Add a redis status check

### Guidance to review

Hit '/healthcheck.json` - you should see something like this:

Screenshot from 2021-05-06 14-48-37

(screenshot from localhost - if you're on a deployed container, the git variables will also be populated)


markhallen (approved):
Nice addition

nigel-lowry (approved):

PR #1673, opened 2021-05-07
Get `rails` gem back in the `Gemfile` and go to newest version

nigel-lowry:
### Context

An explicit reference to the `rails` gem in our `Gemfile` was removed due to problems with a Rails dependency called `mini_mime`. The Rails project seem to have got over this so this change reintroduces the `rails` gem to the `Gemfile` and goes to the current version.

Not having `rails` in the `Gemfile` caused me problems as a new starter last month when RubyMine couldn't find the `rails` gem to start the server and run other commands.

1. Add `rails` to `Gemfile`
2. Update `rails` dependencies


markhallen (approved):

markhallen (approved):

aldavidson (approved):
LGTM :+1:

PR #1689, opened 2021-05-13
Changes to accompany new reset devices guide

dannychadburndfe:


ralph-hawkins:
- there are quite a few places you’re using ‘a’ tags instead of the markdown — it’d be good if this was consistent
- you’re styling the text with ‘app-card’ — can we change those to govuk-inset text? app-card isn't intended for this. 'govuk-inset-text' would be better for consistency, or if you really want to use this, I think we should probably create a new class rather than repurposing this
- there’s a link on app/views/devices_guidance/device_specification.md that's just ‘read our guidance’ — that doesn’t give much context for a screen-reader, could you reword it or include more of the sentence?

nigel-lowry:
Closing because this has been merged into #1699

PR #1710, opened 2021-05-17
Update all trusts not just open ones

kenfodder:
Consider all trusts and not just open ones when updating.


kenfodder:
Superseded by https://github.com/DFE-Digital/get-help-with-tech/pull/1719

PR #1719, opened 2021-05-19
Auto close trust from GIAS if trust has no schools

kenfodder:
PROPOSAL 2 - dealing with Trust closures

Initial naive proposal was https://github.com/DFE-Digital/get-help-with-tech/pull/1710 

- Auto close trusts if safe to do so
- `def close_trusts(last_update)` - This can be used directly in the console with a very old date in the past to retroactively close all Trusts that should be closed due to closed trusts not being synced.

TODO:
- ~~Tests~~
- ~~Sentry notification and logging~~

### DIFF
--- a/app/services/trust_update_service.rb
+++ b/app/services/trust_update_service.rb
@@ -4,6 +4,18 @@ class TrustUpdateService
     last_update ||= DataStage::DataUpdateRecord.last_update_for(:trusts)
 
     # simple updates for trusts that are open
+    update_open_trusts(last_update)
+
+    # auto close trusts that have no schools
+    # skip and notify sentry when a trust is closed and still has schools - needs investigation
+    close_trusts(last_update)
+
+    DataStage::DataUpdateRecord.updated!(:trusts)
+  end
+
+private
+
+  def update_open_trusts(last_update)
     DataStage::Trust.updated_since(last_update).gias_status_open.each do |staged_trust|
       trust = Trust.find_by(companies_house_number: staged_trust.companies_house_number)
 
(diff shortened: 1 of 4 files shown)


app/services/trust_update_service.rb
```diff
@@ -4,18 +4,42 @@ def update_trusts(last_update: nil)
...
+    DataStage::DataUpdateRecord.updated!(:trusts)
+  end
+
+private
+
+  def update_open_trusts(last_update)
     DataStage::Trust.updated_since(last_update).gias_status_open.each do |staged_trust|
       trust = Trust.find_by(companies_house_number: staged_trust.companies_house_number)
 
       next unless trust
```
jits, line 22: Whilst here, is it worth logging when this happens? I know you mentioned posting a notification e.g. to Sentry (though I may have misunderstood about which case to do so), but a simple `warn` log entry could allow us to grep for when this occurs?

kenfodder: Yeah I think more general logging is needed here so we can see what happened each day.

app/services/trust_update_service.rb
```diff
@@ -4,18 +4,48 @@ def update_trusts(last_update: nil)
...
+      if trust.schools.count.positive?
+        trusts_with_schools << trust.id
+        next
+      end
+
+      close_trust(trust)
+    end
+
+    unless trusts_with_schools.empty?
+      Sentry.capture_message("TrustUpdateService#close_trusts - FAILED TO CLOSE: #{trusts_with_schools.inspect}")
```
markhallen, line 46: Do we have an example of what this will look like in sentry? I suppose if it includes the stack we can easily deduce that it failed because the Trust has schools. Otherwise we would need to add that fact to the capture_message.

markhallen: Here is an example from dev: https://sentry.io/organizations/dfe-get-help-with-tech/issues/2411110736/

kenfodder: Yep I shared a link in Slack.

It's a custom message so we know exactly where it's from, not much need for a Stacktrace as it's not an exception we are handling.

markhallen (approved):
Nice work Ken

PR #1728, opened 2021-05-25
Use Sentry scopes

kenfodder:
This call to Sentry doesn't work as the update from Raven isn't backwards compatible with the params passed in.

Updated to use scope and also add test coverage.


markhallen (approved):
LGTM

markhallen:
[Sentry: Exception around sign in token destroy](https://trello.com/c/imbUuFLR/2030-sentry-exception-around-sign-in-token-destroy)

PR #1733, opened 2021-05-26
Tactical fix to allow allocations to be decreased

jits:
… for schools that are not in a virtual cap pool.

https://trello.com/c/LcUxgM3W/1967-reducing-allocation-for-a-single-school-is-rejected

### Context

Before these changes, reducing allocations could cause an error due to the underlying `cap` value not being adjusted to match, in time for the model validation to hit.

Note: I've gone for a "light touch" fix here that doesn't attempt to change too much – there's a longer term piece required for refactoring + code consolidation + potential fixes, later on, after we figure out / confirm what the intended product design for allocations and vcap pools is moving forward.

Changes here:

- Allow allocations to be decreased, but only for schools that are not in a virtual cap pool.
- For schools that are in a virtual cap pool, we a) show a warning message in the UI that decreasing is not allowed, and b) have a validation (at the form model layer) to prevent decreasing.
- The allocation now has a minimum, set to the current `SchoolDeviceAllocation#raw_devices_ordered`, via a validation in the form model. The UI has also been updated to show a hint.
  - Note: this validation is only on the form model layer as we only want to enforce it for the support user flow. See <https://trello.com/c/FxejJUlY/716-remove-the-validation-that-devicesordered-cannot-be-greater-than-cap> for context on why this is not a model validation.
- Updated existing specs + new specs for the `AllocationForm` model.

Note an edge case here: because the main change here mirrors the cap adjustment logic in `SchoolAllocationDevice#cap_implied_by_order_state`, the cap is not adjusted for schools that have an order state of `can_order_for_specific_circumstances`, which could result in the same HTTP 422 error bug hitting if an allocation is decreased below the cap for a school with this order state.

## Screenshots

Screenshot 2021-05-25 at 16 48 21

Screenshot 2021-05-25 at 16 52 18

Screenshot 2021-05-25 at 15 42 07

Screenshot 2021-05-25 at 15 53 49

### Guidance to review

- Allocations can always be **increased**.
- Allocations can be **decreased** for schools not in a virtual cap pool.
- Allocations cannot be **decreased** for schools in a virtual cap pool.
- When decreasing an allocation, it cannot go below the current devices ordered.

### DIFF
--- a/app/form_objects/support/allocation_form.rb
+++ b/app/form_objects/support/allocation_form.rb
@@ -1,18 +1,42 @@
 class Support::AllocationForm
   include ActiveModel::Model
 
-  attr_accessor :allocation, :current_allocation, :school_allocation
+  attr_reader :allocation
+  attr_accessor :current_allocation, :school_allocation
 
-  delegate :cap, :devices_ordered, to: :school_allocation
+  delegate :cap, :raw_devices_ordered, :is_in_virtual_cap_pool?, to: :school_allocation
 
-  validates :allocation, numericality: { only_integer: true, greater_than: -1 }
+  validate :check_decrease_allowed
+  validate :check_minimum
 
   def initialize(params = {})
     super(params)
     @current_allocation = @school_allocation.dup
   end
 
+  def allocation=(value)
+    @allocation = value.to_i
+  end
+
   def order_state
     school_allocation&.school&.order_state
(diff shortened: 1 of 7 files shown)


nigel-lowry (approved):

markhallen (approved):
Nice work!

PR #1754, opened 2021-06-01
New section in /computacenter for ISS/SCL Chromebook domains

kenfodder:
Add a new section to list ISS and SCL provisions and their respective Responsible Bodies that needs to highlight the possibility that multiple chromebook domains are required.

Any ISS or SCL that has said `yes` or `i_dont_know` for Chromebooks is listed essentially.

### DIFF
--- a/app/controllers/computacenter/multi_domain_chromebooks_iss_scl_controller.rb
+++ b/app/controllers/computacenter/multi_domain_chromebooks_iss_scl_controller.rb
@@ -1,0 +1,44 @@
+require 'csv'
+
+class Computacenter::MultiDomainChromebooksIssSclController < Computacenter::BaseController
+  def index
+    @schools = School
+      .includes(:responsible_body)
+      .joins(:preorder_information)
+      .gias_status_open
+      .la_funded_provision
+      .where({
+        provision_type: %w[iss scl],
+        preorder_information: {
+          will_need_chromebooks: %w[yes i_dont_know],
+        },
+      })
+      .sort_by { |s| s.responsible_body.name }
+
+    respond_to do |format|
+      format.html { @show_download_link = @schools.any? }
+      format.csv { send_data csv_generator, filename: make_filename }
+    end
+  end
+
+private
+
+  def make_filename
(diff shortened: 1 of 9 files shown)


app/controllers/computacenter/multi_domain_chromebooks_iss_scl_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+    .joins(:preorder_information)
+    .gias_status_open
+    .la_funded_provision
+    .where({
+      provision_type: %w[iss scl],
+      preorder_information: {
+        will_need_chromebooks: %w[yes i_dont_know],
+      },
+    })
+    .sort_by { |s| s.responsible_body.name }
```
kenfodder, line 16: Sorting via SQL is a little messy with includes in the query it generates, for the number of rows expected this seems cleaner.

nigel-lowry (approved):

app/views/computacenter/home/show.html.erb
```diff
@@ -38,7 +38,13 @@
       <%= govuk_link_to 'Responsible bodies managing multiple Chromebook domains', computacenter_multi_domain_chromebooks_path %>
     </h2>
 
-    <p class="govuk-body">View details of responsible bodies that manage schools with mutliple Chromebook domains.</p>
+    <p class="govuk-body">View details of responsible bodies that manage schools with multiple Chromebook domains.</p>
+
+    <h2 class="govuk-heading-m">
+      <%= govuk_link_to 'ISS or Social Care LAs managing multiple Chromebook domains', computacenter_multi_domain_chromebooks_iss_scl_path %>
```
jits, line 44: Did you mean to use the new `multi_domain_chromebooks_iss_scl` i18n key here?

kenfodder: It should but nothing of that entire page is using it. It is however used on the title of the page it links to which all the other pages do, so I'm tempted to follow the "bad" consistency for now, it's a coincidence it's the "same" string but it could be different.

jits: Ahh, gotcha! Didn't realise that the new key was under the `page_titles` "namespace" – all good.

jits (approved):

PR #1757, opened 2021-06-02
Update documentation to reflect current procedure

nigel-lowry:
### Context

It's been a while since we've needed to generate a CSV to e-mail to CC; changes are now shown in the service UI for
CC users

Remove outdated text and make some parts clearer


nigel-lowry:
Still unsure what correct procedure is

PR #1763, opened 2021-06-03
Update specs for allocation job and refactor job

markhallen:
### Context

Negative deltas were not considered at all in the project from the beginning. We have an urgent need to reduce allocations across thousands of objects.

Expand the scope of the `AllocationJob` specs to cover more complex cases.

Refactoring of the AllocationJob to show the intent more clearly.

### Guidance to review

Focus on the correctness of the `AllocationJob` specs. The specs should be refactored in a subsequent PR. This could not be done for this PR because of time pressure.

The immediate use case will be to pass a zero delta and `order_state` = `cannot_order` to thousands of `School` objects today.

### DIFF
--- a/app/jobs/allocation_job.rb
+++ b/app/jobs/allocation_job.rb
@@ -2,38 +2,89 @@ class AllocationJob < ApplicationJob
   queue_as :default
 
   def perform(allocation_batch_job)
-    school = allocation_batch_job.school
-    order_state = allocation_batch_job.order_state
-    current_allocation = school.std_device_allocation
+    set_instance_variables(allocation_batch_job)
+    set_default_new_raw_allocation_value
+    set_default_new_raw_cap_value
 
-    current_allocation_value = current_allocation&.raw_allocation || 0
-    devices_ordered = current_allocation&.raw_devices_ordered || 0
-    new_allocation_value = [current_allocation_value + allocation_batch_job.allocation_delta, devices_ordered].max
-
-    current_cap_value = current_allocation&.raw_cap || 0
-    new_cap_value = [current_cap_value + allocation_batch_job.allocation_delta, devices_ordered].max
-
-    disable_user_notifications = !allocation_batch_job.send_notification
-
-    allocation = SchoolDeviceAllocation.find_or_initialize_by(school: school, device_type: 'std_device')
+    # A negative delta must check against devices_available_to_order
+    set_negative_allocation_and_cap_values if @allocation_delta.negative?
 
     ActiveRecord::Base.transaction do
-      allocation.update!(allocation: new_allocation_value, cap: new_cap_value)
-
-      service = SchoolOrderStateAndCapUpdateService.new(
-        school: school,
-        order_state: order_state,
-        std_device_cap: new_cap_value,
-      )
-
-      service.disable_user_notifications! if disable_user_notifications
-      service.call
-
-      if disable_user_notifications
-        allocation_batch_job.update!(processed: true)
-      else
-        allocation_batch_job.update!(processed: true, sent_notification: true)
-      end
+      update_current_allocation!
+      call_school_order_state_and_cap_update_service
+      record_batch_job_processed_and_notify
     end
+  end
+
+private
+
+  def set_allocation_batch_job
+    @allocation_batch_job = allocation_batch_job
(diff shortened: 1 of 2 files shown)


app/jobs/allocation_job.rb
```diff
@@ -2,38 +2,89 @@ class AllocationJob < ApplicationJob
...
+      call_school_order_state_and_cap_update_service
+      record_batch_job_processed_and_notify
+    end
+  end
 
-    disable_user_notifications = !allocation_batch_job.send_notification
+private
 
-    allocation = SchoolDeviceAllocation.find_or_initialize_by(school: school, device_type: 'std_device')
+  def set_allocation_batch_job
```
jits, line 21: Doesn't look like this is being used anymore (given the `set_instance_variables` method below)?

spec/jobs/allocation_job_spec.rb
```diff
@@ -279,33 +338,102 @@
...
+          create(:allocation_batch_job, urn: school1.urn, allocation_delta: '0', order_state: 'cannot_order')
+        end
+        let(:batch_allocation_job2) do
+          create(:allocation_batch_job, urn: school2.urn, allocation_delta: '0', order_state: 'cannot_order')
+        end
+
+        it 'does not change the allocation' do
+          expect {
+            described_class.perform_now(batch_allocation_job1)
+          }.to change { school1.std_device_allocation.reload.allocation }.by(0)
```
nigel-lowry, line 369: Don't worry about this now but you can write lines like this as `}.to_not change { school1.std_device_allocation.reload.allocation }`

spec/jobs/allocation_job_spec.rb
```diff
@@ -279,33 +338,102 @@
...
+            described_class.perform_now(batch_deallocation_job)
           }.to change { school1.std_device_allocation.reload.raw_allocation }.by(-1)
         end
       end
 
-      context 'maintain part of allocation if already ordered' do
-        let!(:school) { create(:school, :with_std_device_allocation_partially_ordered) }
-        let(:batch_job) { create(:allocation_batch_job, urn: school.urn, allocation_delta: '-100', order_state: 'cannot_order') }
+      context 'reduce and maintain part of allocation if already ordered' do
+        let!(:devices_available_to_deallocate) { [school1_devices_available_to_order, 198].min }
```
nigel-lowry, line 414: Don't worry about it now but that makes me think too much

nigel-lowry (approved):

app/jobs/allocation_job.rb
```diff
@@ -2,38 +2,89 @@ class AllocationJob < ApplicationJob
...
+    @new_raw_allocation_value = @current_raw_allocation_value + @negative_allocation_delta
+  end
+
+  def set_negative_new_raw_cap_value
+    @new_raw_cap_value = [@current_raw_cap_value + @negative_allocation_delta, @new_raw_allocation_value].min
+  end
+
+  def update_current_allocation!
+    if @allocation_delta.negative?
+      return unless @current_allocation.devices_available_to_order.positive? # Don't update the allocation, but allow order_state to be updated
```
jits, line 64: May be worth using `@current_allocation.devices_available_to_order?`

jits: Thinking about this change some more: is it worth decreasing the allocation amount even if no devices are available to order? I guess it depends on the _intent_ of this batch job – should it "safeguard" allocation (and thus cap) amounts, or should it make the intended change regardless (and then assume the rest of the system can handle the updated allocation amount regardless of the devices available to order)?

app/jobs/allocation_job.rb
markhallen: Correct. Removed. This is all too hurried for my liking.

spec/jobs/allocation_job_spec.rb
markhallen: Easy change. Done!

jits (approved):
Nice work!

spec/jobs/allocation_job_spec.rb
markhallen: Agreed. We already know there is refactoring work to be done here so we will clear this up in the subsequent PR.

PR #1776, opened 2021-06-08
fixes indent on screenshot so numbered list maintains sequence

ralph-hawkins:


amylaurence1 (approved):
Looks good. All fixed.

PR #1806, opened 2021-06-21
Update get_local_admin_and_bios_passwords.html.erb

dannychadburndfe:


ralph-hawkins (approved):
I fixed the quote marks, but otherwise 👍🏻

PR #1847, opened 2021-06-29
2169 update settings list

nigel-lowry:
### Context

https://trello.com/c/hOra3k4r

* Remove links to schools
* Remove status column
* Reorder and rename columns
* Display devices ordered rather than allocations
* Only show "Devices received" column to RBs lacking a virtual cap
* Change breadcrumb


nigel-lowry:
[Update the Settings list](https://trello.com/c/hOra3k4r/2169-update-the-settings-list)

app/components/display_devices_ordered_component.rb
```diff
@@ -1,13 +1,13 @@
...
   def initialize(school:)
     @school = school
   end
 
-  def allocations
-    std_devices_count = school.std_device_allocation&.raw_allocation.to_i
-    coms_devices_count = school.coms_device_allocation&.raw_allocation.to_i
+  def devices_ordered
+    std_devices_count = school.std_device_allocation&.devices_ordered || 0
+    coms_devices_count = school.coms_device_allocation&.devices_ordered || 0
```
jits, line 10: Should this (and the one above) use `raw_devices_ordered` instead?

jits: (`devices_ordered` will show the amount from the `VirtualCapPool`, if applicable – maybe that's intended?)

app/controllers/responsible_body/devices/schools_controller.rb
```diff
@@ -1,7 +1,9 @@
 class ResponsibleBody::Devices::SchoolsController < ResponsibleBody::BaseController
   before_action :load_schools_by_order_status, only: %i[show index]
 
-  def index; end
+  def index
+    @show_devices_ordered_column = @responsible_body.virtual_cap_pools.none?
```
jits, line 5: Might also need to check `@responsible_body.has_virtual_cap_feature_flags?`, just in case? (Not sure if it really matters, but I can imagine a situation with the data where the RB has vcap pools but then the flag for that particular rb was turned off for some reason).

app/components/display_devices_ordered_component.rb
nigel-lowry: The column the devices ordered is displayed in is only shown when the RB does not have a virtual cap pool, so I think we're alright. I'm not clear on `raw_devices_ordered` so please let me know if you think this is still a problem.

app/controllers/responsible_body/devices/schools_controller.rb
nigel-lowry: I'll add that check too to be safe.

app/components/display_devices_ordered_component.rb
jits: The problem is the component can't assume that; `raw_devices_ordered` is just the "raw" value on the `SchoolDeviceAllocation` (i.e. not taking into account vcap pools) – if this component is ever reused in the future that assumption won't necessarily hold. Best to use `raw_devices_ordered`, I think. (Of course, the underlying "issue" is the naming and way vcap pools are done – but that's for another day :D)

jits (approved):
LG

PR #1884, opened 2021-07-29
2078 allocation job

nigel-lowry:
### Context

https://trello.com/c/VJyULra4

Instead of creating an `ActiveJob` per school, creates one `ActiveJob` which will be easier to manage if there's a problem.

Does the bare minimum, a much bigger refactoring or complete rewrite is probably needed.


nigel-lowry:
[Background job refinements for allocations](https://trello.com/c/VJyULra4/2078-background-job-refinements-for-allocations)

PR #1898, opened 2021-08-03
Make requested text changes

nigel-lowry:
### Context

Text change in Trello card https://trello.com/c/bZQcyiHL

### Guidance to review

To keep the structure of the Markdown template unchanged, the heading levels are untouched. This means the headings jump immediately from `h2` to `h4` for the outlined options. This might be how we want it to look but is semantically incorrect.

The dashes are just plain dashes but semantically ought to be em-dashes (but our old UX has a preference for the semantically incorrect en-dashes).

Some of the hyperlinks have been removed as per the document. Someone might want to check that that's the desired behaviour.

https://docs.google.com/document/d/1mkzDbXGK77bTl3ZErSH7R7lSfgZYrBnsNgnQqUfyzCw/edit


nigel-lowry:
[Redraft GOV.UK content for Device distribution and ownership.](https://trello.com/c/bZQcyiHL/2195-redraft-govuk-content-for-device-distribution-and-ownership)

markhallen (approved):

PR #1905, opened 2021-08-05
Add class to encrypt/decrypt strings (for database fields)

nigel-lowry:
### Context

Rails 7 is going to add database encryption but is not released yet. We'll probably upgrade to Rails 7 fairly soon after 
release, so we'll use a lightweight solution in the meantime before then.

https://trello.com/c/fZFmCWR2

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -25,6 +25,10 @@ computacenter:
 cookie_consent:
   expiry_time_months: 6
 
+database_field_encryption:
+  key: 402aebb6e150652727db9bda
+  salt: b84bc74fa39cef36e44f6ced
+
 google:
   analytics:
     tracking_id:
--- a/lib/encryption_service.rb
+++ b/lib/encryption_service.rb
@@ -1,8 +1,8 @@
 class EncryptionService
   # for our enviornments, generate one-off environment variables:
   #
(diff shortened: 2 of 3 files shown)


nigel-lowry:
[Encrypt Admin password, BIOS password and hardware hash](https://trello.com/c/fZFmCWR2/2312-encrypt-admin-password-bios-password-and-hardware-hash)

ltello (approved):

config/settings.yml
```diff
@@ -25,6 +25,10 @@ computacenter:
 cookie_consent:
   expiry_time_months: 6
 
+database_field_encryption:
```
kenfodder, line 28: Does this actually get used? Seem like it just using the env vars in the code?

kenfodder (approved):

config/settings.yml
nigel-lowry: I hope that these variables get picked up by everyone's `development` environment so that it won't complain about missing environment variables when the code using this eventually goes in. The `test` environment doesn't pick up these variables from `settings.yml` so I had to mock them in the specs. I just thought I'd get it in now before I forgot.

kenfodder (approved):

PR #1913, opened 2021-08-11
2208: new approach to add trusts and keep staged trusts and rb trusts…

ltello:
… as separate processes in a pipeline

### User Story
[GIAS - Add new trusts (RFQ-12)](https://trello.com/c/f8wxcnWe)

### Context
On the periodic import of GIAS trusts csv file sometimes new trusts are coming. Even though we are currently staging them, we are not creating the Trust entries associated.

Create a new Trust entry for every trust coming from GIAS on csv import not already existing in RB/Trust table.

### Guidance to review
To keep RB/Trust processing separated from staged trust this new approach will add the missing RB/Trust entries as part of the _ApplyStagedAttributeChangesToTrustsJob_ instead of _StageTrustDataJob_ 

I haven´t renamed _TrustUpdateService_ to _TrustUpsertService_ because the current name is used in old migrations, etc.  I think it still makes sense to use Update word if we take the general sense of it?


ltello:
[GIAS - Add new trusts (RFQ-12)](https://trello.com/c/f8wxcnWe/2208-gias-add-new-trusts-rfq-12)

app/services/trust_update_service.rb
```diff
@@ -15,6 +18,13 @@ def update_trusts(last_update: nil)
 
 private
 
+  def create_trusts
+    existing_trust_ids = Trust.pluck(:companies_house_number)
+    DataStage::Trust.where.not(companies_house_number: existing_trust_ids).find_each do |staged_trust|
```
kenfodder, line 23: I think we might need to do DataStage::Trust.gias_status_open, there is no point importing in closed trusts.

kenfodder (changes requested):

nigel-lowry (approved):

PR #1914, opened 2021-08-11
Hard delete users without TS confirmed account

kenfodder:
### Context

Reduce the number of soft-deleted users in the system.

When deleting users check if they have a TS account confirmed, if not hard-delete otherwise soft-delete as a precaution for now.

### Guidance to review

- Added a service just for deletion, everything else kept the same
- Removed dependency of extra mobile requests to be deleted as they should be retained


ltello (approved):
LGTM

nigel-lowry (approved):

PR #1924, opened 2021-08-13
Bugfix: Soft-deleted users only visible to support section only

kenfodder:
Deleted users should only be visible to support section only, normal users shouldn't see any soft-deleted users.

* In support - added extra "Deleted: Yes" row to indicate deleted users for support reasons.


spec/features/support/viewing_service_performance_spec.rb
```diff
@@ -42,6 +42,7 @@
   end
 
   scenario 'DfE users can query for completions between given dates' do
+    Timecop.freeze
```
kenfodder, line 45: Flakey test if executed across a different minute.

markhallen (approved):

PR #1937, opened 2021-08-18
Destroying a user doesn't delete associated ExtraMobileDataRequests

kenfodder:


ltello (commented):
Should we add something automatic like a migration to destroy all the orphan **extra_mobile_data_request** records existing in the db at deployment time?

Or maybe manually from the console in all environments?

kenfodder:
@ltello The PR is to NOT delete them if a user is destroyed, because they belong to an RB we don't want to lose the data.

markhallen (approved):

ltello:
> @ltello The PR is to NOT delete them if a user is destroyed, because they belong to an RB we don't want to lose the data.

You are right. Sorry I misread it!

PR #1940, opened 2021-08-19
Feature flag to pause gias updates

kenfodder:


README.md
```diff
@@ -172,23 +172,23 @@ Over time, the timestamped tags will accumulate in Docker Hub, and should be pru
 
 Some values are configurable with environment variables:
```
kenfodder, line 174: Just re-aligning this messy table

README.md
```diff
@@ -203,7 +203,11 @@ These are activated by having an environment variable FEATURES_(flag name) set t
 FEATURES_rate_limiting=active bundle exec rails s
 ```
 
-The available flags are listed in `app/services/feature_flag.rb`, and available in the constant `FeatureFlag::FEATURES`. Each one is tested with a dedicated spec in `spec/features/feature_flags/`.
```
kenfodder, line 206: There is no `spec/features/feature_flags/` folder 🤔

README.md
```diff
@@ -203,7 +203,11 @@ These are activated by having an environment variable FEATURES_(flag name) set t
...
+The available flags are listed in `app/services/feature_flag.rb`, and available in the constant `FeatureFlag::FEATURES`.
+
+Name                           | Description
+-------------------------------| -----------------------------------------------------------------------------------------------------------------------------------
+FEATURES_gias_data_stage_pause | Pauses any GIAS data staging of trusts and schools, should be used e.g. in Septemeber when GIAS data is changed a lot and unsettled
```
kenfodder, line 211: Adding this here to explain what it does, we should do that with flags in the future to document

app/services/feature_flag.rb
```diff
@@ -3,6 +3,7 @@ class FeatureFlag
     rate_limiting
     display_sign_in_token_links
     show_component_previews
+    gias_data_stage_pause
```
ltello, line 6: Looking at the FeatureFlag code, a feature is only active if their associated envar has the value *'active'*, otherwise it won't be active.
So we don't strictly need it but are you planning to set the new envar to *'inactive'* manually in all enviroments via the cf command line?

kenfodder: Ideally we should set it to `inactive`, but I purposely made it so we set it when we need it and without it, it's not going to break anything. 

I will in the other ticket do a clean up and make sure they are all set to something and remove the unused ones :)

ltello: Great! LGTM

ltello (approved):

PR #1954, opened 2021-08-23
Only count asset as seen if by non-support user

nigel-lowry:
### Context

Only count asset as viewed if it's not been viewed by a support user (last minute requirement)

https://trello.com/c/01Tm1uXh


nigel-lowry:
[Record when a password has been shown to at least one user](https://trello.com/c/01Tm1uXh/2309-record-when-a-password-has-been-shown-to-at-least-one-user)

markhallen (approved):

kenfodder (approved):

PR #1993, opened 2021-09-03
Add page to check allow support to check assets by serial numbers

nigel-lowry:
### Context

https://trello.com/c/H9b4UXH3

No link to this page yet for UX testing with internal teams. Access by enter this into the address bar https://get-help-with-tech.education.gov.uk/support/asset_serial_uploads/new


markhallen (approved):
It would be nice to also validate the uploaded file type.

PR #1994, opened 2021-09-03
Skip invite if user already belongs to the school

kenfodder:
The shovel `<<` throws a PG index duplicate entry error, so we should check before shovelling a school in.


nigel-lowry (approved):

nigel-lowry:
[BUG: Inviting existing school user can cause an error if already invited](https://trello.com/c/hnTTf4jb/2409-bug-inviting-existing-school-user-can-cause-an-error-if-already-invited)

PR #1997, opened 2021-09-06
2109: add/remove school from rb pools when it changes responsible body

ltello:
### Context
See [card 2109: Totals wrong for responsible body virtual pool](https://trello.com/c/WBc27VQn)

- Add tests to make sure recomputing pool allocations is done properly
- Remove school from pool when it leaves a responsible body. Recompute rb remaining allocations
- Add school to pool when it is moved to a different responsible body. Recompute rb new allocations

### DIFF
--- a/app/models/responsible_body.rb
+++ b/app/models/responsible_body.rb
@@ -54,6 +54,15 @@ class ResponsibleBody < ApplicationRecord
     end
   end
 
+  def remove_school_from_virtual_cap_pools!(school)
+    raise(VirtualCapPoolError, 'Virtual cap feature flags not set') unless has_virtual_cap_feature_flags?
+
+    school.device_allocations.each do |allocation|
+      pool = virtual_cap_pools.send(allocation.device_type).first
+      pool&.remove_school!(school)
+    end
+  end
+
   def can_school_be_added_to_virtual_cap_pools?(school)
     has_virtual_cap_feature_flags? &&
       school.responsible_body_id == id &&
@@ -61,6 +70,14 @@ class ResponsibleBody < ApplicationRecord
       school.preorder_information&.responsible_body_will_order_devices? &&
       school.device_allocations.any? &&
       !has_school_in_virtual_cap_pools?(school)
+  end
+
+  def can_school_be_removed_from_virtual_cap_pools?(school)
+    has_virtual_cap_feature_flags? &&
+      school.responsible_body_id == id &&
+      school.preorder_information&.responsible_body_will_order_devices? &&
+      school.device_allocations.any? &&
+      has_school_in_virtual_cap_pools?(school)
   end
 
   def devices_available_to_order?
--- a/app/models/responsible_body.rb
+++ b/app/models/responsible_body.rb
@@ -63,7 +63,7 @@ class ResponsibleBody < ApplicationRecord
     end
   end
 
-  def can_school_be_added_to_virtual_cap_pools?(school)
+  def school_addable_to_virtual_cap_pools?(school)
     has_virtual_cap_feature_flags? &&
       school.responsible_body_id == id &&
       !school.la_funded_provision? &&
@@ -72,7 +72,7 @@ class ResponsibleBody < ApplicationRecord
       !has_school_in_virtual_cap_pools?(school)
   end
 
-  def can_school_be_removed_from_virtual_cap_pools?(school)
+  def school_removable_from_virtual_cap_pools?(school)
     has_virtual_cap_feature_flags? &&
       school.responsible_body_id == id &&
       school.preorder_information&.responsible_body_will_order_devices? &&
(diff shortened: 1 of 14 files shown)


nigel-lowry (approved):

ltello:
[Totals wrong for responsible body virtual pool](https://trello.com/c/WBc27VQn/2109-totals-wrong-for-responsible-body-virtual-pool)

app/models/responsible_body.rb
```diff
@@ -63,6 +72,14 @@ def can_school_be_added_to_virtual_cap_pools?(school)
       !has_school_in_virtual_cap_pools?(school)
   end
 
+  def can_school_be_removed_from_virtual_cap_pools?(school)
```
markhallen, line 75: It looks like you followed the naming convention of the method above `can_school_be_added_to_virtual_cap_pools?`, however we could take this opportunity to fix the naming.
[https://rubystyle.guide/#bool-methods-prefix](url)

markhallen: "has" is a grey area though. Let's leave it.

app/models/virtual_cap_pool.rb
```diff
@@ -55,6 +60,11 @@ def add_school_allocation(device_allocation)
     recalculate_caps!
   end
 
+  def remove_school_allocation(device_allocation)
```
markhallen, line 63: Naming: `school` is redundant here and maybe we need a bang `!` at the end?

app/services/change_school_responsible_body_service.rb
```diff
@@ -31,14 +34,26 @@ def log_error(e)
...
+  def school_can_be_added?
+    new_responsible_body.can_school_be_added_to_virtual_cap_pools?(school)
+  end
+
+  def school_can_be_removed?
+    initial_responsible_body&.can_school_be_removed_from_virtual_cap_pools?(school)
   end
 
-  def update_school!
+  def set_school_new_responsible_body
```
markhallen, line 49: Does this method name need a bang `!`?

app/services/change_school_responsible_body_service.rb
```diff
@@ -31,14 +34,26 @@ def log_error(e)
...
     school.update!(
-      responsible_body_id: new_responsible_body_id,
+      responsible_body_id: new_responsible_body.id,
       computacenter_change: COMPUTACENTER_CHANGE_STATUS,
     )
   end
+
+  def update_preorder_information
```
markhallen, line 56: Bang here too?

spec/form_objects/support/school/change_responsible_body_form_spec.rb
```diff
@@ -80,17 +80,17 @@
 
     context 'when the form is valid' do
       let(:school) { create(:school, :with_preorder_information) }
-      let(:new_responsible_body_id) { create(:trust).id }
+      let(:new_responsible_body) { create(:trust) }
 
-      subject(:form) { described_class.new(school: school, responsible_body_id: new_responsible_body_id) }
+      subject(:form) { described_class.new(school: school, responsible_body_id: new_responsible_body.id) }
```
markhallen, line 85: Sorry for the pedantry, but this could use: `responsible_body: new_responsible_body`

markhallen (approved):
Minor pedantry from me, but looks good. Nice work!

app/services/change_school_responsible_body_service.rb
ltello: Shouldn´t we add the bang only to the very internal method doing the dangerous thing? 
Otherwise we should add the ending bang to all the chain of methods calling that one?

app/services/change_school_responsible_body_service.rb
ltello: Same comment as above. I don´t really mind to rename it but maybe we should not

app/models/virtual_cap_pool.rb
ltello: The thing is that this removes not all the allocations of the pool but only those of a specific school.

app/models/responsible_body.rb
ltello: 👍  Yes, I was following the existing pattern but I have renamed this one and the twin about adding

ltello:
BTW @markhallen changes to the school RB now included in the timeline

app/services/change_school_responsible_body_service.rb
kenfodder: So for me, it's "more" a bang method than not...
1. It wraps update!
2. It changes the recipient and it returns the modified recipient.
3. It "feels" method that needs to be executed with caution or a warning.

kenfodder (approved):
Bang or not, LGTM.

ltello:
Ok, methods renamed

PR #2016, opened 2021-09-14
2221: make preorder_information a school internal artifact. Do not ex…

ltello:
…pose it

### Context
PreorderInformation is a week entity that fully depends on a School. This idea is supported on:
- It is implemented as a single association to a School.
- PreoderInfomation class code is really entangled with School code with loads of calls back and forth.
- The system code is full of calls like `school.preorder_information.any_message` all over the place.
- This is also a violation of the [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) design principle that rapidly increases the number of dependencies makes the system very difficult to extend/maintain.
  
I believe we don´t need a PreorderInformation entity in the system. Actually all of his methods/attributes can be considered part of a School instead. As a separate entity we need to deal with loads of situations where it exists already or not and/or create/build it which I think don´t model properly the business domain.

Entirely removing it from the system is risky at this point, so as a first step, I have tried to hide it as much as possible to the rest of the system by accessing it only via the parent school entity. That would make it act as an internal artifact chosen by design to implement part of a School behaviour, not accessible externally.

Similarly, SchoolDeviceAllocation entities show the same issues as PreorderInformation. The code is accessing them mostly from the associated school entity. We would need to work much more on it in future PRs. In this one, I have just defined a few public methods in School public interface to access laptop allocation, cap and devices_ordered via the school therefore removing direct access associated SchoolDeviceAllocations again violating Law of Demeter.

### Guidance to review
- Calls `school.preorder_information.any_message` replaced with equivalents messaging school directly.
- Several delegations to preorder_information defined inside School as a way to expose publicly its behaviour from School and hide it from PreorderInformation entities.
- New public interface to access laptop allocation/cap/devices ordered from School instead of device_allocation associations.
- See [2221](https://trello.com/c/fdRvvbDR)
- There is a lot more work to do to keep improving our code and design when it comes to schools preorder information, device allocations, etc... This PR is huge enough as to leave it for future attempts.


ltello:
[Refactor `SchoolDeviceAllocation` methods for `cap`, `allocation` and `devices_ordered`](https://trello.com/c/fdRvvbDR/2221-refactor-schooldeviceallocation-methods-for-cap-allocation-and-devicesordered)

nigel-lowry (approved):

nigel-lowry (approved):

kenfodder (approved):
Can't see anything obviously wrong, it's a lot cleaner 👍

PR #2018, opened 2021-09-14
2434 remove more support portal mentions

nigel-lowry:
### Context

https://trello.com/c/pdwz4eNc

### Guidance to review

This contains an *irreversible* migration

### DIFF
--- a/app/components/school_user_summary_list_component.rb
+++ b/app/components/school_user_summary_list_component.rb
@@ -6,12 +6,7 @@ class SchoolUserSummaryListComponent < UserSummaryListComponent
 
   def rows
     info = super
-    info += [
-      {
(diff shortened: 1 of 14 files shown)


db/migrate/20210914103636_change_user_orders_devices_default.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeUserOrdersDevicesDefault < ActiveRecord::Migration[6.1]
+  def change
+    change_column_default :users, :orders_devices, from: nil, to: false
```
nigel-lowry, line 3: Not sure whether to keep this migration reversible like this. Developer thoughts?

kenfodder (approved):

PR #2019, opened 2021-09-14
Ressurect welcome wizard

kenfodder:
### Context

School user sign in journey

Re-enable the sign in wizard flow from https://github.com/DFE-Digital/get-help-with-tech/pull/1822/files

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -125,8 +125,14 @@ private
   end
 
   def school_root_url_for(user)
-    if user.schools.one?
-      home_school_path(user.schools.first)
+    if user.schools.size == 1
+      if user.school.preorder_information&.school_will_order_devices? &&
+          user.school.preorder_information&.chromebook_info_still_needed? &&
+          !user.school.la_funded_provision?
+        before_you_can_order_school_path(user.school)
+      else
+        home_school_path(user.schools.first)
+      end
     else
       schools_path
     end
--- a/app/services/interstitial_picker.rb
+++ b/app/services/interstitial_picker.rb
@@ -6,20 +6,51 @@ class InterstitialPicker
   end
 
   def call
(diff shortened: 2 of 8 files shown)


app/controllers/application_controller.rb
```diff
@@ -125,8 +125,14 @@ def root_url_for_responsible_body(user)
   end
 
   def school_root_url_for(user)
-    if user.schools.one?
-      home_school_path(user.schools.first)
+    if user.schools.size == 1
+      if user.school.preorder_information&.school_will_order_devices? &&
+          user.school.preorder_information&.chromebook_info_still_needed? &&
+          !user.school.la_funded_provision?
```
markhallen, line 131: Maybe we could create a predicate method on the School instance with a name that describes the intent of the combination of these three predicate method calls?

app/controllers/application_controller.rb
```diff
@@ -125,8 +125,14 @@ def root_url_for_responsible_body(user)
   end
 
   def school_root_url_for(user)
-    if user.schools.one?
-      home_school_path(user.schools.first)
+    if user.schools.size == 1
+      if user.school.preorder_information&.school_will_order_devices? &&
```
markhallen, line 129: We could use a case statement or terminating if statements instead for clarity.

app/services/interstitial_picker.rb
```diff
@@ -6,21 +6,52 @@ def initialize(user:)
...
-                  partial: 'interstitials/default',
+                  partial: 'interstitials/school_user',
                 )
               else
                 OpenStruct.new(
                   title: title_for_default,
-                  partial: 'interstitials/school_user',
+                  partial: 'interstitials/default',
                 )
               end
```
markhallen, line 34: A `case` might read better.

markhallen (approved):
Only pedantry from me :-)
Looks good!

PR #2043, opened 2021-09-22
Export CSV of assets first viewed during date range

nigel-lowry:
### Context

https://trello.com/c/7nZxVgYn

Unfortunately I couldn't find a way to print the error messages if the date range is wrong because of redirect errors, etc. Freddie's said this is acceptable. The UX is that if there's something wrong with the date range, the CSV download button just does nothing. 

Rather than leave the error messaging half-finished, it makes no attempt to print the error messages *at all*.

### DIFF
--- a/app/concerns/csv_value_sanitiser.rb
+++ b/app/concerns/csv_value_sanitiser.rb
@@ -5,7 +5,8 @@ class CsvValueSanitiser
   end
 
   def safe?
-    @value =~ /\A[^=+\-@\t\r]/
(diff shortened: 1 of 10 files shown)


nigel-lowry:
[Support users can export admin/bios password-viewed data (RFQ-X)](https://trello.com/c/7nZxVgYn/2466-support-users-can-export-admin-bios-password-viewed-data-rfq-x)

markhallen (approved):

PR #2058, opened 2021-09-29
Capture allocation change meta-data

markhallen:
### Context
Develop functionality to capture allocation information relating to an over-order and automatically adjust the allocation value to match devices ordered

Currently over-orders are identified by the allocations team by comparing the `allocation` value and the `device_ordered` value.

The downsides of this approach are:
- It skews the allocation totals
- It hides unused allocations
- The over-order data will be lost when we need to increase allocation for an over-ordered school

```
Feature: Capture over-orders

    Historical over-orders should be captured apart from the actual allocation

Scenario: Over-ordered devices
Given I am logged in as a support user
  And a school X has over-ordered devices above their allocation
When I visit the school page
Then I see the historical records for school X's over-orders
```

Rough example:
- 20 Jan 21 over-ordered 12 devices
- 13 Feb 21 over-ordered 40 devices

Introduce an AllocationChange model to record allocation change meta-data when required.

Hook the creation of the AllocationChange record onto the relevant actions.

### Guidance to review

 [Trello card](https://trello.com/c/U9QWfPHC)


nigel-lowry (approved):

ltello:
[Provide an automatic process which allows the service to reconcile the allocations against service partner over-orders (RFQ-07)](https://trello.com/c/U9QWfPHC/2222-provide-an-automatic-process-which-allows-the-service-to-reconcile-the-allocations-against-service-partner-over-orders-rfq-07)

app/models/preorder_information.rb
```diff
@@ -33,32 +33,22 @@ def initialize(*args)
...
-      if responsible_body_will_order_devices? && school.can_order_devices_right_now?
-        'rb_can_order'
-      elsif responsible_body_will_order_devices? && school.has_ordered?
-        'ordered'
+    if school_will_order_devices?
+      if any_school_users?
+        return 'school_contacted' unless chromebook_information_complete?
+        return 'school_can_order' if school.can_order_devices_right_now?
+
+        school.has_ordered? ? 'ordered' : 'school_ready'
```
markhallen, line 42: We could `return` on this line and replace the `else` with `end` on the next line.

app/services/allocation_over_order_service.rb
```diff
@@ -0,0 +1,79 @@
...
+                                   allocation_change_category: :over_order,
+                                   allocation_change_description: 'Over Order').call
+  end
+
+  def over_ordered
+    raw_devices_ordered - raw_allocation
+  end
+
+  def reclaim_allocation_across_virtual_cap_pool
+    to_claim = available_allocations_in_the_vcap_pool.inject(over_ordered) do |needed, member|
```
markhallen, line 57: `to_claim` is a bit confusing. Maybe `to_reclaim` is a bit closer to what we are doing here?

app/services/allocation_over_order_service.rb
```diff
@@ -0,0 +1,79 @@
...
+      needed -= reclaim_allocation_from_vcap_pool_member(member, needed: needed)
+      needed.zero? ? break : needed
+    end
+
+    alert_pool_allocation_reclaim_failed(to_claim) if to_claim
+  end
+
+  def reclaim_allocation_from_vcap_pool_member(member, needed: 0)
```
markhallen, line 65: What about `quantity` instead of `needed`? I think there is enough context in the method name for us to use `quantity`.

app/services/update_school_devices_service.rb
```diff
@@ -32,20 +35,41 @@ def notify_other_agents
     CapUpdateNotificationsService.new(*allocation_ids, notify_school: notify_school).call if allocation_ids.any?
   end
 
+  def record_allocation_change_meta_data(allocation_id:, prev_allocation:, new_allocation:)
+    if allocation_change_category || allocation_change_description
```
markhallen, line 39: A terminating `unless` statement would be preferable to wrapping this in an `if` block.

db/schema.rb
```diff
@@ -29,6 +29,19 @@
     t.index ["batch_id"], name: "index_allocation_batch_jobs_on_batch_id"
   end
 
+  create_table "allocation_changes", force: :cascade do |t|
+    t.bigint "school_device_allocation_id", null: false
+    t.string "category"
+    t.integer "delta"
```
markhallen, line 35: Are we keeping `delta` in the db?

markhallen: Sorry, I see `delta` is still referenced. Are we keeping the three values? `prev_allocation`, `new_allocation` and `delta`

app/models/preorder_information.rb
ltello: refactored

app/services/allocation_over_order_service.rb
ltello: sure. Changed

app/services/allocation_over_order_service.rb
ltello: Renamed

app/services/update_school_devices_service.rb
ltello: The thing is that Rubocop doesn´t like inline `if` or `unless` clauses when the statement is a multiline one.

db/schema.rb
ltello: I wouldn´t like to have the 3 of them. I just didn´t modify it but I will do now.

markhallen (commented):
Great work!

PR #2184, opened 2021-11-01
Improve processing of vcap schools in batch jobs

ltello:
### Context
Updating school device allocations via csv enqueues a background job per school. 
Non vcap schools get processed individually with no issues. 

However, each vcap school persist its new raw allocation numbers, recomputes and updates the pool allocation numbers and notify external agents. This is sometimes causing deadlocks when 2 jobs of schools of the same vcap get processed concurrently and try to update the vcap instance simultaneously.

- This PR removes any automatic callback triggered to recompute a vcap after changes in a school.
- `AllocationJobs` for vcap schools are no longer created when uploading an allocations csv file.
- Instead, `CalculateVcapJob` instances are enqueued for each vcap.
- When processed by the queue backend each `CalculateVcapJob` will:
    - process all `AllocationBatchJobs` of its vcap schools silently with no notifications at all
    - recompute the vcap.
    - send notifications only once per school.

### DIFF
--- a/app/form_objects/support/bulk_allocation_form.rb
+++ b/app/form_objects/support/bulk_allocation_form.rb
@@ -25,17 +25,45 @@ class Support::BulkAllocationForm
 
 private
 
+  def add_school_to_vcap(school)
+    vcaps[school.responsible_body_id] += 1
+  end
+
+  def create_allocation_batch_job(props)
+    job_attrs = props.merge(batch_id: batch_id, send_notification: send_notification)
+    AllocationBatchJob.create!(job_attrs)
+  end
+
+  def recalculate_vcaps
+    vcaps.each_key do |responsible_body_id|
+      CalculateVcapJob.perform_later(responsible_body_id: responsible_body_id,
+                                     batch_id: batch_id,
+                                     notify_school: send_notification)
+    end
+  end
+
   def rows
     @rows ||= CSV.read(upload.path, headers: true)
   end
 
+  def schedule_school(school, props)
+    allocation_batch_job = create_allocation_batch_job(props)
+    school.in_virtual_cap_pool? ? add_school_to_vcap(school) : AllocationJob.perform_later(allocation_batch_job)
+  end
+
   def upload_scheduled?
     rows.each do |row|
-      job_attrs = row.to_h.slice(*CSV_HEADERS).merge(batch_id: batch_id, send_notification: send_notification)
-      AllocationJob.perform_later(AllocationBatchJob.create!(job_attrs))
+      props = row.to_h.slice(*CSV_HEADERS).symbolize_keys!
+      school = School.where_urn_or_ukprn_or_provision_urn(props[:urn] || props[:ukprn]).first
+      schedule_school(school, props) if school
     end
+    recalculate_vcaps
     true
   rescue StandardError
     false
+  end
+
+  def vcaps
+    @vcaps ||= Hash.new { |hash, key| hash[key] = 0 }
   end
 end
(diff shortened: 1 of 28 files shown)


app/form_objects/support/bulk_allocation_form.rb
```diff
@@ -25,17 +25,45 @@ def batch_id
 
 private
 
+  def add_school_to_vcap(school)
+    vcaps[school.responsible_body_id] += 1
+  end
```
markhallen, line 30: We should consider using more descriptive names here so that it is clear that it is not adding to an actual `vcap`.
I'm not 100% sure what the name should be, but something along the lines of `add_school_for_post_processing_of_vcaps`. Also, the `vcaps` accessor method and the `vcap` hash itself should also be a bit more descriptive.

ltello: Renamed

nigel-lowry (approved):

app/mailers/can_order_devices_mailer.rb
```diff
@@ -100,6 +100,7 @@ def notify_support_school_can_order_but_no_one_contacted
 private
 
   def tracked_template_mail(message_type, template_id, mail_params = {})
+    logger.info "--- LOCAL DEBUG: #{self.class.name}.#{message_type}.with(#{params})" if ENV['LOCAL_DEBUG']
```
markhallen, line 103: Do we want to keep this?

app/mailers/computacenter_mailer.rb
```diff
@@ -32,6 +32,7 @@ def notify_of_school_can_order
 private
 
   def setup_params
+    logger.info "--- LOCAL DEBUG: #{self.class.name}.#{caller_locations(1, 1)[0].label}.with(#{params})" if ENV['LOCAL_DEBUG']
```
markhallen, line 35: Do we want to keep this?

app/models/computacenter/outgoing_api/cap_update_request.rb
```diff
@@ -18,6 +18,7 @@ def post
     # overrides when testing
     @payload_id ||= SecureRandom.uuid
     @body = construct_body
+    logger.info "--- LOCAL DEBUG: #{body}" if ENV['LOCAL_DEBUG']
```
markhallen, line 21: Do we want to keep this?

app/models/responsible_body.rb
```diff
@@ -207,31 +209,32 @@ def orders_managed_by_schools?
...
 
-  def recalculate_laptop_vcap
-    Rails.logger.info("***=== recalculating caps ===*** responsible_body_id: #{id} - laptops")
-    values = vcap_schools.select('SUM(raw_laptop_allocation) as allocation_sum, SUM(raw_laptop_cap) as cap_sum, SUM(raw_laptops_ordered) as ordered_sum')[0]
-    update!(laptop_allocation: values&.allocation_sum.to_i,
-            laptop_cap: values&.cap_sum.to_i,
-            laptops_ordered: values&.ordered_sum.to_i)
-    update_cap_on_computacenter(:laptop) if vcap_active? && (laptop_cap_previously_changed? || laptops_ordered_previously_changed?)
+  def calculate_laptop_vcap(**opts)
+    logger.info("***=== recalculating caps ===*** responsible_body_id: #{id} - laptops")
```
markhallen, line 221: Do we want to keep this?

app/mailers/can_order_devices_mailer.rb
ltello: It's not ideal, I know, but it is so useful when analyzing the behavior of each test case...! 
It prints exactly the data and/or emails sent to CC

app/models/computacenter/outgoing_api/cap_update_request.rb
ltello: Anyways. I will remove them now.

markhallen (approved):
Beautiful 👍

PR #2185, opened 2021-11-03
Temporarily remove API route

nigel-lowry:
### Context

Temporarily remove API route to stop CC repeatedly calling our API (and generating new e-mails from our system)


nigel-lowry:
Not the approach we'll take

PR #2193, opened 2021-11-07
Fix over orders

ltello:
### Context
The existing implementation of school allocations overloads and mixes the meaning and role of `raw_cap` and `order_state` fields for schools with some undesirable consecuencies:

- When a school `order_state` is changed, an automatic mechanism to auto-adjust the value of `raw_cap` is triggered before persisting the change. This way, we are persisting twice the fact that a school is set to 'orders for this school not allowed': `order_state: :cannot_order` and `raw_cap: raw_device_ordered`. This leads to eventual data inconsistencies in the system very difficult to detect, prevent and fix. The test suite is full of samples where schools are simultaneously in `:cannot_order` whilst device numbers mismatch that state:
```
 { 
  order_state: :cannot_order,
  raw_laptop_allocation: 5,
  raw_laptop_cap: 3,
  raw_laptops_ordered: 1
 }
```

- Another caveat due to the implementation above affects over-orders preventing us from being able to accurately store the cap stolen from sibbling schools in virtual caps:
 If a school in a vcap has over-ordered devices, its `raw_allocation` and `raw_cap` values are changed to match the new `raw_devices_ordered` value. At the same time, sibbling schools in the vcaps get as many over-ordered devices stolen from their caps to compensate the overall vcap allocation. However, due to the current implementation: 
- If we decrease the `raw_cap` of a sibbling school to represent the stolen devices, it will automatically be set back to match their `raw_allocation` value if the school is in `:can_order` state (because of the auto-adjust mechanism described above).
- To overcome this issue, we need to reduce both, `raw_cap` and `raw_allocation`. This means, we lose the initial allocation assigned to  the sibblings schools involved.

As we can see from above, there are multiple variables to take into account to compute the final effective cap of a school. Currently we are considering a few of them: assigned allocation, order_state, number of stolen/over_ordered devices, number of devices ordered... In the future we might add new ones. 

The current implementation trying to persist the final result in `raw_cap_x` has proved not to be valid in all scenarios. We should be able to persist all the variables affecting a school cap and compute the result based on it whenever we need it. Only that way we can fully cope with all the requirements and leave it open for future extensions.

This PR is aimed to fix the issues above. The following changes are proposed:

- Remove the cap auto-adjust mechanism when allocation numbers are going to be persisted.
- Store devices gained/lost because of own/sibblings over orders in new fields `:over_order_reclaimed_laptops` and `:over_order_reclaimed_routers`
- Store devices lost after changing the state of a school to `:can_order_for_some_circumstances` in new fields `:circumstances_laptops` and `:circumstances_routers`
- Remove `raw_laptop_cap` and `raw_router_cap` from School schema.

School cap is now a function of:
- order state,
- allocation,
- over order devices gained/lost
- circumstances devices lost

```
if cannot_order? 
  raw_devices_ordered
else
  raw_allocation + over_order_reclaimed_devices + circumstances_devices
end
```

This will allow us to:
- Know and display the initial allocation assigned to any school anytime.
- Know and update/give-back the number of devices gained/lost due to over orders transactions.
- Know and update/revert the number of devices lost because of state changes to `can_order_for_some_circumstances` state and back.
- Set back the previous cap of a school  when it is temporarily set to `cannot_order` and back to  a can order state with no loss of information.
- Easily extend cap with new components always keeping/displaying/updating data related to other variables composing the allocation cap.


ltello:
[Over order handling](https://trello.com/c/I3dH0JGI/2545-over-order-handling)

db/migrate/20211102150517_cap_changes_for_over_ordered_schools.rb
```diff
@@ -0,0 +1,26 @@
...
+    change_column_default :cap_changes, :prev_cap, from: nil, to: 0
+    change_column_default :cap_changes, :new_cap, from: nil, to: 0
+
+    restore_school_assigned_allocations
+  end
+
+private
+
+  def restore_school_assigned_allocations
+    School.includes(:cap_changes).find_each do |school|
```
markhallen, line 17: `School.joins(:allocation_changes)` should scope this to Schools with an associated AllocationChange.

app/models/cap_change.rb
```diff
@@ -1,10 +1,11 @@
-class AllocationChange < ApplicationRecord
+class CapChange < ApplicationRecord
   belongs_to :school
 
   enum category: {
     allocation_error_reversal: 'allocation_error_reversal',
     increase: 'increase',
     over_order: 'over_order',
+    over_order_pool_give_back: 'over_order_pool_give_back',
```
markhallen, line 8: `over_order_pool_rollback` ?

markhallen (approved):

markhallen (approved):

db/migrate/20211102150517_cap_changes_for_over_ordered_schools.rb
ltello: True. Changed!

db/migrate/20211105224402_replace_raw_cap_fields_from_schools.rb
```diff
@@ -0,0 +1,47 @@
...
+                                                              THEN 0
+                                                           ELSE raw_router_cap - raw_router_allocation
+                                                       END)")
+    School.update_all('circumstances_routers = raw_router_cap - (raw_router_allocation + over_order_reclaimed_routers)')
+  end
+
+  def populate_raw_cap_fields
+    School.update_all('raw_laptop_cap = raw_laptop_allocation + over_order_reclaimed_laptops + circumstances_laptops')
+    School.update_all('raw_router_cap = raw_router_allocation + over_order_reclaimed_routers + circumstances_routers')
+  end
```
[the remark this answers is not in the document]
ltello: Locally I have a db similar to prod in size. Actually a dump a did from it. When I run it took just a few secs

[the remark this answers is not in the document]
markhallen: Sounds good to me!

app/models/cap_change.rb
ltello: Changed.

markhallen (approved):

markhallen:
LGTM

PR #2199, opened 2021-11-09
fix over order reverting bug

ltello:
- fix method call: replace `quantity.negative` with `quantity.negative?`
- add  methods to school that return  full raw allocation numbers info.
- refactor scope
- rename `give_back_cap_across_virtual_cap_pool` method


markhallen (approved):

ltello:
[Over order handling](https://trello.com/c/I3dH0JGI/2545-over-order-handling)

PR #2205, opened 2021-11-10
2560: recompute router vcaps on CalculateVcapJob

ltello:
### Context
Bug: We also need to update router vcaps when processing  Allocation jobs!


nigel-lowry (approved):

ltello:
[Allocation update not sending vcap router `cap_updates`](https://trello.com/c/sNLqzp7U/2560-allocation-update-not-sending-vcap-router-capupdates)

PR #2215, opened 2021-11-15
2566 devolved management vcap rbs

ltello:
### Context
Some RBs are set to devolve management to schools but still have a vcap to manage for some of their centrally managed schools. This is related to [this card](https://trello.com/c/sd6vRGo8)

Allow rbs like those described above to be vcap active. Currently only rbs managing centrally were allowed.


nigel-lowry (approved):

app/models/responsible_body.rb
```diff
@@ -35,7 +34,7 @@ class ResponsibleBody < ApplicationRecord
   after_update :maybe_generate_user_changes
 
   def self.chosen_who_will_order
```
markhallen, line 36: This method name is also misleading. It should be clear that this is only a default in absence of a specific setting for the school.

In fact this method only seems to be used in the performance dashboard and the same distinction could be made there:
https://github.com/DFE-Digital/get-help-with-tech/blob/cde94d3932fabaff56896480b73fc7f83f88b0b8/app/models/support/service_performance.rb#L384-L388

app/models/responsible_body.rb
```diff
@@ -232,11 +231,11 @@ def next_school_sorted_ascending_by_name(school)
   end
 
   def orders_managed_centrally?
```
markhallen, line 233: The latest findings also throw doubt on the accuracy of this method name.
The method body is only referencing the default in the absence of a per school setting.
Perhaps this method should be checking the `vcap_feature_flag` state or it should be renamed.

app/models/responsible_body.rb
```diff
@@ -232,11 +231,11 @@ def next_school_sorted_ascending_by_name(school)
...
-    who_will_order_devices == 'responsible_body'
+    default_who_will_order_devices_for_schools == 'responsible_body'
   end
 
   def orders_managed_by_schools?
```
markhallen, line 237: It seems that the value `schools` has been cleaned up in production and we should be able to get rid of this.

```
[25] pry(main)> School.where(who_will_order_devices: 'school').count
=> 17191
[26] pry(main)> School.where(who_will_order_devices: 'schools').count
=> 0
```

app/models/responsible_body.rb
```diff
@@ -232,11 +231,11 @@ def next_school_sorted_ascending_by_name(school)
...
-    %w[school schools].include?(who_will_order_devices)
+    %w[school schools].include?(default_who_will_order_devices_for_schools)
   end
```
markhallen, line 239: Both these methods could return `true`. Are we sure that they are being used correctly?

app/models/responsible_body.rb
ltello: I have searched the code and this is not used anymore. Deleted!

app/models/responsible_body.rb
```diff
@@ -261,21 +260,20 @@ def trust?
...
-      .or(schools.who_will_order_devices_not_set)
+    return School.none unless vcap_feature_flag?
+
+    not_laf_schools = schools.excluding_la_funded_provisions
+    not_laf_schools_centrally_managed = not_laf_schools.responsible_body_will_order_devices
+    if orders_managed_centrally?
+      not_laf_schools_centrally_managed.or(not_laf_schools.who_will_order_devices_not_set)
+    else
+      not_laf_schools_centrally_managed
+    end
```
markhallen, line 272: This is difficult to read. Maybe it could be pushed to a scope in the `School` model?

app/models/responsible_body.rb
ltello: It was named like that to exactly match the name of the equivalent method on schools.
What about renaming this to: `responsible_body_will_order_devices_for_schools_by_default?`

app/models/responsible_body.rb
ltello: I have refactored/renamed these 2 like this:
```
  def responsible_body_will_order_devices_for_schools_by_default?
    default_who_will_order_devices_for_schools == 'responsible_body'
  end

  def schools_will_order_devices_by_default?
    default_who_will_order_devices_for_schools == 'school'
  end

```

app/models/responsible_body.rb
markhallen:
```
[29] pry(main)> ResponsibleBody.where(who_will_order_devices: 'schools').count
=> 0
[30] pry(main)> ResponsibleBody.where(who_will_order_devices: 'school').count
=> 3043
```

app/models/responsible_body.rb
```diff
@@ -232,11 +231,11 @@ def next_school_sorted_ascending_by_name(school)
   end
 
   def orders_managed_centrally?
-    who_will_order_devices == 'responsible_body'
+    default_who_will_order_devices_for_schools == 'responsible_body'
   end
 
   def orders_managed_by_schools?
-    %w[school schools].include?(who_will_order_devices)
```
markhallen, line 239: It seems that the value `schools` has been cleaned up in production and we should be able to get rid of this.

```
[25] pry(main)> School.where(who_will_order_devices: 'school').count
=> 17191
[26] pry(main)> School.where(who_will_order_devices: 'schools').count
=> 0
[29] pry(main)> ResponsibleBody.where(who_will_order_devices: 'schools').count
=> 0
[30] pry(main)> ResponsibleBody.where(who_will_order_devices: 'school').count
=> 3043
```

app/models/responsible_body.rb
ltello: What about this?
 ```
  def vcap_schools
    return School.none unless vcap_feature_flag?
    return schools.excluding_la_funded_provisions.school_not_set_to_order_devices if orders_managed_centrally?

    schools.excluding_la_funded_provisions.responsible_body_will_order_devices
  end
````

File: app/models/responsible_body.rb (hunk 6 above)
Comment (ltello) [reply app/models/responsible_body.rb:239]: Yes, removed!

File: app/models/responsible_body.rb (hunk 2 above)
Comment (markhallen) [reply app/models/responsible_body.rb:233]: That name sounds good to me.

File: app/models/responsible_body.rb (hunk 5 above)
Comment (ltello) [reply app/models/responsible_body.rb]: I would keep this into the ResponsibleBody model as it has to do with the internal logic on how a vcap works. Vcaps live in the ResponsibleBody now, not VirtualCapPool or School anymore.

File: app/services/responsible_body_set_who_will_order_devices_service.rb [hunk 7]
```diff
@@ -15,15 +15,15 @@ def call
...
     responsible_body.active_schools.each do |school|
       SchoolSetWhoManagesOrdersService.new(school,
                                            who,
                                            clear_preorder_information: true,
                                            recalculate_vcaps: false,
-                                           notify: responsible_body.orders_managed_by_schools?).call
+                                           notify: responsible_body.schools_will_order_devices_by_default?).call
     end
-    responsible_body.calculate_vcaps! if responsible_body.orders_managed_centrally?
+    responsible_body.calculate_vcaps! if responsible_body.responsible_body_will_order_devices_for_schools_by_default?
```
markhallen, line 26: I may be missing some context here, but would we also need to recalculate the vcap for rbs where the rb does not order by default, but does order for some schools?

app/views/shared/_change_who_will_order.html.erb
```diff
@@ -8,7 +8,7 @@
         <span class="govuk-caption-xl"><%= school.name %></span>
         <%= t('page_titles.change_who_will_order') %>
       </h1>
-      <% if school.responsible_body.vcap_active? %>
+      <% if school.responsible_body.vcap_feature_flag? && school.responsible_body.responsible_body_will_order_devices_for_schools_by_default? %>
```
markhallen, line 11: This seems like complex logic for a view. Can we refactor it?

spec/models/responsible_body_spec.rb
```diff
@@ -699,4 +675,53 @@
...
+    let(:management_not_set_school) { create(:school, responsible_body: responsible_body, who_will_order_devices: nil) }
+    let(:la_funded_place) { create(:iss_provision, responsible_body: responsible_body) }
+
+    before do
+      centrally_managed_school
+      management_not_set_school
+      la_funded_place
+    end
+
+    context 'when the responsible body has not vcaps enabled' do
```
markhallen, line 691: "has vcaps disabled" or "does not have vcaps enabled" reads better here.

markhallen (approved):

app/models/responsible_body.rb
markhallen: Agreed

app/services/responsible_body_set_who_will_order_devices_service.rb
ltello: Actually yes, now we could remove the condition as it is no longer required to manage centrally

PR #2228, opened 2021-11-19
Improve over orders

ltello:
### Context
Allocation numbers of schools in vcaps need adjustment not only when a school allocation numbers change but also when a school is added or removed to/from a vcap to another, a school changes rb, a school changes who will order devices, etc.

This leads us to think the adjustments of cap should be done at vcap level (when recomputing vcaps) rather than when updating school allocations (UpdateSchoolDeviceService).

This should avoid automatically some of the defects we had been having these days in some cases of over ordering and rollbacks, schools in and out of vcaps, vcaps with some schools that can order and some not, ...

This PR:
- moves the lend/borrowing cap mechanism inside vcaps 
- automatically rebalance vcaps over order numbers whenever the vcap numbers need recalculation
- Register ALL raw_cap changes for a school rather than only over-order related as currently.
-  New cap change categories has been added as args to calls to UpdateSchoolDevicesService
- Expect Sentry messages anytime a vcap is recalculated and lend/borrowing numbers can´t be balanced.

### DIFF
--- a/app/form_objects/support/allocation_form.rb
+++ b/app/form_objects/support/allocation_form.rb
@@ -25,7 +25,9 @@
 private
 
   def allocation_updated?
-    UpdateSchoolDevicesService.new(school: school, "#{device_type}_allocation".to_sym => allocation).call
+    UpdateSchoolDevicesService.new(school: school,
+                                   "#{device_type}_allocation".to_sym => allocation,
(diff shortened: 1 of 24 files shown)


nigel-lowry (approved):

markhallen (approved):
Another hole plugged :-)

PR #2239, opened 2021-11-23
Fix where supplier outage banner text wasn't showing

nigel-lowry:
### Context

https://trello.com/c/fA60xOqO


nigel-lowry:
[Supplier outage banner not showing correctly](https://trello.com/c/fA60xOqO/2598-supplier-outage-banner-not-showing-correctly)

PR #2268, opened 2021-11-29
Add absolute allocation column to allocation bulk upload csv files

ltello:
### Context
[Card](https://trello.com/c/jdjjt0oq): allow absolute allocations to be uploaded

Add an alternative new column "allocation" to bulk allocation csv files to set absolute allocation values as opposed to deltas.
A mix of delta and absolute allocation schools are allowed in the same file.
If a school has both set, delta allocation will be ignored and absolute allocation will prevail.

### DIFF
--- a/app/form_objects/support/bulk_allocation_form.rb
+++ b/app/form_objects/support/bulk_allocation_form.rb
@@ -3,7 +3,7 @@ require 'csv'
 class Support::BulkAllocationForm
   include ActiveModel::Model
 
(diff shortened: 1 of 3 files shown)


ltello:
[Add an allocation job which deals with absolute values rather than deltas (RFQ-04)](https://trello.com/c/jdjjt0oq/2305-add-an-allocation-job-which-deals-with-absolute-values-rather-than-deltas-rfq-04)

markhallen (approved):

PR #2282, opened 2021-12-02
2058: refactor SchoolCanOrderDevicesNotifications

ltello:
### Context
[2058](https://trello.com/c/TzKtp4sY)

Refactor SchoolCanOrderDevicesNotifications and make clear what is notified at anytime

### DIFF
--- a/app/models/school_can_order_devices_notifications.rb
+++ b/app/models/school_can_order_devices_notifications.rb
@@ -1,7 +1,7 @@
 class SchoolCanOrderDevicesNotifications
   attr_reader :school, :notify_computacenter
 
-  def initialize(school: nil, notify_computacenter: true)
+  def initialize(school, notify_computacenter: true)
     @school = school
     @notify_computacenter = notify_computacenter
   end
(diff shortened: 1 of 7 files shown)


ltello:
[SchoolCanOrderDevicesNotifications has a single explicit reference to std_devices (RFQ-04)](https://trello.com/c/TzKtp4sY/2058-schoolcanorderdevicesnotifications-has-a-single-explicit-reference-to-stddevices-rfq-04)

app/models/school_can_order_devices_notifications.rb
```diff
@@ -1,108 +1,110 @@
...
+  end
+
+  def rb_user?(user)
+    user.in?(school.responsible_body_users)
+  end
+
+  def relevant_users
+    return @relevant_users if instance_variable_defined?(:@relevant_users)
+
+    @relevant_users = school.organisation_users.presence || school.responsible_body_users.presence
```
markhallen, line 98: Can this be written as:
`@relevant_users ||= school.organisation_users.presence || school.responsible_body_users.presence`

Or am I missing something?

ltello: The thing is that if I do what you are suggesting, if the result is nil, everytime this method is called it will trigger the queries and recompute it all again and again. 
I mean the memoization doesn't work if the result is nil.

markhallen: Makes sense: `nil` is a valid value.

app/models/school_can_order_devices_notifications.rb
```diff
@@ -1,108 +1,110 @@
...
+
+  def org_user?(user)
+    user.in?(school.organisation_users)
+  end
+
+  def rb_user?(user)
+    user.in?(school.responsible_body_users)
+  end
+
+  def relevant_users
```
markhallen, line 95: Do we need to filter out soft deleted users?

markhallen: Could be filtered here either:
https://github.com/DFE-Digital/get-help-with-tech/blob/ef1eddcfbb807adfeb3371fafd40390603477427/app/models/school.rb#L356-L361

ltello: Good idea.
Modified

markhallen (approved):

PR #2286, opened 2021-12-06
Add a site wide banner

markhallen:
### Context

For the upcoming Christmas period a site wide banner is required to notify site visitors that deliveries will be paused.

The optional env var `GHWT__SITE_BANNER_MESSAGE` can be used to set the text for a site wide banner message.

### Guidance to review

In local dev launch the app with the env var present:
`GHWT__SITE_BANNER_MESSAGE=hello bundle exec rails s`

### DIFF
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -2,6 +2,6 @@
 
 class Site
   def self.banner_message
-    @banner_message ||= Settings.site_banner_message
+    @banner_message ||= Settings.site_banner_message.presence
   end
 end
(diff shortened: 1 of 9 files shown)


app/models/site.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class Site
+  def self.banner_message
+    @banner_message ||= Settings.site_banner_message
```
ltello, line 5: To prevent issues with blank values like the empty string maybe we could do here:
`@banner_message ||= Settings.site_banner_message.presence`

markhallen: Done!

ltello (approved):

PR #2287, opened 2021-12-06
2371 clean feature flags

ltello:
### Context
Some envars to enable/disable legacy feature flags still exist on all enviroments.
When we undefine a feature flag, the code calling it must be refactored as well.

- Rubocop custom cop to detect when code is still checking the status of feature flags that have been removed.
- Review all FEATURES_* envars in the existing environments to remove the legacy ones.
- A single list of feature flag keys is enough to cope with our needs. Temporary list is not needed anymore.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -3,7 +3,9 @@ inherit_gem:
     - config/default.yml
     - config/rails.yml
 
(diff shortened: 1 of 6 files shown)


markhallen (approved):
Awesome!
image

PR #2304, opened 2021-12-13
Add a tick box for support audit export users

markhallen:
### Context

Support want the ability to export only school and rb users

Add a checkbox to the support export users form

### Guidance to review

image


markhallen:
[Add scoping to support export user to only include school and rb users](https://trello.com/c/ACyjKN8j/2641-add-scoping-to-support-export-user-to-only-include-school-and-rb-users)

PR #2305, opened 2021-12-13
Revert router value and just hide the router row instead

kenfodder:


markhallen (approved):

app/components/responsible_body/school_details_summary_list_component.rb
```diff
@@ -72,7 +72,7 @@ def device_allocation_row
   def router_allocation_row
     {
       key: 'Router allocation',
-      value: 'routers are not available',
+      value: @school.raw_allocation(:router).positive? ? 'routers are available to order' : 'no routers are available to order',
```
ltello, line 75: `School` class exposes these methods that can be used instead based on what you want to display and wether or not you want to consider the vcap data (if the school is in vcap).
The raw_* methods will only consider the school data no matter if it is in a vcap or not.
The rest of methods take into account if the school is inside a vcap or not:

```ruby
  def can_order_routers_only_right_now?
    eligible_to_order? && !devices_available_to_order?(:laptop) && devices_available_to_order?(:router)
  end

  def devices_available_to_order(device_type)
    [0, cap(device_type) - devices_ordered(device_type)].max
  end

  def devices_available_to_order?(device_type = nil)
    return devices_available_to_order(device_type).positive? if device_type

    devices_available_to_order?(:laptop) || devices_available_to_order?(:router)
  end

  def raw_devices_available_to_order(device_type)
    [0, raw_cap(device_type) - raw_devices_ordered(device_type)].max
  end
```

PR #2339, opened 2022-01-07
Remove donated devices feature

kenfodder:
Remove all code related to Donated Device Requests feature since it was previously disabled and hasn't been reinstated since.

The model, model spec and factory are retained as the data is retained for historical purposes.


markhallen (approved):
image

PR #2371, opened 2022-01-18
2674: store tranche files into S3 to be able to share them with Sidekiq

ltello:
### Context
  Tranches uploaded by a support user needs to made accessible to Sidekiq to read and process them. 
Govuk Paas allows the use of S3 as a backend service for our apps. 

- Added [aws-s3-bucket service](https://docs.cloud.service.gov.uk/deploying_services/s3/#amazon-s3) manually to our apps in all envs
- Bind services - apps in the manifest files.
- Change the logic of bulk allocation uploads to write and read the user file to/from S3.

### Guidance to review
 I would like to isolate S3 stuff into our own wrapper class so the entities using it (`BulkAllocationForm`  and `BulkAllocationJob`) don't have to deal with the S3 client itself. 

Due to deadline restrictions I will leave it for the future.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,8 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 ruby File.read('.ruby-version').chomp
 
+gem 'aws-sdk-s3', '~> 1'
+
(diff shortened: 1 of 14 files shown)


ltello:

PR #2372, opened 2022-01-18
2674: initialize AWS S3 explicitly from automatic envars

ltello:
### Context
  Adding and binding aws-s3-bucket service to the app automatically adds some envars but we still need to explicitly initialiase the client from those envars.


markhallen (approved):

ltello:

PR #2376, opened 2022-01-19
2674: little refactor

ltello:
### Context
 - Rename var to be different to the existing @jobs instance var.
 - Add @batch_id instance var to be used by the template so it gets displayed even when jobs have not started execution yet.

image


markhallen (approved):

ltello:

PR #2422, opened 2022-02-04
2698: successor schools inherit order state from predecessors if not …

ltello:
…in a vcap

### Context
  Make successor schools inherit order_state when they are created if not in a vcap.
  [Card](https://trello.com/c/1kTmB2d8)

### DIFF
--- a/app/services/school_update_service.rb
+++ b/app/services/school_update_service.rb
@@ -24,6 +24,7 @@ class SchoolUpdateService
 
     add_school_links(staged_school, school)
     set_who_manages_orders(school)
+    inherit_order_state(school, staged_school) unless school.vcap?
 
     school
   end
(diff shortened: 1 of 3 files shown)


app/services/school_update_service.rb
```diff
@@ -58,6 +59,11 @@ def find_predecessor(staged_school)
     School.find_by(urn: last_link.link_urn) if last_link
   end
 
+  def inherit_order_state(school, staged_school)
+    predecessor = find_predecessor(staged_school)
```
markhallen, line 63: This feels a bit confusing, but I can't think of a clearer way to write it

markhallen (approved):

ltello:
[cannot_order schools (failure to inherit order state)](https://trello.com/c/1kTmB2d8/2698-cannotorder-schools-failure-to-inherit-order-state)

PR #2447, opened 2022-02-21
2762: only display user organisation name when there is one

ltello:
### Context

[Card](https://trello.com/c/Xw2axoIi)


markhallen (approved):

ltello:
[User logged in as 'false'](https://trello.com/c/Xw2axoIi/2762-user-logged-in-as-false)

PR #2468, opened 2022-03-03
2755: restrict permissions to order devices

ltello:
### Context

[Card](https://trello.com/c/ycFQju9o)


app/models/device_supplier/user_report.rb
```diff
@@ -22,7 +22,7 @@ def add_report_rows
         .relevant_to_device_supplier
         .includes(:schools, :responsible_body, :last_user_change, schools: :responsible_body)
         .find_each do |user|
-      csv << csv_row(user)
+      csv << csv_row(user) if user.relevant_to_computacenter?
```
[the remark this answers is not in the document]
markhallen: OK. It is. I see the subtle difference now! :-)

markhallen, line 25: Let's see how this performs in the real world and refine the scope later if necessary.

app/policies/school_policy.rb
```diff
@@ -26,4 +30,14 @@ def support_third_line_editable?
...
   end
+
+private
+
+  def devolved_management_school_user?
+    record.orders_managed_by_school? && user.school_ids.include?(record.id)
+  end
+
+  def responsible_body_user?
+    user.responsible_body_id.present? && user.responsible_body_id == record.responsible_body_id
```
markhallen, line 41: Should this also check on the user object if orders_devices = true? At least for the devices_orderable? context on line 12. If so we also need to rename this method.

app/models/device_supplier/user_report.rb
ltello: Yes, that is the alternative but that would make that scope very complex.

app/policies/school_policy.rb
```diff
@@ -26,4 +30,14 @@ def support_third_line_editable?
   def update_computacenter_reference?
     user.is_computacenter?
   end
+
+private
+
+  def devolved_management_school_user?
+    record.orders_managed_by_school? && user.school_ids.include?(record.id)
```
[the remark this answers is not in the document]
ltello: The aim of `devices_orderable?` is to check if the user would be eligible to order devices not if already orders. I mean if the school is centrally managed, the user should not be eligible to set as orders_devices: true

app/policies/school_policy.rb
ltello: See my reply above

app/policies/school_policy.rb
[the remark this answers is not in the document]
markhallen: OK. That makes sense.

markhallen (approved):

ltello:
[Investigate VCAP, school users changing to can order and how many users have that ability already](https://trello.com/c/ycFQju9o/2755-investigate-vcap-school-users-changing-to-can-order-and-how-many-users-have-that-ability-already)

PR #2470, opened 2022-03-07
2755: ignore orders_devices param for not eligible users

ltello:

Extension of a previous [PR](https://github.com/DFE-Digital/get-help-with-tech/pull/2468) to:
- exclude the order_devices field from the form when inviting a user to a school centrally managed by its rb.
- ignore the orders_devices param in the update action for a non-order-devices-eligible school user.
- set a false default value on orders_devices property for non-order-devices-eligible users invited to a school.

### DIFF
--- a/app/controllers/school/users_controller.rb
+++ b/app/controllers/school/users_controller.rb
@@ -5,6 +5,7 @@ class School::UsersController < School::BaseController
 
   def new
     @user = @school.users.build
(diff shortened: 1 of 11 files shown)


ltello:
[Investigate VCAP, school users changing to can order and how many users have that ability already](https://trello.com/c/ycFQju9o/2755-investigate-vcap-school-users-changing-to-can-order-and-how-many-users-have-that-ability-already)

PR #2472, opened 2022-03-07
2479 async asset csv download

kenfodder:
### Context

- Export to CSV for /assets will cause timeouts for large downloads as the reverse proxy will timeout before the CSV has finished generating.

- For less than or equals to 50 devices download at request time
- For larger downloads schedule a sidekiq job and email the user when the email is ready for download

### Guidance to review

- Refactored the asset fetching and setting of first_viewed_at is now in a service
- Existing functionality works as is
- For over 50 devices it schedules the job instead with some messaging


ltello (approved):
LGTM

PR #2608, opened 2022-05-20
2923: job to email restricted device passwords to all users

ltello:
### Context
We need to email all school and rb users with DfE restricted devices the passwords of their devices.

Use GovUK Notify service to send the emails including a link to a file stored in Notify containing all their password information.

### DIFF
--- a/app/jobs/bulk_restricted_device_password_emailing_job.rb
+++ b/app/jobs/bulk_restricted_device_password_emailing_job.rb
@@ -1,0 +1,47 @@
+class BulkRestrictedDevicePasswordEmailingJob < ApplicationJob
+  queue_as :default
+
+  ALL_SETTINGS = :all
+
+  def perform(rb_limit: ALL_SETTINGS, school_limit: ALL_SETTINGS)
+    @rb_limit = rb_limit
+    @school_limit = school_limit
+    schedule_settings(rbs)
+    schedule_settings(schools)
+  end
+
+private
+
+  attr_reader :rb_limit, :rbs, :school_limit, :schools
+
+  def rb_limit?
+    rb_limit != ALL_SETTINGS
+  end
+
+  def school_limit?
+    school_limit != ALL_SETTINGS
+  end
+
+  def rbs
+    return @rbs if instance_variable_defined?(:@rbs)
+
+    query = ResponsibleBody.with_restricted_devices_and_users
+    @rbs = rb_limit? ? query.limit(rb_limit) : query
+  end
+
+  def schools
+    return @schools if instance_variable_defined?(:@schools)
+
+    query = School.with_restricted_devices_and_users
+    @schools = school_limit? ? query.limit(school_limit) : query
+  end
+
+  def schedule_settings(settings)
+    settings.find_each do |setting|
+      RestrictedDevicePasswordEmailingForSettingJob.perform_later(
+        setting_id: setting.id,
+        setting_classname: setting.class.name
+      )
+    end
+  end
+end
--- a/app/jobs/bulk_restricted_device_password_emailing_job.rb
+++ b/app/jobs/bulk_restricted_device_password_emailing_job.rb
@@ -3,44 +3,46 @@ class BulkRestrictedDevicePasswordEmailingJob < ApplicationJob
 
   ALL_SETTINGS = :all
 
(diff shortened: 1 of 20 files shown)


ltello:
[Automate the sending of admin/BIOS passwords via notify](https://trello.com/c/5ameqEwA/2923-automate-the-sending-of-admin-bios-passwords-via-notify)

app/jobs/bulk_restricted_device_password_emailing_job.rb
```diff
@@ -0,0 +1,47 @@
+class BulkRestrictedDevicePasswordEmailingJob < ApplicationJob
+  queue_as :default
+
+  ALL_SETTINGS = :all
+
+  def perform(rb_limit: ALL_SETTINGS, school_limit: ALL_SETTINGS)
+    @rb_limit = rb_limit
```
markhallen, line 7: I could understand this, but I had to read ahead foirst. Maybe `rb_limit_scope` might explain more here about the intention?

app/jobs/bulk_restricted_device_password_emailing_job.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def school_limit?
+    school_limit != ALL_SETTINGS
+  end
+
+  def rbs
+    return @rbs if instance_variable_defined?(:@rbs)
+
+    query = ResponsibleBody.with_restricted_devices_and_users
+    @rbs = rb_limit? ? query.limit(rb_limit) : query
```
markhallen, line 29: There is a lot going on here too. Maybe we could extract `ResponsibleBody.with_restricted_devices_and_users` to a method and the use the common memoize `@rbs ||=` to make it a bit easier to understand?

app/jobs/restricted_device_password_emailing_for_setting_job.rb
```diff
@@ -0,0 +1,37 @@
+class RestrictedDevicePasswordEmailingForSettingJob < ApplicationJob
+  queue_as :default
+
+  def perform(setting_classname:, setting_id:)
+    @setting = setting_classname.constantize.find(setting_id)
```
markhallen, line 5: We could just use dependency injection here to pass the class instead of converting it to a String and then back again.
eg

````
      RestrictedDevicePasswordEmailingForSettingJob.perform_later(
        setting_id: setting.id,
        setting_class: setting.class
      )
````

and then...

````
@setting = setting_class.find(setting_id)
````

File: app/jobs/restricted_device_password_emailing_for_setting_job.rb [hunk 4]
```diff
@@ -0,0 +1,37 @@
...
+        .deliver_later
+    end
+  end
+
+  def organisation_name
+    setting.name
+  end
+
+  def users
+    @users ||= setting.users.accepting_restricted_devices_comms
```
markhallen, line 35: Nice scope `accepting_restricted_devices_comms` :-)

app/models/asset.rb
```diff
@@ -24,6 +24,12 @@ class Asset < ApplicationRecord
   }.freeze
 
   NON_SUPPORT_ATTRIBUTES = SUPPORT_ATTRIBUTES.except(:department_sold_to_id, :location_cc_ship_to_account).freeze
+  CLOSURE_EMAILING_ATTRIBUTES = {
```
markhallen, line 27: `EMAILING` feels very specific. What about something more abstract like `NOTIFICATION`?

app/jobs/restricted_device_password_emailing_for_setting_job.rb
ltello: Maybe ActiveJob accepts complex param types but I like to pass simple ones as recommended. 
For example, pass the user_id of an instance as param instead of the user itself.
I applied the same criteria here and passed just the name of the class instead of the class object itself?

app/jobs/bulk_restricted_device_password_emailing_job.rb
ltello: Good, I will rename it

app/models/asset.rb
ltello: Nice. I will rename it

markhallen (approved):

PR #2614, opened 2022-05-24
2900 publish banner

markhallen:
### Context

A site wide header banner is required to share detailed and important information relating to service closure.

Added the `SiteBannerInfoComponent` to be displayed on all pages.

Quirks:

1. The partial used for the message can be changed using an environment variable in case fast switching is required.
2. `Site.long_form_site_banner_message_partial` is used to expose `Settings.long_form_site_banner_message_partial`. Even though it feels redundant, it did not feel correct to use `Settings` directly in the ViewComponent.

### Guidance to review

Locally start the rails server:

`GHWT__LONG_FORM_SITE_BANNER_MESSAGE_FLAG=true bundle exec rails s`

Ensure that the info banner is displayed.

image


markhallen:
[Write and publish banner telling users that access to Techsource will end](https://trello.com/c/7vUCn9ZS/2900-write-and-publish-banner-telling-users-that-access-to-techsource-will-end)

PR #2615, opened 2022-05-24
deliver login emails with top priority

ltello:
### Context
  emails with links for users to log into the system are sometimes delayed or stuck if heavy jobs are running in all the background workers

  Create a top priority queue named :login and enqueue all login email jobs in that queue so a login link email will always be delivered first when a worker becomes free.


markhallen (approved):
Nice!

ltello:
[Prioritize processing of login emails in Sidekiq](https://trello.com/c/qnrjdcxD/2977-prioritize-processing-of-login-emails-in-sidekiq)

